PR #2, opened 2015-07-16
Add getter for validationErrors

hannesj:


andimarek:
thanks

andimarek:
@hannesj just wanted to let you know that I fixed some minor issue in the git history with push force .. so don't be confused when you see different hashs. I'm sorry for that  ... a history rewrite should never happen again. Thanks again for using graphql-java and contributing!

PR #5, opened 2015-08-12
Fix typo in method name

ntkoopman:
Fixes my largest problem with the library at this point.


andimarek:
Embarrassing mistake ... thanks for fixing this.

PR #8, opened 2015-08-13
added the Long built in type

creactiviti:
Added a built-in Long scalar type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@
 .idea
 .gradle
 build
+/bin/
+target
+.classpath
+.project
+.settings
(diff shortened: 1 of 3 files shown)


andimarek:
Thanks for taking your time and creating this PR, but there is an Issues with this approach:

We can't just add a new language element like LongValue. We want to conform to the spec and there is simply no literal for representing long values. The official way for new scalar types is to represent them as Strings.

This is done in the `coerceLiteral` method: Converting String to Long. So I would suggest I will add something like the code snippet I pasted here: https://github.com/andimarek/graphql-java/issues/6

I hope this is ok for you and will fit your needs.

creactiviti:
Cool @andimarek. I think that's very workable. Thanks.

andimarek:
I added the type in this commit: https://github.com/andimarek/graphql-java/commit/948231fb92ea39c47ed639ba8ce44c8e8ea82412

creactiviti:
Thanks @andimarek

PR #23, opened 2015-09-09
Pass name in as the default value if nothing is provided

tberman:
This appears to be in line w/ graphql-js implementation:

https://github.com/graphql/graphql-js/blob/1c5cdbd314200a5bc78ef52b6be9194587a55437/src/type/definition.js#L823


andimarek:
you're right ... thx

PR #26, opened 2015-09-15
Added fetching from raw fields.

erikogenvik:
The FieldDataFetcher will fetch data directly from raw fields.
It can be enabled by calling the "fetchField()" method on the field definition builder.

### DIFF
--- a/src/main/java/graphql/schema/FieldDataFetcher.java
+++ b/src/main/java/graphql/schema/FieldDataFetcher.java
@@ -1,0 +1,55 @@
+package graphql.schema;
+
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Map;
+
+import static graphql.Scalars.GraphQLBoolean;
+
+/**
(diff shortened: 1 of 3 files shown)


andimarek:
Thanks a lot for your PR and especially adding also adding a test!

What I'm not so sure about is the `fetchField` in the builder. Don't you think it would be better just to set the `DataFetcher` instead having extra field? This also doesn't scale very well: Imagine we add a third or fourth provided DataFetcher. 

What do you think?

Thanks

erikogenvik:
Yeah, the builder methods aren't necessary. However, I liked the way you don't have to specify the name of the property twice. It removes a failure possibility.
Another option would be to have a "DataFetcherStrategy" field in the builder containing an enum. It would default to "Property", but could be set to "Field". If more provided DataFetcher's are added they would be represented by additional enums.

andimarek:
Yes, you are right. Not having to specify the name twice is a good thing. I will merge this now. But probably I will change it to something more generic like your suggested "DataFetcherStrategy" field to make it more stable for other DataFetchers in the future.

Thanks again,
Andi

PR #28, opened 2015-09-18
Add equals/hashCode to schema types

kroepke:
In order to compose a schema from a set of independently constructed types and interfaces, all GraphQLType implementations should provide a equals/hashCode method.

I've chosen the `name` property here, because:
- according to the spec no two types can have the same name
- looking up possible implementations of interfaces relies on .equals, when used with DI reference equality breaks
- several places in SchemaUtil already identify types by their names

SchemaUtil could be converted to use `Set<GraphQLType>` in several places instead of `Map<String, GraphQLType>` but this is not part of this PR.


andimarek:
Thanks for this PR! 

I will comment on specific code issues.

src/main/java/graphql/schema/AbstractGraphQLType.java
```diff
@@ -0,0 +1,32 @@
...
+    private final String name;
+
+    public AbstractGraphQLType(String name) {
+        assertNotNull(name, "name can't null");
+        this.name = name;
+    }
+
+    public String getName() {
+        return name.equals("") ? null : name;
+    }
```
andimarek: We talked about this, the real question is here, if we have to return null values.

src/main/java/graphql/schema/AbstractGraphQLType.java
```diff
@@ -0,0 +1,32 @@
+package graphql.schema;
+
+import static graphql.Assert.assertNotNull;
+
+public class AbstractGraphQLType implements GraphQLType {
```
andimarek: I'm not really about adding AbstractGraphQLType with the only purpose is to not having duplicate

kroepke: I understand, usually I would've turned `GraphQLType` into an abstract class to at least keep the name property and equals/hashcode in the same place, because this really is common behavior of all types, but I didn't want to make that incompatible change all over the code base either.

If you prefer I can easily push the property and method implementations down again.

src/main/java/graphql/schema/AbstractGraphQLType.java
kroepke: I don't think that `GraphQLModifiedType` implementations are ever asked for their name.
Of course that's a little bit hard to see from the interface hierarchy, because the base interface declares `getName()`.

Technically we could return anything, it's never used anywhere. Of course that does not hinder any client to ask for name. The reference implementation uses `GraphQLNamedType` for this, I guess if we did that it would be clearer. Not sure if it's really worth it, though.

yrashk:
It looks like #85 is somewhat related to this PR, unless I am missing something?

andimarek:
Yes, this is the "compare the names in equals" version implementation of #85. 

I would suggest to add two tests for every type, to show that equals and hashCode are working as intended. 

Also equals uses `getName()` while hashCode uses `name` directly. This I think should be changed.

And I think it is only a first step because there might code which uses `==` to compare types instead `equals`.

metacosm:
This fix (or a similar one that properly implements hashCode and equals) is actually needed to properly resolve fragments on union types when using graphql-java-servlet. Currently, `FieldCollector.doesFragmentConditionMatch` calls `FieldCollector.checkTypeCondition` where the resolved type is instantiated by my type resolver whereas the condition type is resolved from the schema. They are semantically identical but different instances thus resulting in `conditionType.equals(type)` failing.

src/main/java/graphql/schema/AbstractGraphQLType.java
metacosm: I favor this introduction of a common abstract type to keep that logic in a single spot and to avoid lots of code duplication.

kroepke:
Just popping in, because I received an email today about a comment I can't find now:

I'm currently not actively using/trying to use GraphQL and have no spare cycles do maintain this.
So if someone wants to take this over, please feel free. I'm not even sure of what the state of Java-GraphQL is at the moment.

dminkovsky:
@kroepke cool, thanks for checking in!

@metacosm

> calls FieldCollector.checkTypeCondition where the resolved type is instantiated by my type resolver:

Yes, this issue is mentioned at the bottom of https://github.com/graphql-java/graphql-java/issues/205#issue-176456291: it would be useful it the TypeResolver interface provided an instance of the schema in the resolution method. Otherwise, you need the schema somehow available in the TypeResolver's scope, which might require more work on your part. 

But yes, as you see, the schema definition system as it stands is [all about simplicity and therefore uses reference equality/identity](https://github.com/graphql-java/graphql-java/issues/85#issuecomment-217361404). Not great, but at least simple. 

I, too, would prefer a [value-based](http://docs.oracle.com/javase/8/docs/api/java/lang/doc-files/ValueBased.html) schema system. But, I am not really comfortable with a hashCode/equals that's based entirely on one field like "name".

andimarek:
I am closing this Issue: The main problem here is that a schema contains not only data but also logic (`DataFetcher`, `TypeResolver`). This is something we can't easily change for now, so we will not add `equals/hashcode` to them.

PR #29, opened 2015-09-18
Fit with Relay spec and add some tests.

erikogenvik:


andimarek:
Thanks.

And especially for the tests!

PR #32, opened 2015-09-19
Allow keywords anywhere 'Name' is used.

zourzouvillys:
The GraphQL specification (and reference implementation) doesn't restrict the keywords 'query' or 'fragment' as Names.  Because of ANTLRs behaviour, input with arguments, field names, or types with either of these values were being rejected, e.g:

```
query Something {

  search(query: "theo") {
    ...
  }

}
```

This patch updates the grammar and the conversion to allow these keywords.

Note that the gradle build scripts do not seem to be automatically updating the generated source code, so i manually regenerated it.


andimarek:
Thanks for this PR.

I have a question: Shouldn't we do this also  for `mutation`? I think this is "restricted" in the same way currently ...

zourzouvillys:
Yeah - along with 'on', 'true', and 'false' ... although, the GraphQL grammar isn't context free - so we can't always allow a Name where it should be allowed with ANTLR (at least - my knowledge of it - i've not used it since 3.x, and it would have been very hard if not impossible in it then).

Sigh; The updated patch does the best i can without some major work on the ANTLR grammar, or hand-rolling a parser. Specifically;

- 'true' and 'false', and 'on' can't be used as parameter names, nor as a field name or alias.
- 'on', 'true', or 'false' can't be used as a fragment name.

danielkwinsor:
Well great, you made me read the appendices of the spec, something I avoided until now, lol.

Yes, the spec is pretty clear in that practically anything can be a name.  Given the stated goal of keeping it simple I think they wouldn't care if keywords were used, but at the same time I'm not sure whether this was intentional or not.  I'll ask.

danielkwinsor:
Also, there are a lots of line changes with either name changes or removing @NotNull.  Is it possible to create a patch that avoids this?  There is a conflict anyways.

danielkwinsor:
I ran into this; I accidentally named a field called "query".  Problem was, graphiql was the thing complaining about a parse error, not graphql-java.  So unless most tools/libraries have this support, it may be premature to put this in?

dminkovsky:
Hey @zourzouvillys, thank you for this PR. Sorry it took so long to be resolved! Following up with some updates to the grammar in #243, I've incorporated your concept here in to #277. Will close this and merge that. Thanks again.

PR #33, opened 2015-09-19
Allow providing additional types to schema build.

zourzouvillys:
With the current API, only concrete instances which are registered somewhere from the query or mutation root are registered.  This makes writting tools to automatically generate types a PITA - specifically because you have to ensure there is at least one concrete instance of every type registered.

This change provides a way to pass a set of types into the schema when building it, which will be added into the set of available types when replacing type references.

An alternative could be to provide a resolver to GraphQLTypeReference, which would be used to resolve the type at build time - however, this method seems far less intrusive for existing consumers :-)

### DIFF
--- a/src/main/java/graphql/schema/GraphQLSchema.java
+++ b/src/main/java/graphql/schema/GraphQLSchema.java
@@ -5,10 +5,12 @@ import static graphql.Assert.assertNotNull;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import graphql.Assert;
 import graphql.Directives;
 
 public class GraphQLSchema {
@@ -19,7 +21,7 @@ public class GraphQLSchema {
     private Set<GraphQLType> dictionary;
 
     public GraphQLSchema(GraphQLObjectType queryType) {
-        this(queryType, null, null);
+        this(queryType, null, Collections.<GraphQLType>emptySet());
     }
 
     public Set<GraphQLType> getDictionary() {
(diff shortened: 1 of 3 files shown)


andimarek:
Thanks!

Is this PR related to this one https://github.com/andimarek/graphql-java/pull/28 in your opinion?

To understand the use-case better: Do I understand this right, that you generate the schema automatically and sometimes you have for example a InterfaceType and a some  ObjectTypes implementing this Interface, but nobody is referencing the ObjectType, just the Interface?

zourzouvillys:
It's not directly related to #28.

re use case: yes.  the one that bit me was using a TypeReference everywhere a given type if mentioned  is used (e.g, in parameters or return types), and so when it comes to building the schema, there is no reference to the actual ObjectType instance.

src/main/java/graphql/schema/GraphQLSchema.java
```diff
@@ -80,11 +86,16 @@ public Builder mutation(GraphQLObjectType mutationType) {
         }
 
         public GraphQLSchema build() {
-            GraphQLSchema graphQLSchema = new GraphQLSchema(queryType, mutationType);
-            new SchemaUtil().replaceTypeReferences(graphQLSchema);
-            return graphQLSchema;
-        }
+          return build(null);
```
andimarek: I would prefer not to use null ... just a empty set

src/main/java/graphql/schema/SchemaUtil.java
```diff
@@ -94,18 +95,23 @@ private void collectTypesForObjects(GraphQLObjectType objectType, Map<String, Gr
...
     }
 
-    public Map<String, GraphQLType> allTypes(GraphQLSchema schema) {
+    public Map<String, GraphQLType> allTypes(GraphQLSchema schema,  Set<GraphQLType> dictionary) {
         Map<String, GraphQLType> typesByName = new LinkedHashMap<>();
         collectTypes(schema.getQueryType(), typesByName);
         if (schema.isSupportingMutations()) {
             collectTypes(schema.getMutationType(), typesByName);
         }
+        if (dictionary != null) {
```
andimarek: assuming that dictionary is never null would make this simpler

andimarek:
ok ... I understand .... interesting approach to use TypeReference as default.

I comment on some minor issues I thought could be improved, but nothing seriously.

zourzouvillys:
pushed and update re null instead of empty set usage!

andimarek:
Thanks

PR #34, opened 2015-09-25
GraphQLTypeReference is also a GraphQLInputType.

zourzouvillys:


andimarek:
Quick Thought: I think have I excluded InputType on purpose, because TypeReferences are not always replaced with the real object ... currently you can't use them just everywhere (I think).

 I must look at it in detail.

petervelosy:
Dear Andi,

Do you have any updates about this? I currently have a use case where I would need this functionality (see https://github.com/andimarek/graphql-java/issues/46 ). 

Thanks!
Peter

danielkwinsor:
Right, given the concern that TypeReferences might be handled in a special way, can you provide in your PR some unit tests demonstrating that nothing breaks and that input refs cannot mix with output refs?

dminkovsky:
Closing this in favor of #204, which is more comprehensive.

PR #42, opened 2015-10-20
Supporting null values for complex input fields on mutations.

dieterwachters:


lelong71:
I run into the same issue and wonder why this patch has not been accepted yet.

andimarek:
@lelong71 Thanks for this info. Please try again with the latest version, this should be fixed by now.

lelong71:
Awesome! Thanks for the quick response. I'll pull the latest version and try against my use case.

lelong71:
I pulled the latest code and still run into the same issue. The input object in ValueResolver is still null for some input fields in my case (nested objects and some inputs are not passed in vs passing null). I had to add an NPE check for input to get it to work just like @dieterwachters suggested.

   private Object coerceValueForInputObjectField(GraphQLInputObjectType inputObjectType, Map<String, Object> input) {
        Map<String, Object> result = new LinkedHashMap<>();
        for (GraphQLInputObjectField inputField : inputObjectType.getFields()) {

            if (input == null) {
                return null;
            }
            Object value = coerceValue(inputField.getType(), input.get(inputField.getName()));
            result.put(inputField.getName(), value == null ? inputField.getDefaultValue() : value);

        }
        return result;
    }

andimarek:
The described problem is solved in the latest dev build. I solved it slightly different: https://github.com/andimarek/graphql-java/blob/f56171f47a050bd1f35ce7d28109c7ff685a3ccf/src/main/java/graphql/execution/ValuesResolver.java#L66 

Because this PR is now obsolete I'm closing it.

PR #44, opened 2015-10-20
Fix example for GraphQLUnionType

henrikgs:
Also, static import `newUnionType` method in `GarfieldSchema.java` to make it consistent with the example and other imports in GarfieldSchema.


andimarek:
Thanks for fixing this!

PR #48, opened 2015-11-11
Recursively adding input types

cardinalraven:


cardinalraven:
Oh, editor dropped some crap we don't want, will resubmit

andimarek:
Thanks for this PR!

Do I understand this correct that this is BugFix and currently input types are not collected when they are not top level input types in the Schema?

cardinalraven:
Your understanding is correct.  The unit test I included in SchemaUtilTest will fail if you undo my change to SchemaUtil (it won't get the Range type).

andimarek:
Thanks .... I merge it.

PR #60, opened 2015-12-30
Fix two NPEs and add test case

xuzb:


andimarek:
Thanks for your PR ... looks fine to me ... I merge it now.

Issue #65, opened 2016-01-21
Support for recursive Objects


nmn:
Very often there is a need for Nested structures in data models.

Example 1: Comments:

```
GraphQLObjectType commentType = newObject()
                        .name("Comment")
                        .field(newFieldDefinition()
                                .type(new GraphQLList(commentType))
                                .name("replies")
                                .build())
                        .build();
```

Example 2: Friends:

```
GraphQLObjectType personType = newObject()
                        .name("Person")
                        .field(newFieldDefinition()
                                .type(new GraphQLList(commentType))
                                .name("friends")
                                .build())
                        .build();
```

These fairly common use-cases current seem impossible with GraphQL-Java. This is because the type field cannot be null. GraphQL-js gets around this problem by letting you define your fields as a function:

```
var CommentsType = new GraphQLObjectType({
  name: 'Comments',
  fields: () => ({
    humanName: {type: GraphQLString},
    replies: {type: new GraphQLList(GreetingsType)}
  }),
});
```

Is something like this possible with GraphQL-java? If not can it be added??

I recommend, overloading the `.type` to accept a lambda Function. I'll try to make a pull request, that fill probably not work, but with your help it may be possible to make it work.

andimarek:
I think this solves your problem:

https://github.com/andimarek/graphql-java#recursive-type-references

nmn:
This is awesome. Thanks!

PR #66, opened 2016-01-21
overload .type to accept a callable/lambda

nmn:
Addresses #65

I'm basically no good at Java, so this may have lots of problems with it. Conceptually this does make some sense to me.

As I can tell, Lambda implement something called FunctionalInterfaces. DataFetcher is one such FunctionalInterface that you have defined already.

`Java.util.concurrent.Callable` is also a functional interface. It maps to a lambda that takes no arguments but returns a value.

Therefore `Callable<GraphQLOutputType>` should be a lambda that returns a `GraphQLOutputType`.

Conceptually, this should let us do something like this to create recursive Object types.

```
GraphQLObjectType commentType = newObject()
                        .name("Comment")
                        .field(newFieldDefinition()
                                .type(() -> new GraphQLList(commentType))
                                .name("replies")
                                .build())
                        .build();
```

As I said, I'm no good at Java, and there may be issues, but I do hope that this helps add support for recursive Objects. We need it badly.

### DIFF
--- a/src/main/java/graphql/schema/GraphQLFieldDefinition.java
+++ b/src/main/java/graphql/schema/GraphQLFieldDefinition.java
@@ -4,6 +4,7 @@ package graphql.schema;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
 
 import static graphql.Assert.assertNotNull;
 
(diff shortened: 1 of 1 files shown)


nmn:
Apparently there's an error as I didn't account for the fact that Runnables can throw. I'm not sure what the solution here is. But I'll keep looking.

andimarek:
Closes, because solved already by TypeReferences. But thanks for creating a PR!

PR #67, opened 2016-01-21
Support for only backward or forward pagination arguments

tinnou:
To match the spec: _(...) A field that returns a Connection Type must include **forward pagination arguments**, **backward pagination arguments**, or both. (...)_


andimarek:
Thanks!
The relay part is not the nicest thing currently ... so thanks for improving that.

tinnou:
No kudos to you for building the Java implementation!

PR #76, opened 2016-01-30
Offer a method to specify field name of the target

maquessime:
Currently, if a POJO has a field firstName for instance and you want to query with "first_name" instead of "firstName", you need to provide explicitly a data fetcher to do the map. That could be not quite cool to explicitly define a data fetcher for each field just to do this kind of map.

I am just proposing a new method on the builder to specify a field name to map with the name of the field in the schema. No big deal ;)

### DIFF
--- a/src/main/java/graphql/schema/AbstractReflectionDataFetcher.java
+++ b/src/main/java/graphql/schema/AbstractReflectionDataFetcher.java
@@ -1,0 +1,34 @@
+package graphql.schema;
+
+import java.util.Map;
+
+public abstract class AbstractReflectionDataFetcher
+    implements DataFetcher {
+
+    protected final String name;
+
+    protected AbstractReflectionDataFetcher(
+        String name) {
+
+        this.name = name;
+
+    }
+
+    @Override
+    public final Object get(
+        DataFetchingEnvironment environment) {
+
+        Object source = environment.getSource();
+        if (source == null)
+            return null;
+        if (source instanceof Map) {
+            return ((Map<?, ?>) source).get(name);
+        }
+        return getValue(source, environment.getFieldType());
+    }
+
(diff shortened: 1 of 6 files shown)


tberman:
So, obviously up to @andimarek, but to me, this patch represents an example of an open source etiquette issue. To read up: http://tirania.org/blog/archive/2010/Dec-31.html

maquessime:
Sorry for the formatting change, It's reverted now, was not wanted of course.

I could revert  the "refactor" commit too for the data fetchers, sorry, it's just an habit to refactor before adding something new.

The only change could have been nameAndFieldName method and tests.

danielkwinsor:
My position is somewhere between the 2.  I always refactor my own code, and I'll go so far as to reformat other people's code too.  But in the specifics here, we have a merge conflict.  So could I ask this: create 2 PRs, 1 with the minimalist changes necessary to support your feature, and if that looks good, a second PR that does the refactor.

Thing is, at this time I'm not sure whether nameAndFieldName is suitable on `GraphQLFieldDefinition.Builder`.  I'd prefer to see something clearer like
```
.name("foo")
.fieldToFetch("bar")
```
instead of
```
.nameAndFieldName("foo", "bar")
```
But even this is getting hairy when considering PropertyDataFetcher and other non-default fetchers.  Field-name-to-fetch is really a property of the FieldDataFetcher only (and after a PR, PropertyDataFetcher too).

Is there another way that you can make this change, or can you live with
```
.name("foo")
.dataFetcher(new FieldDataFetcher("bar")
```

maquessime:
@danielkwinsor I think we can live with 

```
.name("foo")
.dataFetcher(new FieldDataFetcher("bar")
```

I close the pull request.

PR #80, opened 2016-02-04
Fix NPE when non-required input fields are missing

lelong71:
There're two fixes in this pull request:

1/ If you don't send all non-required input fields for an input object, it'll throw exception. This fix would just ignore non-required input fields if they're missing
2/ Validate required fields in input object


andimarek:
Thanks a lot for your PR.

But I decided to solve a bit different than in your PR and I also wanted to add some test. So unfortunately this PR is obsolete and I'm closing it now.

PR #81, opened 2016-02-05
Change sourceCompatibility to 1.6.

pt-achang:
We’d like to target using your graphql-java library on some legacy systems that run JDK 1.6 (not for great technical reasons, but this is how certain company / government bureaucracies work). I was wondering if you’d be willing to accept a patch which would change your “sourceCompatibility” to 1.6?

The changes in the patch entail:

(1) Changing all usage of the diamond operator (<>) to use the fully-specified type, e.g.:

becomes:

                Map<String, GraphQLType> typesByName = new LinkedHashMap<String, GraphQLType>();

(2) Unrolling multi-catch statements (e.g., “catch (IllegalAccessException | InvocationTargetException e)”). This only happens in a couple places.

In other words, I think the changes here are mostly cosmetic, but it would greatly increase our ability to roll your awesome library out to more places!

Thanks for your consideration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ plugins {
 apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'maven-publish'
+apply plugin: 'idea'
 
 sourceCompatibility = 1.6
 def releaseVersion = System.properties.RELEASE_VERSION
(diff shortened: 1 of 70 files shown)


andimarek:
Thanks for this PR and thanks to you effort.

About supporting Java 1.6 in general: I'm thinking about it now for some time I don't really now yet, if we should do this. Because even Java 7 is EOL, and I would rather to prefer to switch to Java 8 in the near future.
But of course there are still projects out there which use Java 1.6 (or even older).

I created a entry in the mailing-list to gather some feedback: https://groups.google.com/forum/#!topic/graphql-java/iCCN2T1RgyI 


About this PR specifically: I think the second the second commit doesn't belong in this change.

Thanks,
Andi

pt-achang:
Closing this in favor of https://github.com/andimarek/graphql-java/pull/87. This PR is based off the wrong branch and contains extraneous commits.

PR #84, opened 2016-02-09
Fix invalid javadoc in BatchedExecutionStrategy

loganlinn:
Some erroneous HTML tags were added in b603ffa that cause the Gradle javadoc
task to fail with errors.


andimarek:
Thanks for fixing this!

PR #87, opened 2016-02-15
Change sourceCompatibility to 1.6.

pt-achang:
Properly created version of https://github.com/andimarek/graphql-java/pull/81 (this PR uses the right source branch).

We’d like to target using your graphql-java library on some legacy systems that run JDK 1.6 (not for great technical reasons, but this is how certain company / government bureaucracies work). I was wondering if you’d be willing to accept a patch which would change your “sourceCompatibility” to 1.6?

The changes in the patch entail:

(1) Changing all usage of the diamond operator (<>) to use the fully-specified type, e.g.:

becomes:

                Map<String, GraphQLType> typesByName = new LinkedHashMap<String, GraphQLType>();

(2) Unrolling multi-catch statements (e.g., “catch (IllegalAccessException | InvocationTargetException e)”). This only happens in a couple places.

In other words, I think the changes here are mostly cosmetic, but it would greatly increase our ability to roll your awesome library out to more places!

Thanks for your consideration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'maven-publish'
 
-sourceCompatibility = 1.7
+sourceCompatibility = 1.6
 def releaseVersion = System.properties.RELEASE_VERSION
 version = releaseVersion ? releaseVersion : new SimpleDateFormat('yyyy-MM-dd\'T\'HH-mm-ss').format(new Date())
 group = 'com.graphql-java'
--- a/src/main/java/graphql/ExecutionResultImpl.java
+++ b/src/main/java/graphql/ExecutionResultImpl.java
@@ -7,7 +7,7 @@ import java.util.Map;
 
 public class ExecutionResultImpl implements ExecutionResult {
 
-    private final List<GraphQLError> errors = new ArrayList<>();
+    private final List<GraphQLError> errors = new ArrayList<GraphQLError>();
     private Object data;
 
     public ExecutionResultImpl(List<? extends GraphQLError> errors) {
@@ -37,7 +37,7 @@ public class ExecutionResultImpl implements ExecutionResult {
 
     @Override
     public List<GraphQLError> getErrors() {
-        return new ArrayList<>(errors);
+        return new ArrayList<GraphQLError>(errors);
     }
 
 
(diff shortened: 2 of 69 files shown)


pt-achang:
From the Andi:

> Thanks for this PR and thanks for your effort.
> 
> About supporting Java 1.6 in general: I'm thinking about it now for some time I don't really now yet, if we should do this. Because even Java 7 is EOL, and I would rather to prefer to switch to Java 8 in the near future.
> But of course there are still projects out there which use Java 1.6 (or even older).
> 
> I created a entry in the mailing-list to gather some feedback: https://groups.google.com/forum/#!topic/graphql-java/iCCN2T1RgyI

andimarek:
@pt-achang Thanks for your effort. I decided to merge it. So please test it with the newest build and  open a new Issue if you encounter any problems.

Thanks.

pt-achang:
Thanks!

corydolphin:
@andimarek I'd love to re-open this conversation if you are interested.

I think it would help the project to bump the source version to the much more modern Java 8. This would allow us to remove the reflection based PropertyBasedDataFetcher, and instead simply use Java's native Method references. 

E.G. PropertyDataFetcher("booleanFieldWithGet") could simply be replaced with Object::booleanFieldWithGet.

GrigoryPtashko:
@corydolphin @andimarek Guys, what are the plans for moving to sourceCompatibility to 1.8? We are using Java 7 in production and don't plan to move to 8. I work in the biggest russian touroperator. We've got a very high loaded system and very-very bug codebase and just started to use graphql. It is simply impossible to move to Java 8 and the moment. I'll be very sad if the whole graphql-java project switch to Java 8 :(

dminkovsky:
Hi Grigory. There are no such plans at this time. There is interest in a
Java 8 secondary artifact, but not changing the core. Thanks for posting.


пт, 23 дек. 2016 г. в 2:21, Grigory Ptashko <notifications@github.com>:

dminkovsky:
I've been thinking about making the default data fetcher configurable.
Would that help?

вс, 18 дек. 2016 г. в 2:20, Cory Dolphin <notifications@github.com>:

corydolphin:
@dminkovsky @GrigoryPtashko We can achieve what I desired without needing the library to use src/target of Java 8. Apologies for my confusion.

dminkovsky:
@corydolphin great, good to hear.

Issue #24, opened 2015-09-09
Javadoc


weebl2000:
I really like using graphql, but I am really struggling with this implementation because there is so little javadoc. For example, how should I define my own types with coerce() and coerceLiteral() methods? I can only gather so much information from the code itself.

andimarek:
I'm sorry to hear that so much is unclear.
But you're right: To really understand the details you have to read the source code.

I can't really promise anything when the documentation will be improved: I'm doing this is my free time and my resources are limited. 

But thanks for raising awareness.

Tip: This is a more or less direct port the graphql JavaScript Implementation. So nearly every concept and some documentation about it can be found in [here](https://github.com/graphql/graphql-js) and [here](http://facebook.github.io/graphql).

Again: I'm sorry that I can't give you a better answer at this time. Feel free to open a new Issue for specific questions.

Best regards,
Andi

andimarek:
I will let this Issue be open, as long as the documentation isnt't improved.

danielkwinsor:
Would you accept pull requests of javadoc that others write?  I might want to write some javadoc myself, to aid my learning and perhaps any developers using this.

andimarek:
Yes, of course. Thanks for your help.

andimarek:
@danielkwinsor about your last comment in #89:  
My personal philosophy for source code comments is very puristic: I add rarely javadoc comments. For example I don't think adding author infos is useful (use git for getting the history of the file)
Instead I try to make the source code as readable as possible plus adding tests, which also document whats going on.

I also differentiate between public classed and private implementation details: I think the javadoc for public classes should be improved (e.g. the `GraphQL` or `Relay` class). On the other hand implementation details should not be documented in the same way. Because they are subject to change and if you really want to know whats going on you should read the source-code.

danielkwinsor:
I won't add author fields.  How about version?

When I develop something, I want to use my IDE to use context-assist and to read the javadoc, thereby figuring out from the API what is possible.  That workflow doesn't work here, instead I go to the Readme.md and to tests.
One feedback I have is that, for a given feature, it can be hard to look for example usage in the tests when the schema and groovy test are split among differently named files.  I do understand the background here of Garfield and Star Wars, I'm just saying.
[...]

PR #89, opened 2016-02-19
Adding autogenerated javadoc for versions v1.0, v1.2, v1.3

danielkwinsor:
Andi,
I have autogenerated as much javadoc as the tools could reach, although it appears some classes were missed.
Everything is tagged with your name and versions v1.0, v1.2, v1.3 as applicable.
New classes since the v1.3 tag have had their @version erased.  That's basically anything in graphql.execution.batched.*
From this basis we all should be able to assist in focusing on the important sections of javadoc.

I am now reading on git author/committer, and how to set that, in case you want to maintain history as last edited by you.

FYI, the way I did it was turn this into a maven project then run maven-javadoc-plugin's javadoc:fix

### DIFF
--- a/src/main/java/graphql/Assert.java
+++ b/src/main/java/graphql/Assert.java
@@ -1,9 +1,21 @@
 package graphql;
 
 
+/**
+ * <p>Assert class.</p>
+ *
+ * @author Andreas Marek
+ * @version v1.0
+ */
 public class Assert {
 
 
+    /**
+     * <p>assertNotNull.</p>
+     *
+     * @param object a {@link java.lang.Object} object.
+     * @param errorMessage a {@link java.lang.String} object.
+     */
     public static void assertNotNull(Object object, String errorMessage) {
         if (object != null) return;
         throw new AssertException(errorMessage);
--- a/src/main/java/graphql/AssertException.java
+++ b/src/main/java/graphql/AssertException.java
@@ -1,8 +1,19 @@
 package graphql;
 
 
+/**
+ * <p>AssertException class.</p>
+ *
+ * @author Andreas Marek
+ * @version v1.0
+ */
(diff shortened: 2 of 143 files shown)


andimarek:
Thanks for this RP and I really appreciate it a lot that you are willing to improve graphql-java.

But unfortunately I will not merge this PR: Imho almost all of this auto generated javadoc comments have no value at all. They are even worse than having no JavaDoc, because they just clutter the source code with useless informations.

That said, I'm not against javadoc in general, and I'm aware that it would be nice to have more documentation, but not in this way.

I would recommend to add specific manual written documentation to some classes/methods if you are still willing to help.

Again: I'm sorry that you invested your time and I know that it's never nice to have a PR rejected. 

Thanks,
Andi

danielkwinsor:
Andi,
That's not a problem.  I respect that useless doc is worse than no doc.  I cringed at some of the javadoc generated for constants in the lexer and parser, for instance, and a second pass by hand would remove those.  And for instance, it pisses me off to see commented out lines of code, and I delete those every chance I get, so I'm with you there.

Anyways, I have gone through this myself and I feel like it was a very wrong idea to autogenerate method level javadoc.  I did not realize how bad it was.  "@param value a {@link java.lang.Object} object."  Yeah, thanks, really helpful... now tell me WHY the API was Object instead of some more specific type... that's what should go in.

What do you think of class level autogenerated javadoc only, with the version/author info?  I can either submit auto-class for all, or generate auto-class and submit PRs as classes get handcrafted.  I would be in favor of the first one.

Yes, I still intend to do method level javadoc, and will auto-gen it first for ease, but will submit only that which has been through a hand-crafted process.

andimarek:
I'm closing this PR, because it will not get merged. Let's discuss general javadoc questions here: #24

PR #91, opened 2016-02-25
Java Generics for clarity

fredrik-hultin:
By using generics it is easier to understand how to use the API. It also helps with finding type errors.

I also changed the versioning in the build file to fit the SNAPSHOT strategy from Maven. 
Minor change to disable doclint for java 1.8 builds

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,8 @@ apply plugin: 'maven-publish'
 sourceCompatibility = 1.6
 targetCompatibility = 1.6
 def releaseVersion = System.properties.RELEASE_VERSION
-version = releaseVersion ? releaseVersion : new SimpleDateFormat('yyyy-MM-dd\'T\'HH-mm-ss').format(new Date())
+//version = releaseVersion ? releaseVersion : new SimpleDateFormat('yyyy-MM-dd\'T\'HH-mm-ss').format(new Date())
+version = releaseVersion ? releaseVersion : '2.0.0-SNAPSHOT'
 group = 'com.graphql-java'
 
 
--- a/build.gradle
+++ b/build.gradle
@@ -115,3 +115,11 @@ task wrapper(type: Wrapper) {
     gradleVersion = '2.11'
     distributionUrl = "http://services.gradle.org/distributions/gradle-${gradleVersion}-all.zip"
 }
+
+if (JavaVersion.current().isJava8Compatible()) {
+    allprojects {
+        tasks.withType(Javadoc) {
+            options.addStringOption('Xdoclint:none', '-quiet')
+        }
+    }
+}
--- a/src/main/java/graphql/schema/Coercing.java
+++ b/src/main/java/graphql/schema/Coercing.java
@@ -1,7 +1,7 @@
 package graphql.schema;
 
 
-public interface Coercing {
+public interface Coercing <T2> {
 
 
     /**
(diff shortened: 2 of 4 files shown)


danielkwinsor:
Hey Fredrik,
I think it'll be fine to add generics, but there's a conflict right now.  Can you redo this, but also break it up into multiple PRs?  One for Coercing, another for DataFetcher, and I'm not yet ready to merge in any changes to the build, which itself would still need to be another PR.

markphilpot:
I will pick up the torch on this one.

aschrijver:
Ha Fredrik, you sneakily put in a different versioning scheme convention under this PR title, I see :open_mouth: :wink: . 
While I am all agreeing to that, as its cleaner and more 'rememberable' we should really ask @andimarek for that, and also ideally make it a separate PR. There may be a valid reason for the current scheme, like e.g. continuous delivery support (I dunno, just guessing).

build.gradle
```diff
@@ -11,7 +11,8 @@ apply plugin: 'maven-publish'
 sourceCompatibility = 1.6
 targetCompatibility = 1.6
 def releaseVersion = System.properties.RELEASE_VERSION
-version = releaseVersion ? releaseVersion : new SimpleDateFormat('yyyy-MM-dd\'T\'HH-mm-ss').format(new Date())
+//version = releaseVersion ? releaseVersion : new SimpleDateFormat('yyyy-MM-dd\'T\'HH-mm-ss').format(new Date())
+version = releaseVersion ? releaseVersion : '2.0.0-SNAPSHOT'
```
andimarek: This is used for deploying the newest build on bintray, by using a constant string we would always override the latest version, which is not what I want. The date is unique and clear for everybody who want's to use the latest (or one specific) build.

build.gradle
```diff
@@ -115,3 +116,11 @@ task wrapper(type: Wrapper) {
...
     distributionUrl = "http://services.gradle.org/distributions/gradle-${gradleVersion}-all.zip"
 }
+
+if (JavaVersion.current().isJava8Compatible()) {
+    allprojects {
+        tasks.withType(Javadoc) {
+            options.addStringOption('Xdoclint:none', '-quiet')
+        }
+    }
+}
```
andimarek: I am not sure what this does. Please open a separate PR for that, if this change was intentional.

src/main/java/graphql/schema/Coercing.java
```diff
@@ -1,7 +1,7 @@
 package graphql.schema;
 
 
-public interface Coercing {
+public interface Coercing <T2> {
```
andimarek: I would highly suggest to use T (or K,V ..) instead T2. T2 is a very unusual Type Parameter name.

andimarek:
@fredrik-hultin Thanks for your PR. I commented on some specific aspects in the commit (including the versioning, thanks @aschrijver for pinging)

build.gradle
andimarek: I am sorry, I have overlooked your comment about that change. But please still open a separate PR for that .... thanks a lot.

markphilpot:
If @fredrik-hultin doesn't mind, I'll break this up into a few PRs.  I'll give @fredrik-hultin a few days to check back since this PR was started back in February.

fredrik-hultin:
Sorry guys, lost this for a while. Let me get a chance to have a look at this soon again... I'll get back to you soon if this is still interesting?

markphilpot:
I can also take the lead if the context has been swapped out =) Let me know.

dminkovsky:
Closing this in favor of #280, which includes all of the generics-related changes in this PR. Thank you @fredrik-hultin an reviewers.

PR #94, opened 2016-02-28
Support IntValue tokens with larger value than INT_MAX

okorz001:
IntValue now stores BigInteger instead of int. GraphQLInt and
GraphQLLong parseLiteral will throw if the value overflows int or long
respectively.

StringValue parsing for GraphQLLong is still supported for backwards
compatibility.

Add a test for parsing GraphQLLong larger than INT_MAX.

### DIFF
--- a/src/main/java/graphql/Scalars.java
+++ b/src/main/java/graphql/Scalars.java
@@ -1,6 +1,8 @@
 package graphql;
 
 
+import java.math.BigInteger;
+
 import graphql.language.BooleanValue;
 import graphql.language.FloatValue;
 import graphql.language.IntValue;
@@ -31,7 +33,10 @@ public class Scalars {
         @Override
         public Object parseLiteral(Object input) {
             if (!(input instanceof IntValue)) return null;
-            return ((IntValue) input).getValue().intValueExact();
+            BigInteger value = ((IntValue) input).getValue();
+            // Check if out of bounds.
+            Integer.parseInt(value.toString());
+            return value.intValue();
         }
     });
 
@@ -60,7 +65,10 @@ public class Scalars {
             if (input instanceof StringValue) {
                 return Long.parseLong(((StringValue) input).getValue());
             } else if (input instanceof IntValue) {
-                return ((IntValue) input).getValue().longValueExact();
+                BigInteger value = ((IntValue) input).getValue();
+                // Check if out of bounds.
+                Long.parseLong(value.toString());
+                return value.longValue();
             }
             return null;
         }
--- a/src/main/java/graphql/Scalars.java
+++ b/src/main/java/graphql/Scalars.java
@@ -31,7 +31,7 @@ public class Scalars {
         @Override
         public Object parseLiteral(Object input) {
             if (!(input instanceof IntValue)) return null;
-            return ((IntValue) input).getValue();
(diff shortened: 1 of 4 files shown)


okorz001:
Ah, it appears BigDecimal.intValueExact was added in Java 8. I will find a backwards compatible solution for Java 6.
https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#intValueExact--

andimarek:
I don't think this is the right way: The graphql language only supports Integers on the language level. And this library aims to conform to that.

To quote the spec:
```
Numeric integer values larger than 32‐bit should either use String or a custom‐defined Scalar type, as not all platforms and transports support encoding integer numbers larger than 32‐bit.
```

See also this Issue: https://github.com/facebook/graphql/issues/73

okorz001:
I think there is a distinction here between the query grammar and the runtime types.

The grammar does not specify a maximum number of digits in an IntValue token:
http://facebook.github.io/graphql/#sec-Int-Value

However, the built-in type Int is required to be a 32-bit integer. This PR retains this behavior.
http://facebook.github.io/graphql/#sec-Int

I believe it should be possible for users to register custom scalars (which includes the provided GraphQLLong) which can be parsed from IntValue with an arbitrary number of digits.

The Javascript reference implementation keeps all token values as strings until they are parsed into type instances. Here is parseLiteral for IntValue:
https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L38-L46

Proof of concept:

```js
var graphql = require('graphql');

var longType = new graphql.GraphQLScalarType({
  name: 'Int',
  description: '64-bit integral numbers',
  // TODO: Number is only 52-bit
  serialize: Number,
  parseValue: Number,
  parseLiteral: function parseLiteral(ast) {
    if (ast.kind === graphql.Kind.INT) {
      const num = parseInt(ast.value, 10);
      return num;
    }
    return null;
  }
});

var queryType = new graphql.GraphQLObjectType({
  name: 'Query',
  fields: function fields() {
    return {
      foo: {
        type: longType,
        args: {
          bar: {
            type: longType,
            description: 'The value to return'
          },
        },
        resolve: function getFoo(root, args) { return args.bar || 0 }
      }
    } 
  }
});

var schema = new graphql.GraphQLSchema({query: queryType});

graphql.graphql(schema, '{foo(bar:2147483648)}').then(function(err, data) {
  if (err) return console.error(err);
  console.log(data);
});
```

```
$ node index.js
{ data: { foo: 2147483648 } }
```

andimarek:
You have a point here. I don't think this should work, but I maybe wrong. To clarify this I asked for help: https://github.com/facebook/graphql/issues/73#issuecomment-190310810

andimarek:
You were right (https://github.com/facebook/graphql/issues/73#issuecomment-190322457). This means graphql-java must be changed to support this.

okorz001:
Do you think we should keep the token values as string until they are parsed into types? A downside of this change is unnecessary BigInteger instances being created and discarded.

andimarek:
I don't think this is a problem: Creating BigIntegers  is the right thing todo.

drkarl:
:+1: LGTM

andimarek:
This is merged now. Thanks @okorz001

PR #95, opened 2016-03-04
Builder syntax

Underzenith85:
Cleaned up the builder syntax to allow for lambda's in JDK8, cuts down on the number of .build() calls by wrapping the builder in the field and argument methods.

Fixed long to int coercion. Long values can be defined in a schema, but attempting to use it will cause an exception since it's being stored as an int primitive via Integer.parseInt.

Deleted generated antlr files and moved generation to a gradle task.


Underzenith85:
Oops, meant to merge this back into my own master for now.

PR #99, opened 2016-03-07
Throwable cause instead of Exception

okorz001:
Some JDK APIs such as CompletableFutures generate Throwables instead of
Exceptions.

Exception has no additional methods than Throwable, and all tests are passing.


andimarek:
Thanks for this PR. I understand why you want to change this. But you didn't change any code using this Exception. I'm curious: Is this for a 3rd-party project?

andimarek:
Thanks .. I'm going to merge this.

okorz001:
Sorry, I forgot to get back to you. I am using a custom execution strategy that I currently do not have permission to publish. Thanks for merging!

PR #113, opened 2016-03-22
Support parsing IntValue literals into GraphQLFloat

Macroz:
Hello!

We noticed a problem with the GraphQLFloat data type. If you send a literal number e.g. `2`, it will be parsed as IntValue and therefore throw an exception in the Scalars.parseLiteral for floats. There is a workaround to send e.g. `2.00001` but probably integers should be accepted here.

-Markku
https://github.com/HSLdevcom/digitransit-ui


andimarek:
Thanks for this PR. You are right: This should be changed. I'm merging this.

Issue #114, opened 2016-03-25
NPE when coercing null enum value in BatchedExecutionStrategy


tinnou:
Test here yields the below error. https://github.com/tinnou/graphql-java/commit/359a8e619ee217fe0ca0ae4bbe7bb8607ac024b2

Stack trace: 
```java.lang.NullPointerException
	at graphql.schema.GraphQLEnumType.getNameByValue(GraphQLEnumType.java:49)
	at graphql.schema.GraphQLEnumType.access$000(GraphQLEnumType.java:14)
	at graphql.schema.GraphQLEnumType$1.serialize(GraphQLEnumType.java:23)
	at graphql.execution.batched.BatchedExecutionStrategy.coerce(BatchedExecutionStrategy.java:219)
	at graphql.execution.batched.BatchedExecutionStrategy.handlePrimitives(BatchedExecutionStrategy.java:212)
	at graphql.execution.batched.BatchedExecutionStrategy.completeValues(BatchedExecutionStrategy.java:91)
	at graphql.execution.batched.BatchedExecutionStrategy.resolveField(BatchedExecutionStrategy.java:78)
	at graphql.execution.batched.BatchedExecutionStrategy.execute(BatchedExecutionStrategy.java:52)
	at graphql.execution.batched.BatchedExecutionStrategy.execute(BatchedExecutionStrategy.java:38)
	at graphql.execution.Execution.executeOperation(Execution.java:60)
	at graphql.execution.Execution.execute(Execution.java:33)
	at graphql.GraphQL.execute(GraphQL.java:78)
	at graphql.GraphQL.execute(GraphQL.java:55)
	at graphql.GraphQL.execute(GraphQL.java:47)
	at graphql.GraphQL.execute(GraphQL.java:43)
	at graphql.execution.batched.GraphqlExecutionSpec.runTestBatchingUnbatched(GraphqlExecutionSpec.groovy:40)
	at graphql.execution.batched.GraphqlExecutionSpec.runTest(GraphqlExecutionSpec.groovy:35)
	at graphql.execution.batched.GraphqlExecutionSpec.Legal null value for enum(GraphqlExecutionSpec.groovy:250)
```

Fix coming up.

PR #115, opened 2016-03-25
Fix for #114 - NPE when coercing null enum value in BatchedExecutionStrategy

tinnou:
Fix for https://github.com/andimarek/graphql-java/issues/114 NPE when coercing null enum value in BatchedExecutionStrategy

### DIFF
--- a/src/main/java/graphql/execution/batched/BatchedExecutionStrategy.java
+++ b/src/main/java/graphql/execution/batched/BatchedExecutionStrategy.java
@@ -209,7 +209,7 @@ public class BatchedExecutionStrategy extends ExecutionStrategy {
     private void handlePrimitives(List<GraphQLExecutionNodeValue> values, String fieldName,
                                   GraphQLType type) {
         for (GraphQLExecutionNodeValue value : values) {
-            Object coercedValue = coerce(type, value.getValue());
+            Object coercedValue = value.getValue() == null ? null : coerce(type, value.getValue());
             value.getResultContainer().putResult(fieldName, coercedValue);
         }
     }
--- a/src/main/java/graphql/execution/batched/BatchedExecutionStrategy.java
+++ b/src/main/java/graphql/execution/batched/BatchedExecutionStrategy.java
@@ -90,10 +90,7 @@ public class BatchedExecutionStrategy extends ExecutionStrategy {
         if (isPrimitive(fieldType)) {
             handlePrimitives(values, fieldName, fieldType);
             return Collections.emptyList();
-        } else if (isEnum(fieldType)) {
-            handleEnums(values, fieldName, fieldType);
-            return Collections.emptyList();
-        } if (isObject(fieldType)) {
+        } else if (isObject(fieldType)) {
             return handleObject(executionContext, values, fieldName, fields, fieldType);
         } else if (isList(fieldType)) {
             return handleList(executionContext, values, fieldName, fields, parentType, (GraphQLList) fieldType);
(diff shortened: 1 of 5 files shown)


andimarek:
I don't think null is a valid value in GraphQL (see http://facebook.github.io/graphql/#sec-Types.Non-Null).

What do you think?

Thanks

tinnou:
Yes, `null` is not a valid input value but it is a valid **result** value for enum. 

The test I added tried to follow the `FunWithStringsSchemaFactory` test strategy but in this case it might be a little confusing. Let me try to explain.

The`Legal null value for enum` test checks that the execution strategies coerce a `null` enum result value.
The `stringObjectValueFetcher` takes a string as source and if the source value is `"null"` it transforms it to `null`. All the other cases it just returns the same string. 

Does that make sense?

danielkwinsor:
It does make some amount of sense but is still confusing.  Can you rewrite your test to make a clearer separation between input coercion and result coercion?

I do think that `"null"` is a valid thing to try to coerce to an enum result, maybe `null` too (though not in the base GraphQLEnumValueDefinition).

However, your change to handlePrimitives has the side effect of now all primitives return null without coercion.  The spec says in 3.1.8 that 

> In all of the above result coercion, null was considered a valid value.

So really we should be allowing the result coercion to attempt to take place if null, and fix the NPE someplace else.  You either need a null check in getNameByValue or someplace else.
This should be good:
```
        if (value == null) {
            for (GraphQLEnumValueDefinition valueDefinition : valueDefinitionMap.values()) {
                if (valueDefinition.getValue() == null) return valueDefinition.getName();
            }
        }
        for (GraphQLEnumValueDefinition valueDefinition : valueDefinitionMap.values()) {
            if (value.equals(valueDefinition.getValue())) return valueDefinition.getName();
        }
```

danielkwinsor:
And for easy reference here is [Enum Values Spec](http://facebook.github.io/graphql/#sec-Enum-Value)

tinnou:
And now..I'm confused. What does the following mean? From the spec: `An enum value cannot be “null” in order to avoid confusion. GraphQL does not supply a value literal to represent the concept null.` 
Does it mean an enum result value is never `null` and therefore has to be one and only one of the declared values?

If yes, `ExecutionStrategy` is too lenient and we probably need to change the `ExecutionStrategy.completeValue` to disallow `null` by moving `completeValueForEnum` above the [null check](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/execution/ExecutionStrategy.java#L58)

If not we are back to our original issue and `BatchedExecutionStrategy` shouldn't throw a NPE.
@danielkwinsor: Good catch, indeed the check shouldn't be in `BatchedExecutionStrategy.handlePrimitives` for the reason you stated. Now if we want to be consistent with the other `handleXxx` methods we could probably do the null check before calling `BatchedExecutionStrategy.coerce`. Like so https://github.com/graphql-java/graphql-java/pull/115/commits/cd6848f0648c88903688de32149e3af2ae797ea2.

I also simplified the test method to only showcase the enum `null` result value coercion. We can clearly see the regular `ExecutionStrategy` let `null` return `null` but not the `BatchedExecutionStrategy`.

Either way, we need to make the strategies consistent.

danielkwinsor:
At this time, I see the issue is only when trying to give back a result that is null.  Null is a valid thing to return from a DataFetcher, even a DF for an enum.  The NPE happens when attempting to coerce this null value into some some name as in `getNameByValue`.

So `getNameByValue` needs a null check, it should go here and not in BES.  The only question to me is whether we detect null and return null (to avoid NPE on `value.equals()`) or really attempt to match up `null` with a GraphQLEnumValueDefinition with value Object == `null`.

Can you please limit your changes to a null check in `getNameByValue` and verify whether there are any other breaking places?

tinnou:
@danielkwinsor Made the changes per your comment. I added additional tests in `GraphQLEnumTypeTest` to cover the new code.

PR #121, opened 2016-04-16
Update ExecutionStrategy.java

dminkovsky:
I think this will never happen. `getFieldDef()` throws if `fieldDef` is null.


andimarek:
Thanks.

Issue #118, opened 2016-03-31
Recursive GraphQLTypeReference resolution


yrashk:
I am wondering if type reference resolution should go deeper beyond one level.

Right now, according to my best understanding of https://github.com/andimarek/graphql-java/blob/master/src/main/java/graphql/schema/SchemaUtil.java#L149-L155, for every object's field, it will do a type resolution. However, if that resolved type itself contains further type references, they won't be processed. Is this intentional or should it be improved? Happy to work on a PR if you think this should be improved.

danielkwinsor:
Just hit this too, here's a test case that reproduces the issue.

It affects queries not just on the object, but on the __schema too.


```
    @Test
    public void testRecursionByHand() throws Exception {
        logger.debug("testRecursionByHand");
        GraphQLOutputType recursiveClass = GraphQLObjectType.newObject()
                .name("RecursiveClass")
                .field(newFieldDefinition()
                        .type(Scalars.GraphQLString)
                        .name("recursionLevel")
                        .build())
                .field(newFieldDefinition()
                        .type(new GraphQLTypeReference("RecursiveClass"))
                        .name("recursive")
                        .build())
                .build();
        GraphQLObjectType queryType = newObject()
                .name("testQuery")
                .field(newFieldDefinition()
                        .type((GraphQLOutputType) recursiveClass)
                        .name("testObj")
                        .staticValue(new RecursiveClass(2))
                        .build())
                .build();
        GraphQLSchema recursiveTestSchema = GraphQLSchema.newSchema()
                .query(queryType)
                .build();
        
        String query = ""
        + "query testSchema {"
        + "  __schema {"
        + "    types {"
        + "      name"
        + "      fields {"
        + "        name"
        + "        type {"
        + "          name"
        + "        }"
        + "      }"
        + "    }"
        + "  }"
        + "}";

        String queryString = 
        "{"
        + "  testObj {"
        + "    recursionLevel"
        + "    recursive {"
        + "      recursionLevel"
        + "    recursive {"
        + "      recursionLevel"
        + "    recursive {"
        + "      recursionLevel"
        + "    recursive {"
        + "      recursionLevel"
        + "    }"
        + "    }"
        + "    }"
        + "    }"
        + "  }"
        + "}";
        
        try {
            new GraphQL(recursiveTestSchema).execute(query);
        }
        catch (Exception e) {
            logger.error("Caught exception", e);
            throw e;
        }
[...]

Issue #85, opened 2016-02-11
Reference Equality Check in VariableTypesMatchRule breaks if GraphQLInputObjectType is created multiple times.


davidscottcohen:
Normally in schema creation, it is OK to create multiple instances of the same type object.  However, `VariableTypesMatchRule` uses reference equality, so if I create my schema with multiple instances of an equivalent `GraphQLObjectInputType` then I will get validation failures.  This was tricky to track down because I create input types in methods (e.g. `private GraphQLInputObjectType  createPropertyMatchType()`) and then call those methods when I need it them.

I think the ideal solution is to use deep equality instead of reference equality in `VariableTypeMatchRule`.

Otherwise some process in schema creation should canonicalize the types within the schema.

yrashk:
I think I ran into a similar case with fragments (`PossibleFragmentSpreads`), particularly this https://github.com/andimarek/graphql-java/blob/master/src/main/java/graphql/validation/rules/PossibleFragmentSpreads.java#L76 makes use of reference equality.

yrashk:
@andimarek @alexkrebiehl @danielkwinsor

I want to discuss this issue. I believe it is important to make GraphQL type information equality testing deep as I can't see where reference equality testing is more useful (prove me wrong?).

Would it be sufficient to limit deep equality to names (as the names can't clash) or should we do full deep equality? For a quick fix, name equality works. What do you think?

andimarek:
The graphql schema was designed to be based on the principle that for each type there is exactly one Java Instance representing this type. That's the reason the ckecks are using reference equality. This is just simpler.

Of course we could change that, but I am not sure about the overall consequences.

yrashk:
Andreas, totally agree that this is simpler. However, I have a case where
it is actually the opposite. In graphql-java-annotations I don't know all
of the upcoming schema ahead of the time (it's being processed piece by
piece, particularly important for dynamic schema composition I do with
OSGi) and it's becoming a non-trivial problem, so right now my workaround
is to subclass graphql classes to implement deeper type checks to avoid
checking by reference. I am not sure what would we lose if we were to ditch
reference equality in favour of deeper equality, any thoughts?


Also, if you were to do the deeper equality check, how would you answer the
question in my previous comment?

On Fri, May 6, 2016 at 12:47 AM, Andreas Marek <notifications@github.com>
wrote:

> The graphql schema was designed to be based on the principle that for each
> type there is exactly one Java Instance representing this type. That's the
[...]

PR #123, opened 2016-04-24
Recursive types not being resolved

alexkrebiehl:
Addresses #118 

If we already have a type resolved on the typemap, the resolved type should point to the same instance.

### DIFF
--- a/src/main/java/graphql/schema/SchemaUtil.java
+++ b/src/main/java/graphql/schema/SchemaUtil.java
@@ -169,13 +169,4 @@ public class SchemaUtil {
         }
         return resolvedTypes;
     }
-
-    void linkTypes(GraphQLSchema schema) {
-        Map<String, GraphQLType> typeMap = allTypes(schema, schema.getDictionary());
-        for (GraphQLType type : typeMap.values()) {
-            if (type instanceof GraphQLFieldsContainer) {
-                resolveTypeReferencesForFieldsContainer((GraphQLFieldsContainer) type, typeMap);
-            }
-        }
-    }
 }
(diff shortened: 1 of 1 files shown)


yrashk:
1. What would be the best and shortest way to model a test case here? 
1. If we address #85 (switch from reference to deep equality), would that help? According to the last comment by @alexkrebiehl on #118, there's an issue with multiple instances of what's effectively one type.

alexkrebiehl:
I'll look at making a test case for this.

As far as reference vs deep equality, I don't think it will help here (at least in my use case).  I'm using graphql-java-annotations and building the schema at runtime, so nodes being created don't know about other nodes that were already created.  If I create two object that reference a type, I'll end up with an instance created for each reference.

Since we actually alter the object when resolving type references, we either need to make sure everything is pointing to the same instance of that object, or we have to resolve type references for all of the duplicates.

I don't think this will help #85 very much, since they are generating types on the fly.  However, this fix did solve my problems with fragment spreads.  Did you still have the fragment issue after the fix in this PR?

yrashk:
I would like to bump this issue up. I think the concern behind it is quite important.

yrashk:
Guys, any objections to merging this PR? It looks like the problem addressed here keeps popping up in different projects and this PR fixes it.

danielkwinsor:
Pr looks ok, if we do it I'll check in my test case.

dminkovsky:
Looks good to me.

Daniel Windsor says he has a test case, if I understood
him correctly. Could add that too as part of another PR. Maybe open an
issue for that if you merge this PR, so we don't forget.

пятница, 26 августа 2016 г. пользователь Yurii Rashkovskii написал:

markphilpot:
I agree with @yrashk -- let's merge this and open an Issue to track adding a test case for it.

PR #127, opened 2016-04-28
Update ExecutionContextBuilder.java

dminkovsky:
From [2.2](http://facebook.github.io/graphql/#sec-Language.Query-Document)

> if a GraphQL query document contains multiple operations, each operation must be named. When submitting a query document with multiple operations to a GraphQL service, the name of the desired operation to be executed must also be provided.

### DIFF
--- a/src/main/java/graphql/execution/ExecutionContextBuilder.java
+++ b/src/main/java/graphql/execution/ExecutionContextBuilder.java
@@ -32,7 +32,7 @@ public class ExecutionContextBuilder {
                 fragmentsByName.put(fragmentDefinition.getName(), fragmentDefinition);
             }
         }
-        if (operationName != null && operationsByName.size() > 1) {
+        if (operationName == null && operationsByName.size() > 1) {
             throw new GraphQLException("missing operation name");
         }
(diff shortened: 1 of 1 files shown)


dminkovsky:
I mean I think this is a bug. I convinced myself that this wasn't a bug sometime ago, but now looking at it again, I don't see how.

dminkovsky:
I will add a test

yrashk:
From the looks of it, it seems that this patch is indeed addressing an implementation bug. Just wondering what was the reasoning that you used to convince ourself this wasn't a bug before, if you can remember?  Just want to make sure we're not missing something else here.

danielkwinsor:
Yes I can tell that this fixes a bug, although in a different way than I have a test for.  I'd like to see a validator check that adds this to results.errors, rather than throw an exception here.  We're now escaping outside of the ExecutionContext engine upon receiving bad input.

By allowing this, we'd introduce a (second) source of thrown, uncaught exceptions but fixing thrown uncaught exceptions upon valid input like http://facebook.github.io/graphql/#sec-Operation-Name-Uniqueness

Issue #128, opened 2016-04-30
Union types may not have Scalars, Interfaces, or Unions as member types


danielkwinsor:
http://facebook.github.io/graphql/#sec-Union-type-validation

> Union type validation
> 
> Union types have the potential to be invalid if incorrectly defined.
> 
> The member types of an Union type must all be Object base types; Scalar, Interface and Union types may not be member types of a Union. Similarly, wrapping types may not be member types of a Union.
> A Union type must define one or more member types.

```
        public Builder possibleType(GraphQLType type) {
            assertNotNull(type, "possible type can't be null");
            types.add(type);
            return this;
        }
```

In the builder and the internal list of types, these objects should be GraphQLObjectTypes rather than GraphQLTypes

PR #129, opened 2016-04-30
Union types must have at least 1 member, and members can be only of O…

danielkwinsor:
…bject types, not scalars etc.

References #128  Union types may not have Scalars, Interfaces, or Unions as member types.  From spec.

This PR changes member types to GraphQLObjectType, creates a test case and some validation.


yrashk:
Thanks! This clearly makes graphql-java conform to that piece of the specification, so it should be in. Merged.

PR #140, opened 2016-05-11
Adding scalars for BigInteger, BigDecimal, Byte, Short, Char, and acc…

danielkwinsor:
…ompanying tests.

### DIFF
--- a/src/main/java/graphql/Scalars.java
+++ b/src/main/java/graphql/Scalars.java
@@ -8,15 +8,22 @@ import graphql.language.StringValue;
 import graphql.schema.Coercing;
 import graphql.schema.GraphQLScalarType;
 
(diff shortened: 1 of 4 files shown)


src/main/java/graphql/Scalars.java
```diff
@@ -178,4 +187,156 @@ public Object parseLiteral(Object input) {
         }
     });
 
+    public static GraphQLScalarType GraphQLBigInteger = new GraphQLScalarType("BigInteger", "Built-in java.math.BigInteger", new Coercing() {
```
okorz001: Minor detail, but I think there's value here in clarifying that these are non-standard types, assuming it was not a happy accident that GraphQLLong was not described as "Built-in Long".

PR #146, opened 2016-05-14
JDK6 compiler fixes

danielkwinsor:


danielkwinsor:
I'm going to merge this or else I'll run into merge problems later.  Compile passes but build fails at bintray upload.

Issue #106, opened 2016-03-15
Unspecified fields in Input object are set to null


lelong71:
If my input object has the following fields:

NameInput {
   firstName: String
   lastName: String
}
and if I call

updateNameMutation(input: {
  firstName: "John"
})

the input object I get from calling DataFetchingEnvironment.getAurgument("input") in my DataFetcher returns a map with lastName set to null. This breaks PATCH since we don't know if users explicitly set a field to null or it's filled by the code.
I expected that lastName should not be in the map since users didn't pass it. However if it has defaultValue defined, it should be in the map. This is how graphql-js currently implements

Related issue filed against graphql-js - https://github.com/graphql/graphql-js/issues/133

danielkwinsor:
We can't yet provide an `undefined` literal as javascript does, not sure what else they do in graphql-js.
DFE provides `public boolean containsArgument(String name)`  Possibly we can retrofit `containsArgument` to return false if the argument was never specified, while leaving `getArgument` the same Map style impl.

Let's watch [[RFC] Null value](https://github.com/facebook/graphql/pull/83).

trevor-morris:
I've seen a similar issue, though for me I only saw the problem when the input object was passed as a variable. When it was passed as a literal (as in the example above), only the fields actually present in the literal seemed to be set in the map. The relevant code seems to be in ValuesResolver where there are two methods, one for converting a literal to a map, the other for converting a variable to a map.

Here's the one for converting a literal:
````java
    private Object coerceValueAstForInputObject(GraphQLInputObjectType type, ObjectValue inputValue, Map<String, Object> variables) {
        Map<String, Object> result = new LinkedHashMap<String, Object>();

        for (ObjectField objectField : inputValue.getObjectFields()) {
            GraphQLInputObjectField inputObjectField = type.getField(objectField.getName());
            // illegal field ... no corresponding key in the schema
            if (inputObjectField == null) continue;
            Object fieldValue = coerceValueAst(inputObjectField.getType(), objectField.getValue(), variables);
            if (fieldValue == null) {
                fieldValue = inputObjectField.getDefaultValue();
            }
            result.put(objectField.getName(), fieldValue);
        }
        return result;
    }
```
Note how only the fields present in the inputValue (representing the literal) get transferred to the map. However this is the method used for converting a variable to a map
````java
    private Object coerceValueForInputObjectField(GraphQLInputObjectType inputObjectType, Map<String, Object> input) {
[...]

PR #155, opened 2016-06-04
Fix problem where unspecified fields in an object value for a variable are set to null

trevor-morris:
Addresses #106

When copying a variable value that corresponds to an input object type omit
any fields that are not present in the value, unless those fields have a
default value (in which case use the default value) or non null (which should
force an error)

This fixes a problem where such omitted fields were being copied with a null
value, making it impossible to tell whether the original variable value
omitted the field or explicitly set it to null.


trevor-morris:
I'm closing this request in favor of #156, which fixed another problem as well

PR #158, opened 2016-06-07
Expand README section on use with Relay

### LINKED ISSUE #154: Support for query variables
It looks like express-graphql and graphql-ruby accept queries with query and variables in separate structures.  For example, here is a query generated by the express-graphql client library.

```json
{
   "query":"query AppHomeRoute($uuid_0:String!) {
         account(uuid:$uuid_0) { 
               ...
         }
     }",
   "variables":{"uuid_0":"88888888-4444-4444-4444-121212121212"}
}
```
Here are a couple articles discussing variables:
 * http://rmosolgo.github.io/blog/2016/03/03/using-graphql-without-relay/
 * https://medium.com/the-graphqlhub/graphql-tour-variables-58c6abd10f56#.3lumjqvbm 

It doesn't look like the graphql-java server accepts query variables like this currently.  It seems like this would be a nice feature to have, so you can use graphql-java servers with the relay-starter-kit.

iancw:
Includes an example from the todomvc-relay-java codebase with explanation

resolves #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,28 @@ The `grapqhl-java` root Logger name is `graphql`.
 There is a very basic Relay support included. Please look at https://github.com/andimarek/todomvc-relay-java for an example
 project how to use it.
 
+Relay sends queries to the GraphQL server as JSON containing a `query` field and a `variables` field. The `query` field is a JSON string,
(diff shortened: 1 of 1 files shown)


andimarek:
Thanks!

PR #159, opened 2016-06-07
Update builders to add overrides that accept incomplete builders and jdk8-esque supplier interface through lambdas

IamCornholio:
Lambda functions are supported for data fetchers and type resolves but not graphql type builders. This adds an interface similar to that of the datafetcher interface to give graphql types a more natural lambda based build syntax.

`GraphqlObjectType obj = GraphqlObjectType.newObject();`
`obj.field(field -> {`
`                 return field.name("fieldName")`
`                         .argument(arg -> arg.name("argumentName"));`
`})`

Originally proposed by @Underzenith85.


aschrijver:
Would be nice. Why did you only add this for the 3 builders and not all? 
BTW Without the accolades and return statement the example looks a bit cleaner. You would use it like this:
```java
        GraphQLObjectType.newObject()
                .field(field -> field
                        .name("name")
                        .description("description")
                        .argument(argument -> argument
                                .name("arg")
                                .defaultValue("val"))
                        .deprecate("deprecated"))
                .field(field -> field
                        .name("name2"));
```

IamCornholio:
@aschrijver I don't know. object have fields. fields have arguments. What other builders would benefit from a supplier?
Perhaps @Underzenith85 can explain better

aschrijver:
I was thinking like for example `GraphQLInterfaceType` with `fieldDefinitions`, `GraphQLUnionType` with `types` and anywhere you can construct a list of nested graphql schema objects that have their own builder.

IamCornholio:
Ah yes. I think this works well with GraphQLInterfaceType.Builder#field, GraphQLDirective.Builder#argument, GraphQLEnumType.Builder#value (bit much perhaps?)

But GraphQLUnionType.Builder#possibleType? Or even the non list schema objects like GraphQLFieldDefinition.Builder#type, GraphQLInputObjectField.Builder#type?  I'm not sure about that. I don't know if I'd want to define an object type inline. I would find that confusing because defining that object type there in an anonymous class doesn't add that object type to the schema typeMap. Also I'd rather define it outside so I can keep a reference to it and use it for other union types. You could argue that a reference variable could be maintained through the anonymous classes but I don't know if such a structure would be helpful or better to read.

aschrijver:
Yes, this is probably enough. :+1:

dminkovsky:
This is cool. Improving the builders would be nice. Some thoughts:

- This project is implemented in Java 6. I don't think lambdas can be used unless that requirement is dropped.

- Other improvements to the builder system can be to add overloads that allow builders to be passed in where completed objects are required. Meaning we don't need to explicitly `.build()`

```
GraphqlObjectType obj = newObject()
  .field(newFieldDefinition.name("field"))
  .build();
```

aschrijver:
It doesn't matter if project is still Java 6. As long as it uses interfaces with a single method its okay. They  are identical to `@FuntionalInterface` but without the annotation (its just a marker) and can be used I  lambda's.

On your second point: Could you clarify? The `build()` is needed to start constructing after fluently setting 'constructor parameters' for the instance. Is the `newFieldDefinition` populated with defaults?

dminkovsky:
@aschrijver I updated my example. I didn't mean to exclude required fields in the example.

So, I am saying that it might be worth exploring adding the overload `GraphQLObjectType.Builder#field(GraphQLFieldDefinition.Builder field)`, and so on, everywhere an object that has a builder is expected. Such overloads call `.build()` on the builder, so you can just pass in the builders without calling `.build()` yourself.

I don't know about it also filling in default values. I'd rather such an overload just call `.build()` on the field and otherwise behave completely the same. So you don't have a tail of .`build()).build()).build()).build())...` at the end of your schema.

dminkovsky:
@aschrijver oh, and regarding Java 6 vs Java 8—yes, I figured that might be the case. But then the code should be written in Java 6 and a person using the Java 8 SDK will be able to use lambdas, right? The code needs to compile with the javac set to 1.6, right?

aschrijver:
Yes, that is correct.

On the Java 6 there is an old issue discussing it, and some people wanted 6 because that's what they are stuck with at their company (probably a bank or something).

Personally I am strongly in favour of Java 8. Its not 2007 anymore, that's stone age stuff in IT terms. 6 is no longer supported, 7 as well, and graphql-java misses out on opportunities to make a nice reactive API, use functional programming concepts, etc. Risks making itself irrelevant in the long run.

I'd say to any Java 6 guy 'You have a nice round 2.0 release now, and there will be a 2.1 as a bonus, but then you'll have to upgrade your mainframe to get a 3.x :wink:'

IamCornholio:
@dminkovsky I may have misunderstood what you're trying to say or don't know enough. But the way I have been using this is that the graphql-java jar is being used by code written using JDK8 syntax. And so I am currently able to use lambdas (as shown in my comments) in my code. I thought we could keep the project on java 6 and still be able to use the builder interface for a nicer jdk8 lamda syntax. Am I making any sense at all?

dminkovsky:
@IamCornholio I think we are all on the same page. Your proposed `BuilderFunction` will compile on 1.6 and then enable people on 1.8 to use lambdas. 

What do you think about adding overrides that let you pass unbuilt builders, too? (the overrides just call `.build()` on the builders, and nothing else). That would prevent a trail of `.build()).build()).build())...` at the end of your schema/type definitions.

aschrijver:
@IamCornholio when talking on Java 8 goodness, I do not mean the odd functional interface, but using `Function`, `Consumer`, `Supplier`, `Future` and the like. But also ability to use with other Java 8 library, like RxJava, reactive streams, etc.
A nice example of a Java 8 library that I recently found is JavaSlang. See http://www.javaslang.io/javaslang-docs/

aschrijver:
On Java 8 I would like to add that Java has lost many, many followers to other (arguably more sexy) language implementations already, because of its slow development.
I looked at the upcoming features in Java 9 and was greatly underwhelmed (once again).
I think using a modern Java version is _essential_ to stay relevant in the long run.

IamCornholio:
@aschrijver Ah yes. I agree to the extent of missing out on the goodness of jdk8 and libraries. I will probably remain a skeptic until I see the goodness actually being used though :)

@dminkovsky it took me a while to understand why one would ever see .build()) repeating. I finally got it when I saw the file Harness.java. Yes it is very awkward that we have to call the build on every thing. As I understand it. You are proposing this
```
public Builder argument(GraphQLArgument.Builder builder) {
  this.arguments.add(builder.build());
  return this;
}
```

in addition to this
```
/**
* Take a argument builder in a function definition and apply. Can be used in a jdk8 lambda
* e.g.:
* <pre>
*     {@code
*      argument(a -> a.name("argumentName"))
*     }
* </pre>
*/
public Builder argument(BuilderFunction<GraphQLArgument.Builder> builderFunction) {
  GraphQLArgument.Builder builder = GraphQLArgument.newArgument();
  builder = builderFunction.apply(builder);
  this.arguments.add(builder.build());
  return this;
}
```
I agree that should be done (or perhaps not, see following blurb).  Are you proposing this be done in the same pull request though?

I have this other notion that fields and arguments are dependent on the existence of an object and as such shouldn't be allowed to build() without the context of the object type. That the build method should not be called in the first place and the schema should really build when the GraphQLSchema.build(dictionary) is called.

IamCornholio:
Sorry for the spam. I've updated this PR and it's scope.

dminkovsky:
Appreciate it! Thank you! Will review this afternoon/evening.

пятница, 19 августа 2016 г. пользователь Nikhil Agarwal написал:

dminkovsky:
This looks good to me, thank you.

I couldn't find any other places where the changes in this PR might also be applied. Also, I don't see unit tests for these schema classes, so there's no obvious existing place we could add tests for this, right?

Only thing I'd say is, if possible, could the commit message for 90b6b2f be split over several lines  (git rebase -i, if you're familiar with that?), like

> Summarize changes in around 50 characters or less

> More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical (unless
you omit the body entirely); various tools like `log`, `shortlog`
and `rebase` can get confused if you run the two together.

IamCornholio:
@dminkovsky Yea I couldn't find an obvious place for the tests either.

I'm familiar with editing commit messages but do you want me to change it for https://github.com/graphql-java/graphql-java/commit/90b6b2f51bb68df410c2a03306544c81abcd360d? That's not a commit I wrote. I'll check the format for the one's I wrote.

IamCornholio:
I didn't know that this 50/72 was a standard. Learn something new everyday I guess.
Rebase done now. Messages shortened. Added a simple field/argument test also.

dminkovsky:
Oh my, for some reason I thought `90b6b2f` was from your pull request. Sorry for the confusion. I don't know the width thing to be a standard so much as a widely adopted convention I've seen. 

Yeah if you feel like squashing these down you could do that. I wasn't going to suggest it but please do if you see some commits you think can be easily combined for clarity/atomicity. 

Thank you!

IamCornholio:
@dminkovsky I think it's ok as a single commit. I've updated the javadocs a bit too.

aschrijver:
Just curious, but why did @IamCornholio need to merge into single commit? For atomicity?
If you want that as a convention, it should be documented, because the rebasing/merging is tedious work for any contributor (unless they are real Git fanatics).

IamCornholio:
@aschrijver There were 5 commits. I didn't need to squash them. But I do prefer a clean history and so I asked @dminkovsky and he said sure. He clearly says he wasn't going to suggest it himself.

I say atomicity is a good reason for squashing commits but then I am not a maintainer (yet 😉 )
I can't share an unbiased opinion about my fanaticism either 😄

aschrijver:
Ha ha ha, I understand. I am also a bit of a fanatic sometimes, I must admit :+1:

dminkovsky:
I love `git rebase -i` or even just resetting a branch and redoing the
commits. But it can get way out of hand sometimes so I never wish that upon
others.

суббота, 20 августа 2016 г. пользователь Arnold Schrijver написал:

src/main/java/graphql/schema/GraphQLDirective.java
```diff
@@ -118,6 +118,37 @@ public Builder argument(GraphQLArgument fieldArgument) {
...
+         *
+         * @param builderFunction a supplier for the builder impl
+         * @return this
+         */
+        public Builder argument(BuilderFunction<GraphQLArgument.Builder> builderFunction) {
+            GraphQLArgument.Builder builder = GraphQLArgument.newArgument();
+            builder = builderFunction.apply(builder);
+            this.arguments.add(builder.build());
+            return this;
+        }
```
dminkovsky: Could we call and return the `argument(GraphQLArgument.Builder builder)` overload that's defined below?

src/main/java/graphql/schema/GraphQLFieldDefinition.java
```diff
@@ -131,6 +131,37 @@ public Builder argument(GraphQLArgument argument) {
...
+         * @param builderFunction a supplier for the builder impl
+         * @return this
+         */
+        public Builder argument(BuilderFunction<GraphQLArgument.Builder> builderFunction) {
+            GraphQLArgument.Builder builder = GraphQLArgument.newArgument();
+            builder = builderFunction.apply(builder);
+            this.arguments.add(builder.build());
+            return this;
+        }
+
```
dminkovsky: Same as above?

src/main/java/graphql/schema/GraphQLInputObjectType.java
```diff
@@ -75,6 +75,38 @@ public Builder field(GraphQLInputObjectField field) {
...
+         * @return this
+         */
+        public Builder field(BuilderFunction<GraphQLInputObjectField.Builder> builderFunction) {
+            assertNotNull(builderFunction, "builderFunction should not be null");
+            GraphQLInputObjectField.Builder builder = GraphQLInputObjectField.newInputObjectField();
+            builder = builderFunction.apply(builder);
+            fields.add(builder.build());
+            return this;
+        }
+
```
dminkovsky: Same as above?

src/main/java/graphql/schema/GraphQLInterfaceType.java
```diff
@@ -93,6 +93,38 @@ public Builder field(GraphQLFieldDefinition fieldDefinition) {
...
+         * @return this
+         */
+        public Builder field(BuilderFunction<GraphQLFieldDefinition.Builder> builderFunction) {
+            assertNotNull(builderFunction, "builderFunction can't be null");
+            GraphQLFieldDefinition.Builder builder = GraphQLFieldDefinition.newFieldDefinition();
+            builder = builderFunction.apply(builder);
+            this.fields.add(builder.build());
+            return this;
+        }
+
```
dminkovsky: As above?

src/main/java/graphql/schema/GraphQLObjectType.java
```diff
@@ -97,6 +97,38 @@ public Builder field(GraphQLFieldDefinition fieldDefinition) {
...
+         * @return this
+         */
+        public Builder field(BuilderFunction<GraphQLFieldDefinition.Builder> builderFunction) {
+            assertNotNull(builderFunction, "builderFunction can't be null");
+            GraphQLFieldDefinition.Builder builder = GraphQLFieldDefinition.newFieldDefinition();
+            builder = builderFunction.apply(builder);
+            this.fieldDefinitions.add(builder.build());
+            return this;
+        }
+
```
dminkovsky: One more

src/test/groovy/graphql/GraphQLTest.groovy
```diff
@@ -39,6 +41,36 @@ class GraphQLTest extends Specification {
...
+                    .type(GraphQLString)
+                    .name("field")
+                    .argument(newArgument()
+                        .name("arg")
+                        .type(GraphQLString))
+                    .dataFetcher(dataFetcher))
+        .build()
+
+        GraphQLSchema schema = newSchema()
+                .query(objectType).build()
```
dminkovsky: Could we add an override to `query()` that accepts a `GraphQLObjectType.Builder`, so that we don't have to call `.build()` above? Same with `mutation()` then.

Can we do the same with `GraphQLFieldDefinition#type()`? For that it would be an overload that accepts a `GraphQLObjectType.Builder`. I think thats the only builder it should accept.

dminkovsky:
Hey. Back from vacation here. I added some line notes to your commit. But I think it would be easier if I just made the chances. I'll open a PR against your repo. Thanks.

dminkovsky:
@IamCornholio okay so I opened https://github.com/IamCornholio/graphql-java/pull/1 against your repo. Please review and let me know what you think over on that PR.

IamCornholio:
@dminkovsky Ah yes. Looks like I missed some obvious ones there. The build() calls are still there in the tests. I wonder if we should remove them as part of this pull request?

Also I'm assuming that you want me to cherry-pick those commits and add them to this pull request?

IamCornholio:
@dminkovsky let me know if you want me to remove that last commit and/or squash others

dminkovsky:
@IamCornholio Good morning! 

Yeah, I think removing those `.build()` calls is a great way to test that this works. Looks good to me. Thanks.

Please feel free to squash/edit history however you want. I had those commits set up for clarity within the context of this PR. Combining them all before merging makes sense to me. 

Thank you.

IamCornholio:
@dminkovsky A good morning to you too :smile: 

All done now

bpatters:
LGTM

PR #160, opened 2016-06-07
Add methods to field and argument builders that set the types using a builder method syntax

IamCornholio:
It'd be nice to set the type also using a builder syntax
Instead of doing this:
`field.type(new GraphQLNonNull(GraphQLString))`
do this:
`field.stringType().notNull()`

Originally proposed by @Underzenith85

### DIFF
--- a/src/main/java/graphql/schema/GraphQLArgument.java
+++ b/src/main/java/graphql/schema/GraphQLArgument.java
@@ -1,6 +1,7 @@
 package graphql.schema;
 
 
+import graphql.Scalars;
 import static graphql.Assert.assertNotNull;
 
 public class GraphQLArgument {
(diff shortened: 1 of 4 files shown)


dminkovsky:
I think this would expand the API too much with little or negative gains given the added complexities. Take the example in this PR: the changes cover only scalars that are built-ins. But scalars can be defined ad-hoc. So if you define your own scalar and this API exists, the scalar you defined is somewhat of a second-class citizen because it doesn't have this API.

This also steps deviates from the JS implementation, in the sense that the JS implementation does not have this syntax. I know there is no requirement to make this implementation look like the JS implementation, but I do think it is best follow the reference implementation's lead on look and feel to the extent that language differences allow. 

I want to [improve the builder interface](https://github.com/graphql-java/graphql-java/pull/159#issuecomment-240769189), but I don't think this is a step in the right direction.

IamCornholio:
I missed that it could have become inconsistent with custom defined scalars. I accept that this is not a step in the right direction.

dminkovsky:
Thank you for posting :D. Let's continue on #159. I want to get a next release going with some merges in the next two weeks.

PR #161, opened 2016-06-07
Remove generated antlr code. Add build task dependency

IamCornholio:
Generated antlr code will be added during library build/publish.

Originally proposed by @Underzenith85


IamCornholio:
Also fixed a typo

markphilpot:
Thanks! This makes sense to separate.

Issue #130, opened 2016-05-03
Boolean fields *must* be accessed using an isX method


sazzer:
If you have a DataFetcher that returns an object instance for the data, and the object uses "getX" fields for accessing booleans then this causes the values to not be discovered and so null is returned instead.

As a concrete example of this, the Kotlin Language generates getters for properties automatically, but always in the form "getX". This is allowed by the Java Beans specification (Section 8.3.2), but fails here. I can imagine it might be the case with other places that API model objects are generated instead of hand-written.

sazzer:
The problem is in PropertyDataFetcher.getPropertyViaGetter(). Updating it to support both isX and getX fields looks simple enough, so I might have a go if I get enough spare time.

danielkwinsor:
I like PropertyDataFetcher the way it is, but alternative behavior can be subclassed.  The reason I like it as is: consider isClass and getClass.  Or consider isAnything and getAnything where the return types are different, or what is not expected.  This could be worse than getting null values.

With that said, I've already subclassed these fetchers in my own code and suggest the same here.

sazzer:
If you have both an isXxx and a getXxx method for the same name but with
different return types, you're not a valid Java bean according to the spec.
It's acceptable to have both but they should reflect the same field and
therefore have the same value.

If the PropertyDataFetcher is to be used purely for beans this means this
is a non issue. If it's to be used for any arbitrary class then your right
it causes problems...

ayhanap:
I also had problem with Boolean property accessors generated with [Lombok](https://github.com/rzwitserloot/lombok) and IDEA. The convention suggests `is` prefix for primitive `boolean` getters, so I think supporting `get` prefix for `Boolean` getters should bring no harm.

Aggree @sazzer for classes having both `isXxx `and `getXxx`. Even that is the case supplied a pull request #164 prioritizing `is` over `get` prefix for Boolean properties.

markphilpot:
Closing this since we merged the fix [here](https://github.com/graphql-java/graphql-java/issues/164).

PR #164, opened 2016-06-15
Fix for #130 - Adds support for Boolean property accessors with 'get' prefix in PropertyDataFetcher

ayhanap:
As mentioned in #130 some languages and frameworks uses "get" prefix for `Boolean `types. This was also the case for me. I use Lombok for generating accessors which also uses this convention. Even IDE's generate getters for `Boolean` with "get"

The prefix 'is' generally used for primitive `boolean`, Lombok has a option to disable this but not for `Boolean`. So I think this functionality should be a part of `PropertyDataFetcher`. Otherwise it will be error prone or some unnecessary configuration/code for these fields.

### DIFF
--- a/src/main/java/graphql/schema/PropertyDataFetcher.java
+++ b/src/main/java/graphql/schema/PropertyDataFetcher.java
@@ -30,19 +30,19 @@ public class PropertyDataFetcher implements DataFetcher {
         try {
             if (isBooleanProperty(outputType)) {
                 try {
-                    return getPropertyViaGetterUsingPrefix(object, outputType, "is");
+                    return getPropertyViaGetterUsingPrefix(object, "is");
                 } catch (NoSuchMethodException e) {
-                    return getPropertyViaGetterUsingPrefix(object, outputType, "get");
+                    return getPropertyViaGetterUsingPrefix(object, "get");
                 }
             } else {
-                return getPropertyViaGetterUsingPrefix(object, outputType, "get");
+                return getPropertyViaGetterUsingPrefix(object, "get");
             }
         } catch (NoSuchMethodException e1) {
-            return getPropertyViaFieldAccess(object, outputType);
+            return getPropertyViaFieldAccess(object);
         }
     }
 
-    private Object getPropertyViaGetterUsingPrefix(Object object, GraphQLOutputType outputType, String prefix) throws NoSuchMethodException {
+    private Object getPropertyViaGetterUsingPrefix(Object object, String prefix) throws NoSuchMethodException {
         String getterName = prefix + propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1);
         try {
             Method method = object.getClass().getMethod(getterName);
@@ -63,7 +63,7 @@ public class PropertyDataFetcher implements DataFetcher {
         return false;
     }
 
-    private Object getPropertyViaFieldAccess(Object object, GraphQLOutputType outputType) {
+    private Object getPropertyViaFieldAccess(Object object) {
         try {
             Field field = object.getClass().getField(propertyName);
             return field.get(object);
--- a/src/main/java/graphql/schema/PropertyDataFetcher.java
+++ b/src/main/java/graphql/schema/PropertyDataFetcher.java
@@ -27,20 +27,18 @@ public class PropertyDataFetcher implements DataFetcher {
     }
 
     private Object getPropertyViaGetter(Object object, GraphQLOutputType outputType) {
-        boolean isBooleanProperty = isBooleanProperty(outputType);
-        String prefix = isBooleanProperty ? "is" : "get";
         try {
-            return getPropertyViaGetterUsingPrefix(object, outputType, prefix);
-        } catch (NoSuchMethodException e) {
-            if (isBooleanProperty){
+            if (isBooleanProperty(outputType)) {
                 try {
+                    return getPropertyViaGetterUsingPrefix(object, outputType, "is");
+                } catch (NoSuchMethodException e) {
                     return getPropertyViaGetterUsingPrefix(object, outputType, "get");
-                } catch (NoSuchMethodException e1) {
-                    return getPropertyViaFieldAccess(object, outputType);
                 }
-            }else{
-                return getPropertyViaFieldAccess(object, outputType);
+            } else {
+                return getPropertyViaGetterUsingPrefix(object, outputType, "get");
             }
+        } catch (NoSuchMethodException e1) {
+            return getPropertyViaFieldAccess(object, outputType);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/schema/PropertyDataFetcher.java
```diff
@@ -27,14 +27,31 @@ public Object get(DataFetchingEnvironment environment) {
...
+        }
+    }
+
+    private Object getPropertyViaGetterUsingPrefix(Object object, GraphQLOutputType outputType, String prefix) throws NoSuchMethodException {
         String getterName = prefix + propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1);
         try {
             Method method = object.getClass().getMethod(getterName);
             return method.invoke(object);
 
         } catch (NoSuchMethodException e) {
```
markphilpot: Can you just not have a catch block for `NoSuchMethodException`?

src/main/java/graphql/schema/PropertyDataFetcher.java
```diff
@@ -27,14 +27,31 @@ public Object get(DataFetchingEnvironment environment) {
     }
 
     private Object getPropertyViaGetter(Object object, GraphQLOutputType outputType) {
-        String prefix = isBooleanProperty(outputType) ? "is" : "get";
+        boolean isBooleanProperty = isBooleanProperty(outputType);
+        String prefix = isBooleanProperty ? "is" : "get";
+        try {
+            return getPropertyViaGetterUsingPrefix(object, outputType, prefix);
+        } catch (NoSuchMethodException e) {
+            if (isBooleanProperty){
```
markphilpot:
```java
if(isBooleanProperty){
   try {
      return getPropertyViaGetterUsingPrefix(object, outputType, "get");
   } catch (NoSuchMethodException e1) {
      // Empty
   }
}
return getPropertyViaFieldAccess(object, outputType);
```

markphilpot:
Thanks for the PR!  Let me know if the comments make sense.

aschrijver:
It could also have an additional clause with a search for any property that contains `propertyName`, and return the first match. This way you will catch any reasonable naming convention (only not the completely unrelated names).

dminkovsky:
Yes, thank you for this PR! Let's fix this, given the demand.

What do you think about this: https://github.com/dminkovsky/graphql-java/blob/8efb230b58224488c335cc8409c17a73f0657b51/src/main/java/graphql/schema/PropertyDataFetcher.java

I think it might be clearer?

ayhanap:
Thanks for review @markphilpot The current flow to find property data is as follows
1. Try accessor methods. If the property is boolean use `is` getter methods instead of `get`.
2. If no method exists then .. If the property is boolean, try `get` accessor method also.
3. If no method exists then .. Get property data with field access.
Before this PR the current flow was 1 then 3, without 2.
So the code in your inline comments does not fulfill the requirements. Therefore I suggest you to check working code and let me know if this makes sense.

ayhanap:
@dminkovsky Looks simpler and more explanatory indeed.

dminkovsky:
Whoops didn't mean to overload `getPropertyViaGetter` there. One of those should be renamed.

dminkovsky:
@ayhanap cool. If you like that more, feel free to update your PR with that variant and we can see what other people think. Will be glad to close this one.

ayhanap:
Updated the PR with @dminkovsky's cleaner suggestion.

dminkovsky:
@ayhanap cool, thanks. I think there's also @markphilpot's question regarding that caught and then immediately throw `NoSuchMethodException` in `getPropertyViaGetterUsingPrefix()`. It does seem that it's not necessary?

ayhanap:
Sorry missed that. Fixed that.

src/main/java/graphql/schema/PropertyDataFetcher.java
ayhanap: Right, just removed the unnecessary catch block for `NoSuchMethodException`

aschrijver:
How do you think about earlier suggestion to allow any kind of prefix or postfix https://github.com/graphql-java/graphql-java/pull/164#issuecomment-241156317 ?

aschrijver:
Yet another option, and easily added is a public overload that takes the prefix.

ayhanap:
I think searching for method names containing fielddefinition names could lead to some unexpected results.
All I can think is to search for a method name as exact field definition name. So that way you can define field definitions on your getter method name. And the example scenario I mentioned is as follows.
```
newFieldDefinition()
            .name("shouldUpdateAccount") // instead of using updateAccount
            .description("Should Update Account Details")
            .type(GraphQLString)
            .build()
```
And a POJO like this
```
public class Account{
    boolean updateAccount;
    public shouldUpdateAccount(){
        return this.uppdateAccount;
    }
}
```
What do you think @aschrijver.

aschrijver:
Yes, I agree, you are right with the search. It would be too much of a burden on the developer to get it right, and don't forget when adding a similarly-named member in future.

Passing a prefix parameter could be solution.

On a different note, related but not the same, I have a need to be able to define field aliases. I am discussing that on the metadata-related issues I created earlier.

dminkovsky:
I kind of like the idea of keeping the surface of this class small, like it currently is. Property fetchers, after all, are explicit public APIs. If people need more specific functionality, they can write their own property fetcher! With lambda notation this is basically trivial for a simple field like `shouldUpdateAccount` that just needs to be accessed. To me, this PR is about responding to a specific repeated request from users. 

So, let's revisit adding more functionality to this property fetcher when there is repeated, specific demand? I like how it looks right now as of this PR, especially with the removed unused fields.

aschrijver:
Yeah, that's no problem. Let's close this one. Given how busy I am it will be some time before I revisit this, maybe then I'll come up with a more elegant PR. We'll see..

ayhanap:
Agree, those other requirements need more discussion and evaluation. Keeping in mind that this class is responsible of most of the data fetching, it should be fast and simple.

dminkovsky:
Thanks Mark

суббота, 20 августа 2016 г. пользователь Mark Philpot написал:

aschrijver:
@dminkovsky Hailing from Russia or Ukrain? Then you are making even more late-nighters than me :wink:

markphilpot:
пожалуйста

dminkovsky:
Oh no thankfully on the east coast of the US. Have my phone in Russian
though, so it does this. I try to get to sleep at a reasonable hour :)

суббота, 20 августа 2016 г. пользователь Mark Philpot написал:

aschrijver:
Lucky you, ha ha. I should be horizontal now at 04:00PM on the clock, not in S shape over laptop :grin:

PR #166, opened 2016-06-16
Fix #165

### LINKED ISSUE #165: Relay: mutationWithClientMutationId
I just made my own implementation of `Relay#mutationWithClientMutationId()` before noticing that the method exists in this package. My implementation is identical with one exception. Shouldn't [line 190](https://github.com/graphql-java/graphql-java/blob/2ab1b6a3c28bca930715c245499dacaee44bf22a/src/main/java/graphql/relay/Relay.java#L190), which is currently 

```
.dataFetcher(dataFetcher)
```

instead wrap the user-provided `DataFetcher` and place the `clientMutationId` from input into the result? Like so:

```
.dataFetcher(env -> {
    Map<String, Object> data = (Map<String, Object>) dataFetcher.get(env);
    data.put("clientMutationId", ((Map<String, Object>) env.getArguments().get("input")).get("clientMutationId"));
    return data;
})
```

Without this, how does this field makes its way from input to output and get resolved downstream? Granted, I haven't tried the current implementation but I don't know how it would work...

dminkovsky:
In case my observation in #165 is correct, here is a PR. See also the [JavaScript implementation](https://github.com/graphql/graphql-relay-js/blob/f1dfe502cfeccf80deadce3ce2cfde618a8acb36/src/mutation/mutation.js#L90-L95).

### DIFF
--- a/src/main/java/graphql/relay/Relay.java
+++ b/src/main/java/graphql/relay/Relay.java
@@ -187,7 +187,11 @@ public class Relay {
                         .name("input")
                         .type(new GraphQLNonNull(inputObjectType))
                         .build())
-                .dataFetcher(dataFetcher)
+                .dataFetcher(env -> {
+                    Map<String, Object> data = (Map<String, Object>) dataFetcher.get(env);
+                    data.put("clientMutationId", ((Map<String, Object>) env.getArguments().get("input")).get("clientMutationId"));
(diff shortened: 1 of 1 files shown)


dminkovsky:
Oh I did not notice this failed. Will review.

PR #168, opened 2016-06-19
Parse escaped characters in strings

ntkoopman:
Escaped characters in string literals were not replaced with the correct character.

A ShouldNotHappenException is thrown on unknown escape codes because these should already be rejected during parsing.


markphilpot:
Spec Reference: http://facebook.github.io/graphql/#sec-String-Value

markphilpot:
Thanks!

PR #175, opened 2016-07-26
Fixes for location and directives

IamCornholio:
proposed by @trevor-morris
- Update introspection for directives. Return values for "locations" (the field used in the latest GraphQL spec) and also for "onOperation", "onFragment" and "onField" (the deprecated fields which our current GraphiQL expects in its schema query).
- Fix a problem with the KnownArgumentNames validation check. It was assuming all arguments belonged to a field, but they can also belong to directives. So changed the level of the check - we check fields and directives for unknown argument names, rather than looking at an argument and then assuming it must belong to a field.
- Fix a problem with the Antlr grammar for inline fragments. It had this:
  `inlineFragment : '...' 'on' typeCondition directives? selectionSet;`
  That is the "type condition" (on TypeName) was required. Changed it to be optional, which involved changing the grammar, the way we construct the parse tree and then a bunch of new null checks to deal with the case where the type condition is null.

### DIFF
--- a/src/main/antlr/Graphql.g4
+++ b/src/main/antlr/Graphql.g4
@@ -47,13 +47,13 @@ argument : NAME ':' valueWithVariable;
 
 fragmentSpread : '...' fragmentName directives?;
 
-inlineFragment : '...' 'on' typeCondition directives? selectionSet;
+inlineFragment : '...' typeCondition? directives? selectionSet;
 
-fragmentDefinition : 'fragment' fragmentName 'on' typeCondition directives? selectionSet;
+fragmentDefinition : 'fragment' fragmentName typeCondition directives? selectionSet;
 
 fragmentName :  NAME;
 
-typeCondition : typeName;
+typeCondition : 'on' typeName;
 
 // Value
 
--- a/src/main/java/graphql/execution/FieldCollector.java
+++ b/src/main/java/graphql/execution/FieldCollector.java
@@ -79,6 +79,9 @@ public class FieldCollector {
 
 
     private boolean doesFragmentConditionMatch(ExecutionContext executionContext, InlineFragment inlineFragment, GraphQLObjectType type) {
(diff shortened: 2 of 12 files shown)


IamCornholio:
This is failing because the grammar needs to be regenerated. It seems like @andimarek had some IDE settings that are different from mine and so my generated grammar diff is wildly different. In any case once https://github.com/graphql-java/graphql-java/pull/161 is merged then the generated grammar doesn't have to be checked in. It is unnecessary.

IamCornholio:
We have discovered that this has a bug where declaring a directive on an inline fragment doesn't work. Fix is in progress.

IamCornholio:
Fixed the inline fragment directive bug as well. Author (@trevor-morris) says this is a bug even on the graphql-js and graphiql.

IamCornholio:
Removed conflicts. Ready for review/merge.

crazytoucan:
I am quite excited for this PR, as this will get us closer to out-of-the-box correctness with latest versions of GraphiQL. Is there anything blocking this merge?

dminkovsky:
@crazytoucan thanks for pinging on this. Nothing in particular was holding up this PR. Just lack of attention, unfortunately. With your interest, though, I will mark this for 2.3.0.

An update though; re:

> * Fix a problem with the KnownArgumentNames validation check. It was assuming all arguments belonged to a field, but they can also belong to directives. So changed the level of the check - we check fields and directives for unknown argument names, rather than looking at an argument and then assuming it must belong to a field.

I recently merged a PR (#180) that fixed this rather embarrassing issue! So that is already on master.

I will look at the rest of this in the coming days.

dminkovsky:
Hey @IamCornholio, thank you for these fixes. The second bullet point was already addressed in a #190. I had to merge this PR as a result, so in the process I broke the remaining two bullet points into PRs #274 and #275. Closing this and merging those. Thank you!

PR #178, opened 2016-07-28
Add overload of GraphQLEnumType.Builder.value() with deprecationReason

### LINKED ISSUE #177: GraphQLEnumValueDefinition.getDeprecationReason is never initialized
The constructor below in `GraphQLEnumValueDefinition` is only called by its overload so `deprecationReason` is always `null`.
```
public GraphQLEnumValueDefinition(String name, String description, Object value, String deprecationReason)
```

aschrijver:
Fixes #177


markphilpot:
Thanks!

Issue #16, opened 2015-09-05
directive


mmjd:
I am trying to use directive in the following code example but if raises error.

```java
GraphQLObjectType d = newObject()
				.name("Data")
				.field(newFieldDefinition()
						.name("ID")
						.type(GraphQLFloat)
						.build())
						.field(newFieldDefinition()
								.name("Name")
								.type(GraphQLString)
								.build())
								.build();

		GraphQLObjectType pageObject = newObject()
				.name("Page")
				.field(newFieldDefinition()
						.name("Data")
						.type(d)
						.argument(
								GraphQLArgument.newArgument()
								.name("ID")
								.type(GraphQLFloat)
								.build()
								)
								.argument(
										GraphQLArgument.newArgument()
										.name("Name")
										.type(GraphQLString)
										.build()
										)
										.dataFetcher(new DataFetcher() {
											@Override
											public Object get(DataFetchingEnvironment environment) {

												Map<String, Object> arguments = environment.getArguments();
												Float id = (Float) arguments.get("ID");
												String name = (String) arguments.get("Name");

												Map<String, Object> result = new HashMap<String, Object>();
												result.put("ID", id);
												result.put("Name", name);
												return result;
											}
										})
										.build()
						)
						.build();


		GraphQLObjectType query = newObject()
				.name("sample")
				.field(newFieldDefinition()
						.name("Page")
						.type(pageObject)
						.staticValue(pageObject)
						.build())
						.build();

		GraphQLSchema schema = GraphQLSchema.newSchema()
				.query(query)
				.build();

		System.out.println("Schema: " + schema.toString());
		//@include(Name: \"my\")
		String input = "query sample ($id: Float = 1.0, $name: String, $condition: Boolean) "
				+ "{ Page { Data (ID: $id, Name:$name)  { ID , Name @skip(if: $condition) } } }";

		Map<String, Object> arguments = new HashMap<String, Object>();
		arguments.put("id", "2");
		arguments.put("name", "myName");
		arguments.put("condition", false);
		ExecutionResult result = new GraphQL(schema).execute(input, new Object(), arguments );
```

```
ValidationError{validationErrorType=MissingDirectiveArgument, sourceLocations=[SourceLocation{line=1, column=105}], description='Missing directive argument skip'}
ValidationError{validationErrorType=UnknownArgument, sourceLocations=[SourceLocation{line=1, column=111}], description='Unknown argument if'}
[...]

PR #190, opened 2016-08-22
Directive argument name check against directive instead of field

p-guldbaek-elsevier:
The current "KnownArgumentNames" implementation does not take directive arguments into account and verifies all arguments as if they are arguments for fields. It can be worked around (see #16 that suggests adding an "if" argument to fields), but the change suggested in this pull request avoids cluttering the schema.


arlampin:
Could this be merged? Currently graphql-java doesn't realistically support directives, as you would need to add all arguments from all directives to all fields where you would like to use those directives.

dminkovsky:
Yes, will be fixing this now

dminkovsky:
Patch and tests look good. Thank you.

Sorry about the horrible delay here! I had this patched locally and somehow didn't realize this hadn't been merged. 

>  Currently graphql-java doesn't realistically support directives

Yes, yikes. I am going to see about getting this released to Maven soon. Please ping more if other issues are like are not addressed in the future.

PR #191, opened 2016-08-23
some code cleanup

Jimexist:
- always use UTF-8 for decoding and encoding global id
- use getter instead of field in global id


markphilpot:
Makes sense.  Thanks for the PR!

PR #201, opened 2016-09-06
illegal global id should yield illegal argument exception

Jimexist:
instead of array index out of bound exception


dminkovsky:
Thanks @Jimexist

Issue #37, opened 2015-10-06
Asynchronous DataFetcher?


ybayk:
~~Please consider adding support of Future return type for DataFetcher API.~~
Actually scratched that. We just noticed it is possible to provide a custom executor service per fetcher. But it will still block for each fetcher.
Is it actually possible to go async all the way?  
1) Across all fetchers
2) Make top level graphql execute() async as well.

tberman:
I have an implementation of an ExecutionStrategy that uses rxjava that I will be publishing as open source soon.

yurgis2:
That would be exactly what we are looking for. Any hint on how soon? :)

tberman:
https://github.com/nfl/graphql-rxjava is available now, readme coming soon, and trying to get it put into central ASAP

akhahaha:
I would also like to second @ybayk's request to go async all the way (via fetchers and execute()). @tberman's work-around still leaves something to be desired.

I'm new to GraphQL and have only looked at this Java implementation thus far, but it seems to me that this would more closely follow the Javascript specification, which states that [ExecutionResults should be returned via a promise](http://graphql.org/docs/api-reference-execution/).

If others agree with this sentiment, I'd be happy to start working or collaborate on a pull request to make this happen.

akhahaha:
@dminkovsky I'd integrate some sort of callback interface as the primary method of returning data from execute() and fetchers.

Query executions would probably end up looking something like this:
```
new GraphQL(schema).execute("{hello}", new GraphQLCallback() {
            @Override
            public void onComplete(ExecutionResult result) {
                // Handle result
            }
        });
```

and the DataFetcher interface would need to change to:
```
public interface DataFetcher {
    Object get(DataFetchingEnvironment var1, DataFetchCallback callback);
}

// Sample async DataFetcher
new DataFetcher() {
    @Override
    public Object get(DataFetchingEnvironment environment, DataFetchCallback callback) {
        // Start async operation with callback, such as web or DB lookup
       async(new AsyncCallback() {
           @Override
           public void onComplete(Object result) {
               // Return value to ExecutionStrategy via callback to be collated
               callback.onComplete(value);
           }
       });
    }
};
```

This implementation seems simple enough, although it does have the slight problem of requiring the callback to be manually called, and I'm not sure how to enforce that at the moment.

As for the callback interfaces themselves, I imagine a few simple classes would suffice, although I'm sure we can also use a more established callback library such as JDeferred (actual Promise functionality).

[...]

PR #202, opened 2016-09-09
Implement promises via CompletableFuture

akhahaha:
Related to Asynchronous DataFetcher request (#37)

Massive change that deprecates nearly all key interfaces and moves the SDK requirement to 1.8. Should be considered for a 3.0.0 release.

Error handling and propagation was a little tricky and could probably use some refinement by someone more talented than I, but all tests seem to pass.

Special thanks to @hepin1989 for his input on using CompletableFutures.

### DIFF
--- a/src/main/antlr/Graphql.g4
+++ b/src/main/antlr/Graphql.g4
@@ -16,7 +16,7 @@ operationDefinition:
 selectionSet |
 operationType  NAME? variableDefinitions? directives? selectionSet;
 
-operationType : 'query' | 'mutation';
+operationType : NAME;
 
 variableDefinitions : '(' variableDefinition+ ')';
 
--- a/src/test/groovy/graphql/parser/ParserTest.groovy
+++ b/src/test/groovy/graphql/parser/ParserTest.groovy
@@ -373,4 +373,17 @@ class ParserTest extends Specification {
 
     }
 
+    def "`'query'` can be used as field name"() {
+        given:
+        def input = "{ query }"
+
+        when:
+        Document document = new Parser().parseDocument(input)
+
+        then:
+        document.definitions.size() == 1
+        document.definitions[0] instanceof OperationDefinition
+        document.definitions[0].operation == OperationDefinition.Operation.QUERY
+        assertField(document.definitions[0] as OperationDefinition, "query")
+    }
 }


oexza:
Hello @akhahaha nice work, I think its better to return CompletionStage rather than CompletableFuture, since CompletableFuture is an  implemention of CompletionStage, like its done in Akka and PlayFramework which use this Interface extensively for asyncronous operations. Thanks.

dminkovsky:
Hi @oexza. Can you please provide your thoughts regarding https://github.com/graphql-java/graphql-java/issues/37#issuecomment-245829611 and on?

hepin1989:
@oexza in fact I like the syntax of 
```
return env -> CompletableFuture
```

akhahaha:
@oexza @hepin1989 Lol damn, well just changed to CompletionStage, but I can always revert it. What are the pros and cons?

dminkovsky:
@akhahaha thanks again for your work on this PR. Following the discussion on #37, I think it does make sense to close it for now. 

If you see any other aspects of this project that you could improve—especially docs are you mentioned on #37—please do. We are chatting on [Gitter](https://gitter.im/graphql-java/graphql-java).

dminkovsky:
I think this makes sense to do as a sub project, if anyone would like to do that.

akhahaha:
What would that entail specifically?

dminkovsky:
@akhahaha I'm not sure! I would like to explore concrete use-cases. For example, @pcarrier [wrote the other day](https://gitter.im/graphql-java/graphql-java?at=57d97043512e05f23de23879)

> @dminkovsky for example to run (strictly ordered) operations on netty's event loop
> @dminkovsky including non-blocking I/O 

I'd like see a code sample that shows this usage, so we can figure out how it might be targeted. If you're trying to get each field (and ultimately a whole query) resolved on an event loop, then those tasks need to somehow get on to that event loop. An appropriate execution strategy needs to somehow take this in to account.

I've been Googling around for some pure Java examples of async request processing. Some things I've found:

* [Dropwizard async example](https://github.com/skidder/dropwizard-async-servlet-demo/blob/master/src/main/java/async/server/AsyncServlet.java): uses the `AsyncContext` API to asynchronously handle the request. `AsyncContext#start()` still delegates to an underlying pool of threads. An article that [discusses this API](http://www.javaworld.com/article/2077995/java-concurrency/java-concurrency-asynchronous-processing-support-in-servlet-3-0.html) uses an explicitly declared thread pool.

* [Jersey async documentation](https://jersey.java.net/documentation/latest/async.html): like the Dropwizard example, the crux of this asynchronicity is suspending/resuming connections so that threads can be used for processing instead of maintaining idle connections.

I've not explored Netty and am hoping @pcarrier will provide his use-case.


http://www.nurkiewicz.com/2015/11/which-thread-executes.html
https://www.mailinator.com/tymaPaulMultithreaded.pdf

akhahaha:
It sounds like he might be having similar "issues" from what we discussed in #37. Have you shown him that thread and perhaps this PR?

dminkovsky:
@akhahaha thanks for you response.

> Have you shown him that thread and perhaps this PR?

I @-tagged him in my post above, and just in case just highlighted him in the Gitter chat with a link here. 

> To reiterate, to accomplish the async functionality that I think and he and I were looking for, simply use blocking DataFetchers and then perform the serial execution using the ExecutorService 

Yeah, that seems to be the Java/JVM way. But for some reason this topic keeps coming back up, in issues and in the Gitter chat. People seem to want to minimize blocking and context switching. My Google searching suggested that this approach [might not actually yield performance gains](https://www.mailinator.com/tymaPaulMultithreaded.pdf), but this has come up time and time again. I personally like the simple code style that comes with blocking threads, and appreciate that [Dropwizard claims 30-50k reqs/seq with a default 1024-count thread pool](http://www.dropwizard.io/0.9.0/docs/manual/configuration.html). Then again, maybe people can achieve even more on a tight event loop?

> It's not intuitively async like we'd expect, but the result should be equivalent. I guess it does beg the discussion again of whether or not async functionality like this should be built-in by default.

I think so too. But given the repeated interest I think I closed this PR too hastily. I would like to keep this main project on Java 1.6 to facilitate "legacy" users experimenting with GraphQL on Java. But at the same time perhaps a motivated, interested party will be interested in doing a sub-project.

hepin1989:
async is for not blocking on IO,not for speed.

akhahaha:
I'm reworking this

dminkovsky:
Hi Alan,

Cool, sounds good.

In terms of getting this incorporated into this project or built out as a
subproject, I think it would make sense for people to discuss further to
target particular use cases.

akhahaha:
I think I'm onto a fairly simple solution right now. I've extended DataFetcher into an AsyncDataFetcher abstract class.
```
import java8.util.concurrent.CompletableFuture;

import java.util.concurrent.ExecutionException;

public abstract class AsyncDataFuture implements DataFetcher {
    public Object get(DataFetchingEnvironment environment) {
        try {
            return getDataFuture(environment).get();
            // TODO Handle exceptions
        } catch (InterruptedException e) {
            e.printStackTrace();
        } catch (ExecutionException e) {
            e.printStackTrace();
        }

        return null;
    }

    public abstract CompletableFuture<Object> getDataPromise(DataFetchingEnvironment environment);
}
```

Users can then return their data in a promise in getDataPromise(). This particular CompletableFuture class uses the [streamsupport-cfuture](https://github.com/streamsupport/streamsupport) backport library, which should be compatible at least back to 1.6, although I'm not entirely sure if it's interchangable with the native 1.8 versions.

Unlike my previous implementation, this method should not require changing any other part of the library.

I think there should also be GraphQLAsync class that wraps the default GraphQL in an async wrapper (via callbacks or more CompletableFutures). It should also use an ExecutorService by default in order to achieve the parallel completion necessary for equivalent async performance. I'm not too familiar with the pattern, so I'm not sure if we can provided a default ExecutorService implementation or if users are supposed to provide their own.

Thoughts?

dminkovsky:
@akhahaha that is pretty interesting. I thought about this yesterday and here are my thoughts:

---

> This particular CompletableFuture class uses the streamsupport-cfuture backport library, which should be compatible at least back to 1.6, although I'm not entirely sure if it's interchangable with the native 1.8 versions.

I was thinking this could be a subproject and therefore just simply target 1.8. I think the async/non-blocking crowd is all using 1.8.

> I think there should also be GraphQLAsync class that wraps the default GraphQL in an async wrapper (via callbacks or more CompletableFutures). It should also use an ExecutorService by default in order to achieve the parallel completion necessary for equivalent async performance. I'm not too familiar with the pattern, so I'm not sure if we can provided a default ExecutorService implementation or if users are supposed to provide their own.

I'm not sure what to say about this because I don't know how this would integrate with the bigger picture for people who are doing async/non-blocking request handling already. These users already have a way they're handling concurrency, and I suspect they want this to integrate nicely with the way of handling concurrency. Take for example the [RxJava README](https://github.com/ReactiveX/RxJava/blob/1.x/README.md). There they mention as one of its features:

> * Non-opinionated about source of concurrency (threads, pools, event loops, fibers, actors, etc)

I wonder how they do this... If we have a slew of `CompletableFuture`s that run on the [default runtime pool](www.nurkiewicz.com/2015/11/which-thread-executes.html), async people will definitely not want that. I suspect giving users the option to specify their own `ExecutorService` won't be what they're looking for either, at least not entirely. I think users will want fields to get resolved within their existing source of concurrency. 

So, with all that said, I am not sure how to proceed. It's almost like there's no generic solution to be had? I want to further study the Servlet 3.0 asynchronous handling API and Netty. I think an answer, if there is one, might be found somewhere around there.

Please let me know if any of this is unclear or needs more elaboration. Thank you.


PS. Please see also [8 Asynchronous & Non Blocking](https://ratpack.io/manual/current/async.html#comparison_to_blocking_frameworks__containers)

akhahaha:
> Non-opinionated about source of concurrency (threads, pools, event loops, fibers, actors, etc)

I think my proposed solution is in fact non-opinionated about the source of concurrency. From what I can tell, both `CompletableFuture` and `Observable` are simply fancy callback/runnable handlers. They do not themselves run on another thread.

Some sort of ExecutorService is necessary to achieve the same functionality though. I agree that it does seem undesirable, but the other alternative is to follow the guidelines listed in your Ratpack link, namely:
> "Request handling is organised as a pipeline of asynchronous functions."

Incidentally, I believe this is how my original PR worked, by creating a pipeline of `CompletableFutures` from `DataFetcher`s to `ExecutionResult`. However, I don't believe it would be possible to make this a subproject, nor should we maintain it as a sister project. If we are to use this method, I believe this should be become the default implementation.

Issue #192, opened 2016-08-24
DataFetcher isn't called? 


danielschonfeld:
I have the following schema:

```
query { 
  viewer {
    someInfo(type:X,id:500) {
        name
        address
    }
  }
}
```

I have modelled it so that the arguments `type` and `id` are put on the someInfo `GraphQLObject` and so is the `dataFetcher()`.  I start my java server in debug, breakpoint inside my `DataFetcher` and nothing.  

Is there a reason for that? Am I missing something really obvious or is this a bug?

I am using the `viewer` object to get over the limitation of relay JS allowing only 1 or 0 arguments in root queries.

dminkovsky:
Hi Daniel,

Could you please post your schema definition code? Hard to tell without
seeing that.

Thanks

среда, 24 августа 2016 г. пользователь Daniel Schonfeld написал:

danielschonfeld:
@dminkovsky sure!

```
public class SomeInfoTestSchema {
    public GraphQLSchema getSchema() {
        return schema;
    }

    private GraphQLSchema schema;

    public SomeInfoTestSchema() {
        GraphQLObjectType queryType = newObject()
                .name("Query")
                .description("A query")
                .field(newFieldDefinition()
                        .name("viewer")
                        .type(viewerType())
                        .build()
                )
                .build();

        schema = GraphQLSchema.newSchema()
                .query(queryType)
                .build();
    }

    private GraphQLObjectType viewerType() {
        return newObject()
                .name("Viewer")
                .field(
                    newFieldDefinition()
                    .name("someInfo")
                    .type(someInfoType())
                    .argument(Lists.newArrayList(
                            newArgument()
                                    .name("type")
                                    .type(new GraphQLNonNull(depthType()))
                                    .build(),
                            newArgument()
                                    .name("id")
                                    .type(new GraphQLNonNull(GraphQLInt))
                                    .build()
                    ))
                    .dataFetcher(new SomeInfoFetcher())
                    .build()
                )
                .build();
    }

    private GraphQLObjectType someInfoType() {
        return newObject()
                .description("Some Random Info")
                .name("SomeInfo")
                .field(
                        newFieldDefinition()
                        .name("id")
                        .description("id")
                        .type(GraphQLInt)
                        .build()
                )
                .field(
                        newFieldDefinition()
                        .name("name")
                        .description("Name")
                        .type(GraphQLString)
                        .build()
                )
                .field(
                        newFieldDefinition()
                        .name("type")
[...]

PR #206, opened 2016-09-13
Least Astonishment DataFetchers

dminkovsky:
There has been a lot of activity around the `DataFetcher` interface and its built-in implementations. A lot of this activity reflects user confusion regarding how data fetching works. I believe this confusion has stemmed from:

1. A lack of documentation of the data fetchers, and;
2. The nebulousness of the built-in `DataFetchers` and the way they are configured at the field definition level.

This PR attempts to [lessen the astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) users—particularly new users—face around the `DataFetcher` interface by:

* Shrinking its overall footprint
* Simplifying the built-in `DataFetcher` implementations
* Improving the documentation

This is a breaking PR and requires a bump to 3.0.0.

---

As it stands, the `DataFetcher` interface is very simple. Its a `FunctionalInterface` from `DataFetchingEnvironment` to `Object`. This simplicity, however, belies the reality that:

1. `GraphQLFieldDefinition` provides three ways (`#dataFetcher()`, `#staticValue()`, `#fetchField()`) (the last is undocumented) to configure a field's data fetching functionality.
2. The default data fetcher—`PropertyDataFetcher`—actually does much more than just fetch properties; and
3. The other built-in data fetchers (`StaticDataFetcher` and `FieldDataFetcher`) are undocumented.

The problem with (1) is that `#staticValue(value)` and `#fetchField()` are special public methods that target specific built-in data fetchers (`StaticDataFetcher` and `FieldDataFetcher`, respectively). The convenience these methods present is not obvious. The former is equal to `#dataFetcher(new StaticDataFetcher(value))` and the latter is equal to `#dataFetcher(new FieldDataFetcher(fieldName))`. The existence of these methods also suggests that adding new built-in `DataFetcher` implementations means adding more special methods. This is undesirable considering that `#dataFetcher()` is already so simple.

The problem with (2) is that `PropertyDataFetcher` does more than just fetch a Bean-style property. It is actually a pretty good catch-all attempt at accomplishing _almost_ everything most users might want. As it stands, `PropertyDataFetcher`:

* First attempts to use the source as a `Map`.
* Then attempts to fetch the field from the source as a Bean-style property.
* Finally attempts to fetch the field from the source as a public field, duplicating code from `FieldDataFetcher`.

Although this is simple and gets a lot done given its footprint, the major problem with this approach is that it seems to encourage people to think they should further expand `PropertyDataFetcher` to cover their edge cases instead of implementing their own data fetchers. This library should encourage users—especially Java 8 users—to just write their own fetchers when one of the very simple built-ins does not suffice.

The problem with (3) is that the other data fetchers are undocumented. `StaticDataFetcher` is actually so buried that even `#staticValue()` re-implements its functionality.

---

This PR attempts to solve these issues by:

* Factoring the built-in data fetchers into simple classes that are true to their names
* Removing the special methods from `GraphQLFieldDefinition`
* Improving the documentation


README.md
```diff
@@ -88,7 +88,7 @@ public class HelloWorld {
                         .field(newFieldDefinition()
                                 .type(GraphQLString)
                                 .name("hello")
-                                .staticValue("world"))
+                                .dataFetcher(new StaticDataFetcher("world"))
```
hepin1989: +1

src/main/java/graphql/schema/DataFetcher.java
```diff
@@ -3,5 +3,5 @@
 
 public interface DataFetcher {
 
-    Object get(DataFetchingEnvironment environment);
+    Object get(DataFetchingEnvironment env);
```
hepin1989: +1

src/main/java/graphql/schema/FieldDataFetcher.java
```diff
@@ -2,52 +2,36 @@
...
-     */
     private final String fieldName;
 
-    /**
-     * Ctor.
-     *
-     * @param fieldName The name of the field.
-     */
     public FieldDataFetcher(String fieldName) {
+        assertNotNull(fieldName, "`name` can't be null");
```
hepin1989: fieldName

dminkovsky: Thanks!

src/main/java/graphql/schema/FieldDataFetcher.java
```diff
@@ -2,52 +2,36 @@
...
         this.fieldName = fieldName;
     }
 
     @Override
-    public Object get(DataFetchingEnvironment environment) {
-        Object source = environment.getSource();
+    public Object get(DataFetchingEnvironment env) {
+        Object source = env.getSource();
         if (source == null) return null;
-        if (source instanceof Map) {
```
hepin1989: then we need a `MapDataFetcher`

src/main/java/graphql/schema/MapDataFetcher.java
```diff
@@ -0,0 +1,22 @@
...
+    public MapDataFetcher(String keyName) {
+        assertNotNull(keyName, "`keyName` can't be null");
+        this.keyName = keyName;
+    }
+
+    @Override
+    public Object get(DataFetchingEnvironment env) {
+        Object source = env.getSource();
+        if (source == null || !(source instanceof Map)) return null;
+        return ((Map<?, ?>) source).get(keyName);
```
hepin1989: if ... return map.get else return null

src/main/java/graphql/schema/PropertyDataFetcher.java
```diff
@@ -3,74 +3,52 @@
...
     public PropertyDataFetcher(String propertyName) {
+        assertNotNull(propertyName, "`name` can't be null");
         this.propertyName = propertyName;
     }
 
     @Override
-    public Object get(DataFetchingEnvironment environment) {
-        Object source = environment.getSource();
+    public Object get(DataFetchingEnvironment env) {
+        Object source = env.getSource();
```
hepin1989: make sure `env` is not null

src/main/java/graphql/schema/FieldDataFetcher.java
dminkovsky: This PR includes a [`MapDataFetcher`](https://github.com/dminkovsky/graphql-java/blob/7f483eff91416a67822e615035157453b3b87e24/src/main/java/graphql/schema/MapDataFetcher.java)

src/main/java/graphql/schema/PropertyDataFetcher.java
```diff
@@ -3,74 +3,52 @@
...
-        if (source instanceof Map) {
-            return ((Map<?, ?>) source).get(propertyName);
-        }
-        return getPropertyViaGetter(source, environment.getFieldType());
-    }
-
-    private Object getPropertyViaGetter(Object object, GraphQLOutputType outputType) {
         try {
-            if (isBooleanProperty(outputType)) {
+            if (isBooleanProperty(env.getFieldType())) {
```
hepin1989: I'm using Spring's `PropertyAccessorFactory`

src/main/java/graphql/schema/MapDataFetcher.java
dminkovsky: I don't understand.

hepin1989:
the behavior changes for `First attempts to use the source as a Map.` now it did not.

src/main/java/graphql/schema/PropertyDataFetcher.java
dminkovsky: Why would `env` be null?

dminkovsky:
Yes, now if you expect a `Map` you should use `MapDataFetcher`. A main idea of this PR is that built-in data fetchers should be very specific/discrete in what they do.

src/main/java/graphql/schema/PropertyDataFetcher.java
hepin1989: when user new one.

src/main/java/graphql/schema/PropertyDataFetcher.java
dminkovsky: How does that relate to this?

README.md
```diff
@@ -292,38 +292,48 @@ GraphQLObjectType person = newObject()
...
-```java
+```
+newFieldDefinition()
+    .name("fieldName")
+    .type(GraphQLString)
+    .dataFetcher(new DataFetcher() {
+        @Override
+        Object get(DataFetchingEnvironment env) {
+            // env.getSource() is the value of the surrounding object. 
+            Object source = env.getSource()
```
IamCornholio: Uhh.. rename the variable source here?

src/main/java/graphql/schema/GraphQLFieldDefinition.java
```diff
@@ -185,11 +163,7 @@ public Builder deprecate(String deprecationReason) {
 
         public GraphQLFieldDefinition build() {
             if (dataFetcher == null) {
-                if (isField) {
-                    dataFetcher = new FieldDataFetcher(name);
-                } else {
-                    dataFetcher = new PropertyDataFetcher(name);
-                }
+                dataFetcher = new PropertyDataFetcher(name);
```
IamCornholio: I don't get why this has to be defaulted to a PropertyDataFetcher. Shouldn't this be defaulted to FieldDataFetcher if at all defaulted?

src/test/groovy/graphql/GraphQLTest.groovy
```diff
@@ -45,17 +47,19 @@ class GraphQLTest extends Specification {
                 .field(
                 newFieldDefinition()
                         .name("id")
+                        .dataFetcher(new MapDataFetcher("id"))
```
IamCornholio: Question: Is this interchangeable with a FieldDataFetcher? I'm basically struggling to understand the use case of a MapDataFetcher/PropertyDataFetcher over a FieldDataFetcher. I get that an object has fields that can be "fetched". Those fields could potentially be fetched as properties too using their getters and since its all a json object one could possibly serialize it to a map (not sure how) and cast the object to a map. But when would a user choose one over the other. Sorry if I'm missing the obvious...

README.md
dminkovsky: Oh yeah, thanks!

dminkovsky: Kind of wish this was a better example anyway. Also this section of the docs should document `DataFetchingEnvironment`, right?

src/main/java/graphql/schema/GraphQLFieldDefinition.java
dminkovsky: Yeah, I don't either. You think maybe no default?

src/test/groovy/graphql/GraphQLTest.groovy
dminkovsky: > Is this interchangeable with a FieldDataFetcher?

Nope. [`MapDataFetcher`](https://github.com/dminkovsky/graphql-java/blob/eb75d57f0bbcaed455812218fd81beda028a8379/src/main/java/graphql/schema/MapDataFetcher.java#L17-L21) takes behavior that used to be part of both `PropertyDataFetcher` and `FieldDataFetcher`. Currently, both of these fetchers attempt to use the source as a `Map` before doing anything else. In this PR, neither of these fetchers attempt using the source as a `Map`.

dminkovsky: > I get that an object has fields that can be "fetched".

Well, an object only has fields in the Java sense of the word "field" if it has public fields that can be accessed like in [`FieldDataFetcher`](https://github.com/dminkovsky/graphql-java/blob/eb75d57f0bbcaed455812218fd81beda028a8379/src/main/java/graphql/schema/FieldDataFetcher.java#L26), using reflection.

> Those fields could potentially be fetched as properties too 

Yes, if they can be fetched as properties too (because the object has getter methods that return the field values), you can alternatively use `PropertyDataFetcher`. 

What I am trying to do here is to avoid having something called `PropertyDataFetcher` that actually does 3 different things.

dminkovsky: I just updated the README on this branch to fix the mistake you found. The README also explains what the data fetchers to. They have very discrete, specific functionality.

src/main/java/graphql/schema/GraphQLFieldDefinition.java
IamCornholio: Yes. +1 to no default. Never liked em hidden defaults.

My question however was more from the code change perspective. If isField then instantiate a FieldDataFetcher. Else a PropertyDataFetcher. Now you've removed the isField flag which tells me that this is always going to be a field. Right? And by that logic it should always _sigh_ default to FieldDataFetcher. Did I miss something?

Still maintain that it shouldn't default but I wonder what is the impact for those already in production with this.

src/main/java/graphql/schema/DataFetcher.java
```diff
@@ -3,5 +3,5 @@
 
 public interface DataFetcher {
```
markphilpot: Is it worth adding generics in this PR?  It might make more sense in this context:

```java
public interface DataFetcher<T> {
   T get(DataFetchingEnvironment env);
}```

File: src/main/java/graphql/schema/GraphQLFieldDefinition.java (hunk 9 above)
Comment (markphilpot) [inline src/main/java/graphql/schema/GraphQLFieldDefinition.java]: While I like the explicit nature of `no defaults` I think PropertyDataFetcher is the one to go with.  It makes schema building even more verbose than it is however if there is no default.

Comment (dminkovsky) [inline src/main/java/graphql/schema/GraphQLFieldDefinition.java]: > Actually I don't know how one can assume that the fields or the getters will be publicly accessible

@IamCornholio my assumption is that the developer knows what kind of object is returned by `#getSource()` at every level! The flexibility provided by the current `PropertyDataFetcher` is not because the user doesn't know what type of object they're dealing with, but because `PropertyDataFetcher` is meant to work for maps, fields, and properties.

So, if you know that a field's source will be an object with a publicly accessible field, use `FieldDataFetcher`. If you know it will be a `Map`, use `MapDataFetcher`.

> I think PropertyDataFetcher is the one to go with.

@markphilpot you don't think users are returning `Map`s more often than objects with getters? `Map` is the first test in the current implementation.

File: README.md [hunk 12]
```diff
@@ -292,38 +292,60 @@ GraphQLObjectType person = newObject()
...
+```
+newFieldDefinition()
+    .name("fieldName")
+    .type(GraphQLString)
+    .dataFetcher(env -> env.getSource())
     .build();
-
 ```
 
+`DataFetcher#get()` receives one parameter: the `DataFetchingEnvironment`, which provides field resolution context. The
```
markphilpot: +1

src/main/java/graphql/schema/DataFetcher.java
dminkovsky: I'm having trouble wrapping my mind around this :). Wouldn't this have some pretty far reaching consequences?

hepin1989: @markphilpot In fact I like the more typesafty too:)
See Sangria

src/main/java/graphql/schema/GraphQLFieldDefinition.java
IamCornholio: @markphilpot agreed that verbose is not good. but i'd prefer verbose to ambiguous. the ambiguity i think in using the propertyfetcher is the assumption that field is available with getters when it could very well be a map. or worse no getters or setters :laughing:. improbable but possible.

src/main/java/graphql/schema/DataFetcher.java
dminkovsky: so then `#dataFetcher(DataFetcher fetcher)` becomes `#<T>dataFetcher(DataFetcher<T> fetcher)` and then inside `ExecutionStrategy#resolveField()` you get something like `T result = fetcher.get(...)` and then all the `completeValue()` signatures change. Do you really get much added type safety from this?

dminkovsky: Anyway, I think I'd prefer to leave this for a different PR. This one is already pretty big IMO.

IamCornholio: Generics. I'm looking forward to that one too. No special reason. Like the thought of adding type safety as a feature. It has the potential for users to reduce code using generics. Doesn't sound like a breaking change though. Worst case scenario is that someone using the non typed version gets compilation warnings. I think.

Agree though that the scope of this PR should be limited to the clean up of what a data fetcher does.

src/main/java/graphql/schema/GraphQLFieldDefinition.java
markphilpot: I personally use objects for the type safety, but I can understand a `Map` might be a new user's first choice.  In that light, would vote for being explicit with no default since we are breaking out `Map` and `Property` functionality from `Field`

dminkovsky: No default fetcher at the `GraphQLFieldDefinition.Builder` level sounds
good to me too. I'll update the code and docs to reflect that.

I do sort of wish there could be a good default, but the presence of the
good default has in its own right [caused confusion](
https://github.com/graphql-java/graphql-java/issues/192#issuecomment-242254224
).

вторник, 13 сентября 2016 г. пользователь Mark Philpot написал:

dminkovsky: Done

dminkovsky: Just realized an issue with not having a default fetcher means that
interface field definitions now require a fetcher. How about, instead, we
return `null` fields without a fetcher?

src/main/java/graphql/schema/GraphQLFieldDefinition.java
```diff
@@ -184,13 +162,6 @@ public Builder deprecate(String deprecationReason) {
         }
 
         public GraphQLFieldDefinition build() {
```
IamCornholio: @dminkovsky has raised concern about the fetcher now being required by the field definition interface. This is my response to that. The original comments seem to have been lost in the diff update.

Hmm. I didn't quite get that. Required how?

In any case. I don't know about null. null is a valid value to return I say. I wish there was an undefined :wink:  how about throwing an exception from the build method? I've always thought of the builder as a mechanism to not just build but also validate the object being built.

dminkovsky: > @dminkovsky has raised concern about the fetcher now being required by the field definition interface. ... Required how?

Thanks @IamCornholio. To clarify: we were thinking that perhaps there should be no default `DataFetcher`. If there is no default `DataFetcher`, and the user does not specify a data fetcher, then building a field will cause an exception. That's fine if you're specifying a field for a `GraphQLObjectType`. You want every field to have a `DataFetcher`. But if you're specifying a field for a `GraphQLInterfaceType`, then you want to be able to omit the data fetchers.

IamCornholio: @dminkovsky Ah! Yes I forgot about that.

Perhaps we define (inner or static) a no op data fetcher whose get method will throw a unsupportedoperationexception. The field definition defaults to this exception throwing no op data fetcher. When a user uses the interface they will define an updated field definition I presume and in the process re-write the data fetcher as well.

We could make it a private inner class of the field definition builder class so that no one else may use it as a data fetcher.

src/main/java/graphql/schema/PropertyDataFetcher.java
bpatters: PropertyDataFetcher implies to me that it can fetch via a Method or via the field directly. If your changing this to only support methods then I'd rename it to MethodDataFetcher.

Personally I believe PropertyDataFetcher being capable of fetching from both a method or a field is much more inline with the Generic Bean Construct and more intuitive than splitting this into two different types of data fetchers.

src/main/java/graphql/schema/MapDataFetcher.java
dminkovsky: @bpatters yeah I was confused about that too (that's how it was in the current `PropertyDataFetcher` and `FieldDataFetcher`s).

> So you might as well cast the Map to a Map

So, like this?

```
return ((Map) source).get(keyName);
```

Per your comment and my confusion, why not:

```
return ((Map<String, ?>) source).get(keyName);
```

bpatters: whoops I didn't quote my code in a code tag. Yes:
```return ((Map<String, ?>) source).get(keyName);```

PR #214, opened 2016-09-20
Fix #213

### LINKED ISSUE #213: `'query'` and `'mutation'` cannot be used as field names
The following query fails to parse:

```
query queryDetailQuery($serviceId: ID!, $queryId: ID!, $client: String, $timeFrom: Timestamp!, $timeTo: Timestamp!) {
  service(id: $serviceId) {
    report(from: $timeFrom, to: $timeTo) {
      usageStats(clientName: $client) {
        query(id: $queryId) {
[...]
}
```

It triggers an `org.antlr.v4.runtime.NoViableAltException`.

To the best of my knowledge this should be valid GraphQL.

dminkovsky:
Fix #213

### DIFF
--- a/src/main/antlr/Graphql.g4
+++ b/src/main/antlr/Graphql.g4
@@ -16,7 +16,7 @@ operationDefinition:
 selectionSet |
 operationType  NAME? variableDefinitions? directives? selectionSet;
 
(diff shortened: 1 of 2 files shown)


andimarek:
@dminkovsky looks good to me. I will merge it.

PR #217, opened 2016-10-05
Allow serialization of any Iterable collection as a list

adw45:
Removes the need to write a custom `DataFetcher` when serializing a `Set` to a `GraphQLList`.

### DIFF
--- a/src/main/java/graphql/execution/ExecutionStrategy.java
+++ b/src/main/java/graphql/execution/ExecutionStrategy.java
@@ -93,7 +93,7 @@ public abstract class ExecutionStrategy {
             result = Arrays.asList((Object[]) result);
         }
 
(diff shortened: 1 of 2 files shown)


dminkovsky:
Hey, thanks for the PR.

The implementation change looks good to me. I just want to consider whether this has implications with regard to the spec. Off the top of my head I can't remember anything in the spec that says a GraphQL list's ordering needs to be consistent from execution to execution. 

Regarding the test, it should be re-written as a Spock `Specification`. If you're not familiar with Spock, not problem, I can do that for something this small.

hepin1989:
The only issue is the result of the list will have order changes over time:)

dminkovsky:
Hey I made some cosmetic changes and re-did the test in Spock. To keep things simple I just opened a new PR: https://github.com/graphql-java/graphql-java/pull/235. Thanks!

PR #218, opened 2016-10-06
Minor fixes to hello world in README

gtod:
Just FYI these are the small changes I had to make just now to get the "hello world" example copy pasted out of the README to run against version 2.1.0.


gtod:
Ahh, and only now I see #212 addresses this, closing.

PR #238, opened 2016-10-31
4. parses as FloatValue

pcarrier:
2.9.2 seems somewhat ambiguous as it indicates a list of digits in fractional part but doesn't specify whether it can be empty or not. `graphql-js` has:

```
if (code === 46) { // .
    isFloat = true;

    code = charCodeAt.call(body, ++position);
    position = readDigits(source, position, code);
    code = charCodeAt.call(body, position);
  }
```

### DIFF
--- a/src/main/antlr/Graphql.g4
+++ b/src/main/antlr/Graphql.g4
@@ -118,7 +118,7 @@ NAME: [_A-Za-z][_0-9A-Za-z]* ;
 
 IntValue : Sign? IntegerPart;
 
-FloatValue : Sign? IntegerPart ('.' Digit+)? ExponentPart?;
+FloatValue : Sign? IntegerPart ('.' Digit*)? ExponentPart?;
 
 Sign : '-';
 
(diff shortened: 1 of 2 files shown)


dminkovsky:
Looks good to me. Want me to merge?

pcarrier:
No, review is fine, thanks :)

okorz001:
I think this is a bad change.

Section 2.9.1 explicitly marks a list as optional (list, opt) while 2.9.2 does not. Thus I believe any list is one or more, unless explicitly marked as optional.

However, if you believe that all lists are implicitly optional, then there are more grammar changes to be made, e.g. "1e" should also be a valid float.

pcarrier:
@leebyron any thoughts on this?

dminkovsky:
@okorz001 yes that is very interesting indeed. Thank you!!!

leebyron:
Sorry for not seeing this before. I've been closing out old issues on the graphql branch.

This PR unfortunately introduced a bug into graphql-java where previously the behavior was correct. 

`4.` in a number form should be a parsing error. See https://github.com/graphql/graphql-js/blob/master/src/language/__tests__/lexer-test.js#L570-L582 for examples of test cases which are expected to produce parse errors.

@okorz001 is correct about the difference between `(list)` and `(list, opt)`. http://facebook.github.io/graphql/October2016/#sec-Grammar-Notation explains the meaning of these and illustrates that `(list)` means one or more of that token

Issue #241, opened 2016-11-16
Support GraphQL Schema Language?


brimworks:
The documentation for GraphQL uses a schema language that apparently is still a little in flux given this pull request:

https://github.com/facebook/graphql/pull/90

...and some tooling code is already being created around the schema language:

https://github.com/apollostack/graphql-tools

If I was to send a pull request to support the specification in the above pull request would it have a chance of being accepted?

Thanks,
-Brian

dminkovsky:
Hi @brimworks, thanks for posting.

I would say yeah, definitely, to the the extent that the schema language is supported by the [JS implementation](https://github.com/graphql/graphql-js/blob/v0.8.2/src/language/kinds.js#L55-L68). I'm not sure what that extent is, but it seems to be [parsing](https://github.com/graphql/graphql-js/search?utf8=%E2%9C%93&q=SCALAR_TYPE_DEFINITION).

brimworks:
Thanks @dminkovsky for the green light. I'll be working on this and posting a PR (hopefully by the end of this week).

Thanks,
-Brian

dminkovsky:
Cool, thanks Brian. Post if you have any questions. Are you familiar with Spock?

brimworks:
Not to familiar with Spock, but I'm adding tests to ParserTest.groovy and it seems pretty straight forward. So far I've added to the ANTLR grammar the new syntax, then modified GraphqlAntlrToLanguage.java to generate the new "language" nodes of the AST. I can push my work in progress if you want to start the review process early.

dminkovsky:
Cool sounds great. No, please take your time :). Thank you.

dminkovsky:
By the way, a related issue if you have any time and interest is #126. I think it just needs tests ported-adapted from the JS implementation and any bug fixes that result. Would really bolster the tooling.

brimworks:
I quickly skimmed that issue, but it was unclear to me what format the printer prints in. I'm ASSUMING it iterates over the AST and pretty prints the original input in GraphQL syntax. If that is the case, then all these new AST symbols I'm adding will have a direct impact on the pretty printer (since there will be a host of new symbols to handle).

Unfortunately, the work I'm doing currently doesn't require such a pretty printer, but I do see value in it. So, I probably won't have time to contribute to that effort in the near future.

dminkovsky:
Cool, thank you. Yeah, that code will have to be updated to accommodate the
AST symbols you're adding. It won't be anything.

[...]

PR #243, opened 2016-11-17
Initial take on adding GraphQL Schema parsing.

brimworks:
The definitions come from this pending pull request:

https://github.com/facebook/graphql/pull/90

Specifically the grammar comes from:

https://github.com/facebook/graphql/blob/de5433206643b2ec9e87e18e142f2e630bfa05b9/spec/Appendix%20B%20--%20Grammar%20Summary.md

Note that the grammar is still not solidified, but tooling is already
being created that is based on this grammar. For example:

https://github.com/apollostack/graphql-tools

My hope is to enable schema-first development for Java by creating
tooling similar to the above, but for Java.

Note that I had to change the 'NAME' token to be a 'name' rule. This
was necessary to support "keywords" throughout the grammar. For
example, without this change there are tests that define "type" and
"scalar" as field names in various queries, and these would fail to
parse.

See https://github.com/graphql-java/graphql-java/issues/241


src/main/java/graphql/language/ObjectTypeDefinition.java
```diff
@@ -0,0 +1,68 @@
+package graphql.language;
+
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ObjectTypeDefinition extends AbstractNode implements TypeDefinition {
+    private String name;
+    private List<Type> implementz = new ArrayList<Type>();
```
brimworks: implements is a keyword, otherwise I would use that name :).

src/main/java/graphql/parser/GraphqlAntlrToLanguage.java
```diff
@@ -109,14 +118,13 @@ public Void visitOperationDefinition(GraphqlParser.OperationDefinitionContext ct
         OperationDefinition operationDefinition;
         operationDefinition = new OperationDefinition();
         newNode(operationDefinition, ctx);
-        newNode(operationDefinition, ctx);
```
brimworks: This looked like a harmless bug, so I took the liberty to fix it.

src/main/java/graphql/parser/GraphqlAntlrToLanguage.java
```diff
@@ -132,7 +140,7 @@ public Void visitOperationDefinition(GraphqlParser.OperationDefinitionContext ct
         } else if (operationTypeContext.getText().equals("mutation")) {
             return OperationDefinition.Operation.MUTATION;
         } else {
-            throw new RuntimeException();
+            throw new RuntimeException("InternalError: unknown operationTypeContext="+operationTypeContext.getText());
```
brimworks: This was a helpful message when I was debugging... but it does make me wonder if the grammar should be handling the mutation and query keywords. Now that 'name' is a rule rather than a token, we should be able to use the mutation and query keywords to the grammar.

src/main/java/graphql/parser/GraphqlAntlrToLanguage.java
dminkovsky: thanks that's been bugging me :)

brimworks:
Ya, that throw new RuntimeException() code should never be hit now that you have added it to the grammar... I was thinking about adding that to the grammar, but I like to keep my PRs focused.

Thanks!
-Brian

dminkovsky:
Really appreciate the focus. I may have gotten a bit carried away there with the excitement of using the GitHub branch/network functionality.

PR #257, opened 2016-12-06
Allow classes implementing an interface to be referenced in graphql queries

fderose:


fderose:
I don't know why StarWarsIntrospectionTests is failing in CI. With the f41a062 commit, it runs locally for me.

PR #285, opened 2016-12-22
omit null variable values

Jimexist:


src/main/java/graphql/GraphQL.java
```diff
@@ -72,7 +70,7 @@ public ExecutionResult execute(String requestString, String operationName, Objec
             RecognitionException recognitionException = (RecognitionException) e.getCause();
             SourceLocation sourceLocation = new SourceLocation(recognitionException.getOffendingToken().getLine(), recognitionException.getOffendingToken().getCharPositionInLine());
             InvalidSyntaxError invalidSyntaxError = new InvalidSyntaxError(sourceLocation);
-            return new ExecutionResultImpl(Arrays.asList(invalidSyntaxError));
+            return new ExecutionResultImpl(Collections.singletonList(invalidSyntaxError));
```
dminkovsky: This is 100% a good change but maybe you could accumulate these on something like a "cleanup" branch and submit them without functionality changes. Or even submit them as one-off PRs. It would make evaluating PRs easier. 

Same for the comment addition below.

src/main/java/graphql/execution/ValuesResolver.java
```diff
@@ -9,11 +9,13 @@
...
 
-
     public Map<String, Object> getVariableValues(GraphQLSchema schema, List<VariableDefinition> variableDefinitions, Map<String, Object> inputs) {
         Map<String, Object> result = new LinkedHashMap<String, Object>();
         for (VariableDefinition variableDefinition : variableDefinitions) {
-            result.put(variableDefinition.getName(), getVariableValue(schema, variableDefinition, inputs.get(variableDefinition.getName())));
+            Object variableValue = getVariableValue(schema, variableDefinition, inputs.get(variableDefinition.getName()));
+            if (variableValue != null) {
+                result.put(variableDefinition.getName(), variableValue);
+            }
```
dminkovsky: I am wondering what the compatibility implications are here. I can't imagine any, but I can't imagine many things :).

Related to the above question: Why would you want this to be the way this works? I mean, this method returns the resulting `Map`, and the client then `get()`s from that `Map`. With our without the PR, they'd get `null` for missing keys. Right?

Jimexist: Sorry for the lack of explanation here. Indeed in the situation you mentiond here it didn't matter. But in cases the map is converted using jackson the key being absent is different than null value. One of the cases will trigger an unknown field exception.

src/main/java/graphql/GraphQL.java
Jimexist: You are right for this specific line change i should've put into a separate pull request that fixes similar things. Will be more careful next time.

src/main/java/graphql/execution/ValuesResolver.java
dminkovsky: Ah okay, thank you. That should have been obvious, actually. Makes sense.

But, so then what about people who may be relying on the behavior that's currently in place (my compatibility question)? My feeling is that someone might want the current behavior while someone else might want this behavior. Is there some reason this behavior is more "correct"?

src/main/java/graphql/GraphQL.java
dminkovsky: Thank you!

src/main/java/graphql/execution/ValuesResolver.java
dminkovsky: Please disregard my previous comment, I was confused. 

But I am still confused. The `input` parameter may or may not be the result of deserialization, but at this stage of execution it doesn't matter where that `Map` came from. I'm sorry I'm not seeing this at all. A concrete example—or even better a failing test—would help a lot.

Jimexist: will do - and i'll update this PR soon. thanks for the comment.

dminkovsky:
Thank you. Looking forward.

чт, 29 дек. 2016 г. в 0:39, Jiayu Liu <notifications@github.com>:

Jimexist:
@dminkovsky I tried but failed to find a concrete example of replaying my issues - i'll come up with a better test case should i encounter my original problem again. Sorry for the mistake - and please feel free to close this PR.

dminkovsky:
No problem! Thank you for posting and apologies for my delay in responding. I'll be faster if the PRs are more direct and illustrate the problem directly. Will close this in a bit if it remains idle.

PR #297, opened 2017-01-12
Fix #296: Introspection on default value for input object is not following GraphQL spec

### LINKED ISSUE #296: Introspection on default value for input object is not following GraphQL spec
I defined an input object with a `HashMap` as default value. Introspection on the default value gives the `toString()` value of the `HashMap` (see `graphql.introspection#__InputValue`). However, [GraphQL specification](http://facebook.github.io/graphql/#sec-The-__InputValue-Type) requires the string to be encoded using GraphQL language.

As a result, GraphiQL would fail because of the incorrect introspection result.

dminkovsky:
Relevant lines: https://github.com/graphql-java/graphql-java/blob/6b94b458f6738bde331444d307d1081aecb6ab10/src/main/java/graphql/introspection/Introspection.java#L84-L89

I think this will require https://github.com/graphql-java/graphql-java/issues/126

### DIFF
--- a/src/test/groovy/graphql/Issue296.groovy
+++ b/src/test/groovy/graphql/Issue296.groovy
@@ -1,0 +1,41 @@
+package graphql
+
+import spock.lang.Specification
+
+import static graphql.Scalars.GraphQLString
+import static graphql.schema.GraphQLArgument.newArgument
+import static graphql.schema.GraphQLFieldDefinition.newFieldDefinition
+import static graphql.schema.GraphQLInputObjectField.newInputObjectField
+import static graphql.schema.GraphQLInputObjectType.newInputObject
+import static graphql.schema.GraphQLObjectType.newObject
+import static graphql.schema.GraphQLSchema.newSchema
+
+class Issue296 extends Specification {
+
+    def "test"() {
+
+        def graphql = GraphQL.newGraphQL(newSchema()
+            .query(newObject()
+                .name("Query")
+                .field(newFieldDefinition()
+                    .name("field")
+                    .type(GraphQLString)
+                    .argument(newArgument()
+                        .name("argument")
+                        .type(newInputObject()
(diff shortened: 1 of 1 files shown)


src/test/groovy/graphql/Issue296.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+import static graphql.Scalars.GraphQLString
+import static graphql.schema.GraphQLArgument.newArgument
+import static graphql.schema.GraphQLFieldDefinition.newFieldDefinition
+import static graphql.schema.GraphQLInputObjectField.newInputObjectField
+import static graphql.schema.GraphQLInputObjectType.newInputObject
+import static graphql.schema.GraphQLObjectType.newObject
+import static graphql.schema.GraphQLSchema.newSchema
+
+class Issue296 extends Specification {
```
dminkovsky: Use `TestUtil.schemaWithInputType(inputType)`

src/test/groovy/graphql/Issue296.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+    def "test"() {
+
+        def graphql = GraphQL.newGraphQL(newSchema()
```
dminkovsky: Use `TestUtil.schemaWithInputType(inputType)`

rayngwf:
This unit test checks the exact reported issue. Thanks.

To make unit test for default value more complete, we should include testing with other value types as well, including list and enum etc.

dminkovsky:
Yes definitely.

чт, 12 янв. 2017 г. в 20:50, rayngwf <notifications@github.com>:

bbakerman:
This PR fixes this issue and adds a shed load more functionality at the same time

https://github.com/graphql-java/graphql-java/pull/388/files

bbakerman:
This is no longer needed as is

PR #319, opened 2017-02-07
Remove prefix string from getMessage

gando999:
Suggestion: Can we remove the prefix string from the getMessage method of this exception wrapper? It makes things a little tricky and can expose the toString implementation of an exception where it is not desired.

### DIFF
--- a/src/main/java/graphql/ExceptionWhileDataFetching.java
+++ b/src/main/java/graphql/ExceptionWhileDataFetching.java
@@ -17,10 +17,9 @@ public class ExceptionWhileDataFetching implements GraphQLError {
         return exception;
     }
 
-
     @Override
     public String getMessage() {
-        return "Exception while fetching data: " + exception.toString();
+        return exception.getMessage();
     }
 
     @Override


exbe:
But I would prefer to keep "actual" message, it is a burden to deal from raw exception without a hint.

gando999:
Sure, I was just found that although I could override toString() on my exception (raised during data fetching), I still did not want to deal with the 'Exception while fetching data' string being appended to the message I raised the exception with. Is there another known way to recover my original message without resorting to string parsing? (I am still new to GraphQL so I am probably missing something)

exbe:
It should be okay to use 'ExceptionWhileDataFetching.getException()'  in most situations instead `getMessage()` for your case. It should return untouched exception with your original message.

gando999:
Ok, its not part of the GraphQLError interface but I guess I can use the getErrorType() to perform an explicit cast. Just felt a bit awkward.

exbe:
@gando999 yes, 'getErrorType ()' is a better way. Another solution would be to introduce fair localization for all strings.

I am closing this PR for now. Feel free to re-open it anytime.

gando999:
Yep - no worries - thanks for looking

PR #321, opened 2017-02-08
Update 315

exbe:
@bbakerman this #315 with respect to #317 
Could you review it once? 

No need to close #315 - it should be updated automatically once this one is taken.


src/main/java/graphql/Scalars.java
```diff
@@ -22,13 +22,45 @@
     private static final BigInteger SHORT_MAX = BigInteger.valueOf(Short.MAX_VALUE);
     private static final BigInteger SHORT_MIN = BigInteger.valueOf(Short.MIN_VALUE);
 
+    private static boolean isWholeNumber(Object input) {
```
Jimexist: why not just `instanceOf Number`?

exbe:
Closing this one, sine #315 has been merged already

Issue #227, opened 2016-10-18
Shouldn't type builders validate names according to the spec?


metacosm:
The spec states that valid names must conform to the `/[_A-Za-z][_0-9A-Za-z]*/` regex (http://facebook.github.io/graphql/#sec-Names) but it doesn't seem like this validation is applied by this implementation.

dminkovsky:
That section refers to query documents (the spec does not deal very much with schema definition, if at all). We parse query documents with ANTLR, and [names are validated per the spec](https://github.com/graphql-java/graphql-java/blob/v2.1.0/src/main/antlr/Graphql.g4#L117).

Please see (2) in my response on our [previous issue](https://github.com/graphql-java/graphql-java/issues/226#issuecomment-254669228). It's important to differentiate between the query document and schema definition.

dminkovsky:
That said, the builders are kind of lacking IMO. For example, unlike the JS implementation, our builders do not validate that objects implement interfaces correctly. It would be nice if this were improved. Names might also be validated in the builders.

dminkovsky:
> the spec does not deal very much with schema definition

In terms of how that definition system is implemented.

metacosm:
I agree that the spec for names addresses only query documents. However, if the schema doesn't conform to this naming scheme, you won't be able to query your data. As an example, if you attempt to use a schema that doesn't follow this naming format, GraphiQL will throw an error. A possible spec oversight but shouldn't the schema produce names that can be queried, regardless?

dminkovsky:
Yeah, validation (name and other properties) is definitely a deficiency. In addition to this deficiency, there are several other problems that have been identified with the schema definition system:

* #215
* #198 and #28 
* #172 

I've explored these and discovered that these are all related in one way or another.

I'm not entirely sure how best to resolve them at the moment. On one hand, it's easy to iterate on the simpler problems like validating names and interfaces. On the other hand, it's not as simple to work through the issues surrounding type references and hashCode()/equals(). But anyway, while all these things are issues, I wouldn't call them blocking issues and have been thinking about how to best address them while working more with the library in my applications.

Please feel free to submit PRs if you would like to iterate on these.

Thanks.

okorz001:
+1 for detecting illegal fields at schema creation time instead of query execution time. There is no value in declaring a field that cannot be queried.

Issue #215, opened 2016-09-21
GraphQLObjectType does not validate interface fields


dminkovsky:
The reference implementation ensures that object types which inherit interfaces correctly define the interfaces' fields. This implementation does not, yet.

PR #323, opened 2017-02-08
Validate name field matches current spec's conventions

deruf:
Validation of schema names on the server side (not in the query document)

### DIFF
--- a/src/main/java/graphql/Assert.java
+++ b/src/main/java/graphql/Assert.java
@@ -15,4 +15,18 @@ public class Assert {
         return c;
     }
 
+    private static final String invalidNameErrorMessage = "Name must be non-null, non-empty and match [_A-Za-z][_0-9A-Za-z]*";
+    /**
+     * Validates that the Lexical token name matches the current spec.
+     * currently non null, non empty, 
+     * @param name - the name to be validated.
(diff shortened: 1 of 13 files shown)


dminkovsky:
Thanks for this. More schema definition validation is necessary and this is a good step in that direction!

deruf:
Please don't accept this yet, I found a problem and have some additional validation to add.

dminkovsky:
Sounds good. Thank you!

dminkovsky:
Just dropping a few related notes:

* #227 can be closed after this is merged
* #215 is also good for follow up

bbakerman:
@deruf - do you have follow up commits you want?  This seems worthwhile as it is?

src/main/java/graphql/Assert.java
```diff
@@ -15,4 +15,18 @@
         return c;
     }
 
+    private static final String invalidNameErrorMessage = "Name must be non-null, non-empty and match [_A-Za-z][_0-9A-Za-z]*";
```
bbakerman: https://facebook.github.io/graphql/#sec-Names

Nice!

bbakerman:
I am going to merge this PR and follow fixes can come in another one

PR #358, opened 2017-04-05
Support for user defined subscription type

vaant:
Adding support for subscriptions with an implemenation to enable adding a user defined subscriptionType. Subscriptions can be thought of as mutations by a different name and with a different purpose and, as such, the implementation here is the same as that of mutations.

For more see: http://graphql.org/blog/subscriptions-in-graphql-and-relay/


andimarek:
@vaant Thanks for this PR. 

How does this PR relate to https://github.com/facebook/graphql/pull/267 ?

kaqqao:
I must say I love the simplicity of this implementation.

andimarek:
@vaant Would you mind giving us the right to modify the PR (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) ? We wanna add some documentation for example before we merge it.

vaant:
@andimarek edits from maintainers should be allowed.  As far as graphql#267, I would say this is a partial implementation.  From my reading, there isn't anything that would preclude this as an initial implementation for the Subscription system but the corresponding Events system is entirely ignored.  I suppose this limits usefulness, at least in regards to the RFC, but does provide a very simple way for users to set and manage their own SubscriptionType.

andimarek:
@vaant thanks a lot!

PR #361, opened 2017-04-07
use java.util.Base64, fix #251

### LINKED ISSUE #251: graphql.relay.Base64#fromBase64 throws ArrayIndexOutOfBoundsException on unicode string
```
Base64.fromBase64("🏃")
```

when the input is an invalid unicode, the error will be:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 57283
	at javax.xml.bind.DatatypeConverterImpl.guessLength(DatatypeConverterImpl.java:659)
	at javax.xml.bind.DatatypeConverterImpl._parseBase64Binary(DatatypeConverterImpl.java:692)
	at javax.xml.bind.DatatypeConverterImpl.parseBase64Binary(DatatypeConverterImpl.java:434)
	at javax.xml.bind.DatatypeConverter.parseBase64Binary(DatatypeConverter.java:342)
	at graphql.relay.Base64.fromBase64(Base64.java:26)
```

Jimexist:
use built-in version


andimarek:
Hi,

thanks for the PR.

It seems like you also reformatted quite a few things: Could you please change it so that it only changes the Base64 related stuff?

Thanks!

Jimexist:
@andimarek sorry about the format - it is fixed now

src/main/java/graphql/relay/Base64.java
```diff
@@ -1,24 +1,25 @@
...
 import java.nio.charset.Charset;
-import javax.xml.bind.DatatypeConverter;
-import java.io.UnsupportedEncodingException;
-
 
+/**
+ * @deprecated use {@link java.util.Base64} instead.
+ */
+@Deprecated
 public class Base64 {
```
andimarek: Is this class used anywhere else? If not we should just delete it.

Jimexist: not in this package - but it is always good to be cautious before deleting a public class. having said that, it is fair enough to delete it in a breaking release.

andimarek: I agree in general but the next release is a breaking release. So please delete it. :-) thanks

src/main/java/graphql/relay/Base64.java
```diff
@@ -1,24 +1,25 @@
...
 
     private Base64() {
     }
 
+    private static final Charset UTF8 = Charset.forName("UTF-8");
```
jplock: This could be replaced with `StandardCharsets.UTF_8`

Jimexist: ok

src/main/java/graphql/relay/Base64.java
Jimexist: ok

Jimexist:
@andimarek @jplock this is fixed now

PR #363, opened 2017-04-07
make schema util static, fix #362

### LINKED ISSUE #362: make graphql.schema.SchemaUtil static?
Use private ctor and make all methods static

Jimexist:
#362


andimarek:
Hi,

thanks for your PR.

It seems like you already toke advantage of Java 8: Could you create a separate PR for these changes to make it more clean?

About the change: Why making schema util static? 
Thanks!

Jimexist:
@andimarek the fact that it is named with `Util` makes it useful in many other occasions, and making it stateless (i.e. static utility functions only) will make it easier to reason about thread safety, etc. And i think it is just a matter of convention that utility functions are static and they operate solely on the inputs, not the containing object's state.

Jimexist:
@andimarek separation of PRs makes sense - so I created this https://github.com/graphql-java/graphql-java/pull/364

andimarek:
@Jimexist The problem with static calls ist that you can't mock it easily and it is not easily and it is general a very "strong" dependency. 

Therefore I never use static methods anymore, except very very simple methods, even when I call the class *Util.

And SchemaUtil is definitely not simple, there I prefer to keep it a "real" class.

Jimexist:
i don't have a strong argument against mock-ability - so feel free to close this if you believe this is unnecessary.

bmsantos:
I like that it is static so that the methods can be statically imported, making code much cleaner and leaner and simple to parse by the human eye.

Testing a static method can be problematic but shouldn't be the main reason to discard "static" methods altogether.

Jimexist:
@bmsantos @andimarek 

I agree that

👍:
- static functions are great for static importing
- static functions are easier to reason since they are meant to be stateless, and *pure*

👎:
- some times static importing makes code less readable
- static classes are bad for mocking

My opinion is that:
- personally I am not a big fan of big util class, and I don't think it is good practice to name any class with xxxUtil (no offense intended), since it gets big pretty quickly and stuff gets dumped into it carelessly. it is generally a good idea in terms of maintainability to tear it apart into specific named classes e.g. (`GraphQlTypeCollector` or `TypeReferenceResolver`) and have them made mockable with interface + impl. class

andimarek:
I am closing this PR, because it will not get merged. 

I know that this is to some degree a matter of taste, but based on my experience I avoid static methods as much as I can, except maybe very very simple little helper methods. 

It is simply to dangerous to have such a strong, not mockable dependency, which is also somehow "hidden" if you use static imports. It never ends well in the long-term. 

But thanks @Jimexist for the PR and your help!

PR #436, opened 2017-05-15
Cleanup code formatting in README

benmccann:
To see how this looks rendered, see https://github.com/benmccann/graphql-java/blob/patch-2/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,11 +226,11 @@ Example:
 
 ```java
 GraphQLInputObjectType inputObjectType = newInputObject()
-        .name("inputObjectType")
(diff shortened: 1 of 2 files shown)


benmccann:
I don't see that file. Where does it live?

benmccann:
I've updated README.next.md as requested

Is there any reason that file lives in `src/test/groovy/readme/` instead of the root directory with the other README?

andimarek:
@benmccann Thanks for this PR, but we migrated to readthedocs recently, so this is obsolete. Thanks again!

Issue #446, opened 2017-05-20
Scalar handing changes


andimarek:
Currently our Scalar handling is not correct in several aspects:

- `parseLiteral` should return null if it can't be parsed and not throw an Exception. `ArgumentsOfCorrectType` will check that the type is correct.

-  It is not allowed to throw away some infos when coercing. For example "124.3" is not a valid value for Int. Instead an Error must be thrown. See https://github.com/graphql-java/graphql-java/issues/171 and  https://github.com/graphql/graphql-js/issues/827

- NonNull Handling Error: https://github.com/graphql-java/graphql-java/issues/438

andimarek:
Here is a test illustrating one of the current problems: https://github.com/graphql-java/graphql-java/commit/be44723b8e5fb1201a0bf9adc2ad575d372307fc

`parseLiteral` of `GraphQLInt` throws an exception instead of returning null.

andimarek:
#455 fixes everything except NonNull Handling error (#438)

andimarek:
closing it ... #438 is still open and will be handled there

PR #455, opened 2017-05-23
Scalar changes

andimarek:
This PR addresses #446:

*  It adds a ton of tests: each scalar has now it own test file.

* `parseLiteral` returns null if not valid.

* `serialize/parseValue` throws Exception if not valid.

*  an input is only valid if it can converted without lost of information: e.g. "123.0" is ok for Float, but not "123.3"

* GraphQLBoolean was changed to map 0 to false and every other number to true. Before that every positive number was true.

* Everywhere where a number is expected we accept now consistently `Number`.

* null is never a valid input for `parseValue/serialize/parseLiteral` so it is never tested or paid attention to.

* `GraphQLEnumType` coercions is also changed to be aligned with the rest: throws Exception for invalid input, `null` is never accepted or returned, `parseLiteral` returns null for an invalid literal. This means Enums are more likely to break something, because before `GraphQLEnumType` would just return silently `null`.


andimarek:
@bbakerman I know it is a rather big PR, but I would appreciate if you could have a look. Thx

PR #484, opened 2017-06-03
remove unused `extensions` field from ExecutionResult(Impl)

andimarek:
As discussed here: https://github.com/graphql-java/graphql-java/pull/480#discussion_r119765358


andimarek:
This is a breaking change because we removed the field completely.

PR #486, opened 2017-06-05
printer improvments

andimarek:
aligns schema printer and AstPrinter regarding indention and new lines

adds test which completes a full round trip with SchemaPrinter,
Introspection and AstPrinter

changes IntrospectionResultToSchema to only create a SchemaDefinition
if needed because non default names are used


andimarek:
not sure this is a breaking change because I changed the output format of `SchemaPrinter` a bit

PR #507, opened 2017-06-16
Reduce code duplication in NoOpInstrumentation

tklovett:
Here is one granular portion of my earlier PR #504 Refactorings.


andimarek:
@tklovett thanks

PR #517, opened 2017-06-21
Correct execution example

lanyonm:
There's a very small error in the execution example in the docs. I believe this corrects it.


andimarek:
thank you!

PR #531, opened 2017-06-24
Preparsed document provider/cache to avoid costly parse and validation

LarsKrogJensen:
Query parsing and validation can be costly in terms of execution performance, this PR introduces a preparsed provider/cache that allows implementations to hook in a cache to avoid re- parse and validation.

Another possible use case could be white listing queries.

One showstopper for this might be that it requires the Document tree to be thread safe, and from what I can see it is not modified after its been created.

### DIFF
--- a/src/main/java/graphql/GraphQL.java
+++ b/src/main/java/graphql/GraphQL.java
@@ -10,6 +10,9 @@ import graphql.execution.instrumentation.InstrumentationContext;
 import graphql.execution.instrumentation.NoOpInstrumentation;
 import graphql.execution.instrumentation.parameters.InstrumentationExecutionParameters;
 import graphql.execution.instrumentation.parameters.InstrumentationValidationParameters;
+import graphql.execution.preparsed.NoOpPreparsedDocumentProvider;
+import graphql.execution.preparsed.PreparsedDocumentEntry;
+import graphql.execution.preparsed.PreparsedDocumentProvider;
 import graphql.language.Document;
 import graphql.language.SourceLocation;
 import graphql.parser.Parser;
@@ -39,6 +42,7 @@ public class GraphQL {
     private final ExecutionStrategy subscriptionStrategy;
     private final ExecutionIdProvider idProvider;
     private final Instrumentation instrumentation;
+    private final PreparsedDocumentProvider preparsedDocumentProvider;
 
 
     /**
@@ -79,7 +83,7 @@ public class GraphQL {
      */
     @Internal
     public GraphQL(GraphQLSchema graphQLSchema, ExecutionStrategy queryStrategy, ExecutionStrategy mutationStrategy) {
-        this(graphQLSchema, queryStrategy, mutationStrategy, null, DEFAULT_EXECUTION_ID_PROVIDER, NoOpInstrumentation.INSTANCE);
+        this(graphQLSchema, queryStrategy, mutationStrategy, null, DEFAULT_EXECUTION_ID_PROVIDER, NoOpInstrumentation.INSTANCE, NoOpPreparsedDocumentProvider.INSTANCE);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


src/main/java/graphql/execution/ExecutionContextBuilder.java
```diff
@@ -130,7 +130,7 @@ public ExecutionContext build() {
                 fragmentsByName,
                 operation,
                 variableValues,
-                root,
-                context);
+                context,
+                root);
     }
 }
```
andimarek: This is fixed in #532 .

src/main/java/graphql/GraphQL.java
```diff
@@ -289,18 +300,27 @@ public ExecutionResult execute(ExecutionInput executionInput) {
     }
 
     private ExecutionResult parseValidateAndExecute(ExecutionInput executionInput) {
-        ParseResult parseResult = parse(executionInput);
-        if (parseResult.isFailure()) {
-            return toParseFailureExecutionResult(parseResult.getException());
+        PreparsedDocumentEntry preparsedDoc = preparsedDocumentProvider.get(executionInput.getQuery(), query -> {
```
andimarek: can you write tests for the lamba expression?

andimarek: also: this should be extracted in an extra method

src/main/java/graphql/execution/preparsed/NoOpPreparsedDocumentProvider.java
```diff
@@ -0,0 +1,13 @@
+package graphql.execution.preparsed;
+
+
+import java.util.function.Function;
+
+public class NoOpPreparsedDocumentProvider implements PreparsedDocumentProvider {
+    public static final NoOpPreparsedDocumentProvider INSTANCE = new NoOpPreparsedDocumentProvider();
+
+    @Override
+    public PreparsedDocumentEntry get(String query, Function<String, PreparsedDocumentEntry> compute) {
```
andimarek: this should be unit tested

src/main/java/graphql/execution/preparsed/PreparsedDocumentEntry.java
```diff
@@ -0,0 +1,41 @@
...
+import graphql.language.Document;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * An instance of a preparsed doucument entry represents the result of a query parse and validation, like
+ * an either implementation it contains either the correct result in th document property or the errors.
+ */
+public class PreparsedDocumentEntry {
```
andimarek: this instance fields should be final

src/main/java/graphql/GraphQL.java
```diff
@@ -289,18 +300,27 @@ public ExecutionResult execute(ExecutionInput executionInput) {
...
+                final List<ValidationError> errors = validate(executionInput, document);
+                if (!errors.isEmpty()) {
+                    return new PreparsedDocumentEntry(errors);
+                }
+
+                return  new PreparsedDocumentEntry(document);
+            }
+        });
+
+        if (preparsedDoc.hasErrors()) {
```
andimarek: preparsedDoc can be null, right?

src/test/groovy/graphql/execution/preparsed/PreparsedTest.groovy
```diff
@@ -0,0 +1,138 @@
...
+        def query = """
+        query HeroNameAndFriendsQuery {
+            hero {
+                id
+            }
+        }
+        """
+
+        //
+        // for testing purposes we must use SimpleExecutionStrategy under the covers to get such
```
andimarek: no need for that comment imho

LarsKrogJensen:
I would argue that a cache is out of scope of this project, as you said even a simple LRU cache is tricky. There are many quality open source caching solutions out there, looking especially at caffeine. With only a few lines of code you, as a user, have a quality cache in place

Whats more is that to achieve high cache hit ratio it is recommended that field arguments are passed in as variables instead of directly in the query.

I will update the docs (executions) with usage and submit a PR.

src/main/java/graphql/execution/preparsed/NoOpPreparsedDocumentProvider.java
LarsKrogJensen: Check

src/main/java/graphql/execution/preparsed/PreparsedDocumentEntry.java
LarsKrogJensen: Check

src/main/java/graphql/GraphQL.java
LarsKrogJensen: It can if the document provider for what ever reason returns null. We could either let it throw a NPE or wrap it into a more descriptive exception.
The lambda will always return a non-null result though.

src/main/java/graphql/GraphQL.java
LarsKrogJensen: Converted lambda to function, will look into if I can target this function with test.

LarsKrogJensen:
Updated the execution documentation on how to use it.

PR #561, opened 2017-07-08
Move boolean inversion to correct place

backjo:
While looking around the code for hasInvalidLocation, I noticed that the inversion appeared to be incorrect inside of the InlineFragment use case - I believe it should be on the outside of the parentheses like the other cases, not inside on the .validLocations() check)


andimarek:
@backjo thank you for your PR. 

Could you provide a test to verify that this is the correct implementation? That would be great.

backjo:
Yep - will do!

PR #587, opened 2017-08-02
Optimize the case when there are no directives

justinsb:
directivesByName is relatively expensive - it builds a Map of all
directives and then fetches the directive by name.  In the common case
when there are no directives at all, we can immediately return null.


andimarek:
thanks

PR #588, opened 2017-08-02
Optimize fields with no arguments

justinsb:
For the common case of fields with no arguments, we can early exit out
of getArgumentValues.


src/main/java/graphql/execution/ValuesResolver.java
```diff
@@ -46,6 +46,11 @@
 
 
     public Map<String, Object> getArgumentValues(List<GraphQLArgument> argumentTypes, List<Argument> arguments, Map<String, Object> variables) {
+        if (argumentTypes.isEmpty()) {
+            // Fast-path for common case of simple fields with no arguments
```
andimarek: could you remove that comment? we avoid commenting without a very good reason.

justinsb:
Thanks - removed that comment and just amended the commit (as the change is so small)

andimarek:
Thank you! We will merge it soon.

PR #589, opened 2017-08-02
Avoid defensive copy in GraphQLFieldDefinition::getArguments

justinsb:
By making arguments an unmodifiable list, we can avoid the need for a
defensive copy.


justinsb:
Similar to #586

andimarek:
thanks!

PR #714, opened 2017-09-16
Improving exception for duplicated field definition

jdorleans:
At first, I found difficult to understand what the exception meant and where to find the problem to fix. So, I'm providing clearer description for that. Supposing the following schema:
```
type User {
  ...
  password: String
  password: String  // definition duplicated
}
```
We'll get the following exception:

Current:
`Exception in thread "main" graphql.AssertException: field password redefined`

Update:
`Exception in thread "main" graphql.AssertException: Duplicated definition for field 'password' in type 'User'`

### DIFF
--- a/src/main/java/graphql/schema/GraphQLObjectType.java
+++ b/src/main/java/graphql/schema/GraphQLObjectType.java
@@ -14,6 +14,7 @@ import java.util.stream.Collectors;
 
 import static graphql.Assert.assertNotNull;
 import static graphql.Assert.assertValidName;
+import static java.lang.String.format;
 
 
 @PublicApi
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/schema/GraphQLObjectType.java
```diff
@@ -160,7 +160,6 @@ public Builder field(UnaryOperator<GraphQLFieldDefinition.Builder> builderFuncti
          * from within
          *
          * @param builder an un-built/incomplete GraphQLFieldDefinition
-         *
```
andimarek: could you please not remove this extra line?

src/main/java/graphql/schema/GraphQLObjectType.java
```diff
@@ -145,7 +146,6 @@ public Builder field(GraphQLFieldDefinition fieldDefinition) {
          * </pre>
          *
          * @param builderFunction a supplier for the builder impl
-         *
```
andimarek: could you please not remove this extra line?

jdorleans:
No worries, I can add them back. I just followed the other files format which seems to be aligned with java standards better.

andimarek:
thanks

PR #811, opened 2017-10-31
TracingSupport: no ! in parent type names

pcarrier:
This breaks the Engine UI. Eg:

image


bbakerman:
I think we am happy to accept this but can you explain your comment

> This breaks the Engine UI. Eg:

What engine UI... Apollo?  Something else??  

Can we please have more background on this PR.

pcarrier:
@bbakerman yes this is a screenshot of https://www.apollographql.com/engine/

The background is that I work on the Engine team. The Engine backend, a GraphQL endpoint itself, is written with `graphql-java` and as of today, it runs behind `engineproxy`. As a result it is now instrumenting itself :).
This is the first time we have our visualizations for a GraphQL server and we find problems with the trace format implementation in `graphql-java` that might have gone unnoticed by its developers.

pcarrier:
Confirmed to fix the Apollo Engine UI:

image

PR #813, opened 2017-11-01
TracingSupport: startOffset is between query start and field start

pcarrier:


bbakerman:
The spec says differently 👍 

https://github.com/apollographql/apollo-tracing

> The startOffset of parsing, validation, or a resolver call is in nanoseconds, relative to the request start.

that is 'a resolver call is in nanoseconds, relative to the request start.'

So which is it?

I am happy to accept this this way as long as the spec gets updated as well.

pcarrier:
My code might be wrong. I believe my change makes `startOffset` refer to the start of the field's execution (`startFieldFetch`) as opposed to its end (`now`).

pcarrier:
Confirmed to fix the Apollo Engine UI:

image

PR #891, opened 2018-01-19
Platform agnostic field selection separator

nickbabcock:
I don't think it's a good idea to have the DataFetchingFieldSelectionSetImpl to list field names separated by a platform specific separator (considering these aren't files we're dealing with).

This fixes the build on windows machines (see test "field selection can be captured via data environment").

This is backwards incompatible change but IMO relatively minor and what looks like an oversight. If not, let me know the reasoning and feel free to close.


nickbabcock:
Looks like CI is failing for unrelated reasons

andimarek:
marked as breaking change ... correct @bbakerman ?

PR #937, opened 2018-02-21
getField method on DataFetchingEnvironment 

andimarek:
this simplifies getting the current Field getting queried.


andimarek:
There is one commit in this PR which doesn't belong here. I need to clean it up before we can merge.

andimarek:
@bbakerman please give it a quick review ... thanks

PR #945, opened 2018-02-26
rename TypeDefinitionRegistry.typeExtensions to objectTypeExtensions

andimarek:


andimarek:
it is a small breaking change

PR #946, opened 2018-02-27
Cached DataFetcherFactory and getSchema for DFFEnvironment

andimarek:
This adds a `CachedDataFetcherFactory` which makes sure that a `DataFetcherFactory` is only called once.

Also: the DataFetcherFactoryEnvironment gets a new `getSchema` method. Because the `DataFetcher` is created lazily the whole Schema is available, which is a significant advantage to normal `DataFetcher`.

It is a breaking change because `getDataFetcher` in `GraphQLFieldDefinition` needs now a `GraphQLSchema` argument.


andimarek:
@bbakerman maybe this should even be the default impl? I mean maybe an extra step after creating the schema with one single call to the `DataFetcherFactory`? This would make this thread safe lazy initialisation obsolete.

andimarek:
this will not merged: see #947 for the discussion.

PR #971, opened 2018-03-13
Coerce any type to id

tamland:
Since GraphQL doesn't care about the format of ID types, is there any reason to restrict it to these?


andimarek:
yes, we will go with the other approach, thanks for providing different options!

PR #1024, opened 2018-05-01
Fix Async.eachSequentiallyImpl is throwing NoSuchElement Exception

### LINKED ISSUE #992: Current Async.eachSequentiallyImpl throws lots of exceptions during non-async execution
Hey guys,

The current Async.eachSequentiallyImpl has a bug which causes an exception to be thrown pretty much every time it is called when running with the AsyncSerialExecutionStrategy. The exception thrown is NoSuchElementException. Thanks.

Basically the fix is the following:
---------------------------------
private static <T, U> void eachSequentiallyImpl(Iterator<T> iterator, CFFactory<T, U> cfFactory, int index, List<U> tmpResult, CompletableFuture<List<U>> overallResult) {
        if (!iterator.hasNext()) {
            overallResult.complete(tmpResult);
            return;  <------------- NEED TO ADD THIS RETURN HERE!
        }
        CompletableFuture<U> cf;
        try {
            cf = cfFactory.apply(iterator.next(), index, tmpResult);
            Assert.assertNotNull(cf, "cfFactory must return a non null value");
        } catch (Exception e) {
            cf = new CompletableFuture<>();
            cf.completeExceptionally(new CompletionException(e));
        }
        cf.whenComplete((cfResult, exception) -> {
            if (exception != null) {
                overallResult.completeExceptionally(exception);
                return;
            }
            tmpResult.add(cfResult);
            eachSequentiallyImpl(iterator, cfFactory, index + 1, tmpResult, overallResult);
        });
    }

david-tamjidi:
-Added return call to avoid un-necessarily throwing and catching the
    exception

fixes #992

### DIFF
--- a/src/main/java/graphql/execution/Async.java
+++ b/src/main/java/graphql/execution/Async.java
@@ -66,6 +66,7 @@ public class Async {
     private static <T, U> void eachSequentiallyImpl(Iterator<T> iterator, CFFactory<T, U> cfFactory, int index, List<U> tmpResult, CompletableFuture<List<U>> overallResult) {
         if (!iterator.hasNext()) {
             overallResult.complete(tmpResult);
+            return;
         }
         CompletableFuture<U> cf;
         try {


andimarek:
we should add a unit for for that

andimarek:
after looking it: no unit test is ok. the exception never surfaced because `CF.completeExceptionally` is silent when the `CF` is already completed.

PR #1056, opened 2018-05-21
Add specific UnknownOperationException when operation name is missing

tinnou:
## Description

When executing a GraphQL query, if multiple operations are defined in the query and the operation name is missing or there is no matching operation name contained in the GraphQL query, then GraphQL-java throws the general purpose `GraphQLException`. 

The problem with throwing `GraphQLException` is it's hard for the library consumer to catch for that specific use case and translate to a customer error. To do so we currently have to catch all `GraphQLException` and match on the exception message.

## Proposal

I added a new `UnknownOperationException` that is to be used specifically for this kind of errors, whether the operation name is missing or not matching the operation name provided in the query.
I also added it to the documentation for completeness.

## Testing 

I added unit tests in `NodeUtil` to cover the failure scenarios.

### DIFF
--- a/docs/exceptions.rst
+++ b/docs/exceptions.rst
@@ -42,6 +42,11 @@ These are not graphql errors in execution but rather totally unacceptable condit
  is thrown if the schema is not valid when built via
   graphql.schema.GraphQLSchema.Builder#build()`
 
+ -  `graphql.execution.UnknownOperationException`
+
+if multiple operations are defined in the query and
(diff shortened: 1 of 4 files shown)


alamothe:
Awesome! 👍 

One question: is there any situation where `GraphQLException` will be thrown after this fix? 

Is the expectation that `execute` should always return a result with errors as a part of it? Or do the clients still need to be on a lookout for `GraphQLException`?

PR #1066, opened 2018-06-04
add the git hash to the development build version

andimarek:


andimarek:
@tinnou thoughts on that change? thx

tinnou:
Just tested and prints `2018-06-03T20-03-43#9348430`. 

With the entire hash length it would be:
```2018-06-03T20-03-43#93484301d902a48eb6d5fbc8d19ba87287158766```

Looks good to me.

andimarek:
@tinnou yes it is the short version of the hash, which still should be enough to identify the commit.

PR #1074, opened 2018-06-08
Code improvements

abhijithtn:
- Using Objects.equals in equals method instead of manual comparison
 - Use map.computeIfAbsent instead of putIfAbsent to avoid additional
   lookup
 - Replace collection.size() > 0 with isEmpty calls
 - Used Map.entrySet instead of keySet to iterate

### DIFF
--- a/src/main/java/graphql/ExceptionWhileDataFetching.java
+++ b/src/main/java/graphql/ExceptionWhileDataFetching.java
@@ -1,16 +1,16 @@
 package graphql;
 
 
-import graphql.execution.ExecutionPath;
-import graphql.language.SourceLocation;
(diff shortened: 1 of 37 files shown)


andimarek:
I am sorry, but we can't accept this PR in this form.: it contains a lot of changes including formatting changes.

I would highly recommend that you reach out first before making bigger changes.

And if you create a PR:
Please provide small PRs and focus on one change at the time. 
Please provide a sufficient explanation for the change: just changing the style of from for loop to streams for example is not enough. 

Also for example switching from `putIfAbsent` to `computeIfAbsent` for performance is not enough of a reason: we need to see evidence that the current implementation is a problem.

Thanks!

PR #1097, opened 2018-06-23
make NodeTraverser public and improves it with preOder/postOrder methods

andimarek:
graphql-java should provide a good way to traverse a Document/Query. This makes the NodeTraverser public and adds special pre/post Order methods for easy usage.


andimarek:
This includes making some parts of the general Traverser public, but not everything.

andimarek:
@bbakerman @gkesler @exbe I would like to link to some good explanation about general traversal and for example pre/post order is a special case of depthFirst. Any recommendations? thx

PR #1112, opened 2018-07-09
refactor parser and making Nodes immutable 

andimarek:


andimarek:
breaking change because Nodes are created now by Builder. 
The internal Parsing changes are just internal

PR #1118, opened 2018-07-13
bugfix issue 1115

andimarek:
handle modified parent types (list and nonull) of fields correctly


andimarek:
this is merged into 9.x branch

PR #1117, opened 2018-07-12
Fix mistakes in defer.rst

RusticFlare:
Fixes some very simple errors in `defer.rst`

PR #1125, opened 2018-07-13
Fix improper quoting for post and comments

ehudon:
The words are not properly formatted because of improper quoting. Discovered while reading http://graphql-java.readthedocs.io/en/latest/defer.html


ehudon:
Duplicated of https://github.com/graphql-java/graphql-java/pull/1117. Sorry about that.

PR #1133, opened 2018-07-25
restrict add to SDLDefinitions

andimarek:
use the new SDLDefinitions Interface to make it more clear that only SDL Types are accepted.


andimarek:
this is a very small breaking change, but still breaking because we changed the signature of a public method

andimarek:
yea ... this is a pure compiler breaking change, so it is fine imho.

PR #1137, opened 2018-07-26
WIP - LocalDate and LocalDateTime scalars REF: Issue 991

rpmcdougall:
This isn't ready for merging yet purely from a testing point of view, but this is in reference to https://github.com/graphql-java/graphql-java/issues/991 
I am wondering if I am on the right track with this, or if there is more that is expected for functionality with regards to ISO standards to satisfy this issue.  I have some issues in the LocalDateTime scalar, but I would like some feedback/suggestions or any help anyone would like to give on this. I really enjoy using this library, so I figured I'd take a hand at trying to help out. Thanks!

### DIFF
--- a/src/main/java/graphql/Scalars.java
+++ b/src/main/java/graphql/Scalars.java
@@ -10,10 +10,16 @@ import graphql.schema.CoercingParseLiteralException;
 import graphql.schema.CoercingParseValueException;
 import graphql.schema.CoercingSerializeException;
 import graphql.schema.GraphQLScalarType;
+import graphql.util.DateTimeHelper;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
 import java.util.UUID;
+
+import java.time.LocalDateTime;
 
 import static graphql.Assert.assertShouldNeverHappen;
 
@@ -50,7 +56,7 @@ public class Scalars {
 
     /**
      * This represents the "Int" type as defined in the graphql specification : http://facebook.github.io/graphql/#sec-Int
-     *
+     * <p>
      * The Int scalar type represents a signed 32‐bit numeric non‐fractional value.
      */
     public static final GraphQLScalarType GraphQLInt = new GraphQLScalarType("Int", "Built-in Int", new Coercing<Integer, Integer>() {
@@ -116,7 +122,7 @@ public class Scalars {
 
     /**
      * This represents the "Float" type as defined in the graphql specification : http://facebook.github.io/graphql/#sec-Float
-     *
+     * <p>
      * Note: The Float type in GraphQL is equivalent to Double in Java. (double precision IEEE 754)
      */
     public static final GraphQLScalarType GraphQLFloat = new GraphQLScalarType("Float", "Built-in Float", new Coercing<Double, Double>() {
@@ -258,7 +264,7 @@ public class Scalars {
 
     /**
      * This represents the "ID" type as defined in the graphql specification : http://facebook.github.io/graphql/#sec-ID
-     *
+     * <p>
      * The ID scalar type represents a unique identifier, often used to re-fetch an object or as the key for a cache. The
      * ID type is serialized in the same way as a String; however, it is not intended to be human‐readable. While it is
      * often numeric, it should always serialize as a String.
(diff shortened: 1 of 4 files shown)


michaelschlies:
A title prefixed with WIP may help prevent accidental merges too!

bbakerman:
I am torn by this issue of datetime support in graphql-java

On the one hand its is NOT defined by the graphql spec and hence not a fundamental scalar

On the other hand people clearly want it (in graphql-java and graphql-js) because a date/datetime is a pretty fundamental scalar value.

I really dont know if we should accept a datetime scalar in the base implementation since there are a number of alternatives.

Thoughts @andimarek ?  @kaqqao ?  @tinnou ?

oexza:
I think a graphql-java-scalars project under the graphql-java organization
is proper.

andimarek:
It is a tough one tbh: I see the need and I think we as graphl-java org should provide it. But I am not sure about the place:

Inside the base library is the most convenient place, but you could argue that it should not really be in there and depending on the complexity and feature requests around I might agree. (For example joda-time support support is not possible inside the base library, because we will not add a joda-time dependency.)

The separate project like @oexza suggested makes sense, but has of course the extra overhead of creating and maintaining an extra project.

I don't have a clear favour atm and I would like to hear other opinions.

src/main/java/graphql/Scalars.java
```diff
@@ -709,4 +715,96 @@ public Character parseLiteral(Object input) {
...
+                );
+            }
+            return result;
+        }
+
+        @Override
+        public LocalDate parseLiteral(Object input) {
+            if (input instanceof StringValue && !((StringValue) input).isEqualTo(null)) {
+                return LocalDate.parse(((StringValue) input).getValue());
+            } else {
```
bbakerman: I would move ALL data parsing into the helper below

kaqqao:
This is what I can come up with:

1. Only supply readily usable examples in a separate repo, not actual deployed artifacts. This is similar to how [Gson project handles this](https://github.com/google/gson/tree/master/extras). The extras are maintained by the Gson team, but only as examples. You have to copy the code you need into your project.
2. Add it to graphql-java itself. I think this option only makes sense if it includes nothing but scalars for Java built-in types (like `java.util.time.*`). Adding more extras (apart from scalars) or library-specific scalars that drag in extra dependencies into the main project does not sound reasonable to me.
3. Have `extras` or `helpers` as a separate project (as already suggested). I wouldn't call it `scalars` as it immediately means any future expansion in scope would mean yet another separate project. This is annoying as it requires not only the maintenance of a separate project, but also a version compatibility matrix or a clever versioning scheme (e.g. does extras v1.2 work with graphql-java 10.0?)... But is appealing because it enables adding often-demanded features without bloating the core library. I'd go for this if the support for JodaTime and other libraries is seen as desirable, or if future expansions in scope are to be allowed for.

One interesting point that I encountered while solving this same problem in SPQR: how do you name the scalars that represent the same thing but in a different way? E.g. if `java.util.LocalDate` is a scalar called `LocalDate`, what is `org.joda.time.LocalDate` called? You can't call it `JodaLocalDate` as it reveals internal details of no use to the client. Can't call it `LocalDate` as it's potentially taken, even though it's unlikely both are in use at the same time.
I have yet to come up with a good solution for this... I've currently implemented the scalars for all `java.util.time.*`, `java.sql.*` and JodaTime types, and avoided naming collisions by ugly _name hacks_, e.g. `LocalDate` vs `DateLocal`. I'm not happy with this, and as a result haven't yet released the JodaTime module.
Also, I think I've opted for the separate module approach...

In any case, I'd happily contribute my implementations or help out with this differently...

kaqqao:
I've updated the previous comment substantially, so if you're reading this from an email, please the updated version on the web. Sorry for that...

rpmcdougall:
I suppose another alternative approach (which I am unsure of) is to be opinionated on which implementations of date/datetime etc. are implemented if the consensus is that it would be accepted to add additional dependencies(I didn't think about this immediately when I started working on this PR). For example, only supporting `java.util.Date` or the joda implementation. I'm not sure this would be a sane approach, but just a thought. I think this more or less gives more credit to the examples into helpers approach in my opinion since there are so many different date/time dependencies one could use as already has been stated.

andimarek:
@rpmcdougall we will not add additional dependencies to graphql-java itself: this is a fundamental design decision of the library itself in order to keep our maintenance effort low and most importantly to avoid transitive dependency problems.

rpmcdougall:
Would we maybe want to come to a consensus on what a good example module might look like here and then close this PR and open a feature request to discuss what all might make good candidates for an examples project?

rpmcdougall:
Going to close this since there hasn't been much activity on it. Thanks for the earlier discussion everyone.

andimarek:
I am sorry that we didn't get back in time, but we are quite constrained regarding the time we can invest in this project.

In general I would like to see a proper "official" graphql-java date scalar support (and probably others) but this topic is quite complex and to be honest we don't have the time at the moment to tackle that.

Thanks for your understanding and for the PR in the first place!

rpmcdougall:
I totally understand, I'll keep an eye out for any other issues I can attempt to tackle!

PR #1207, opened 2018-09-10
DataFetcher declared to throw Exception

kaqqao:
No longer forces the client code to wrap checked exceptions

### DIFF
--- a/src/main/java/graphql/schema/AsyncDataFetcher.java
+++ b/src/main/java/graphql/schema/AsyncDataFetcher.java
@@ -64,7 +64,13 @@ public class AsyncDataFetcher<T> implements DataFetcher<CompletableFuture<T>> {
 
     @Override
     public CompletableFuture<T> get(DataFetchingEnvironment environment) {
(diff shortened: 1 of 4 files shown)


src/main/java/graphql/schema/AsyncDataFetcher.java
```diff
@@ -64,7 +64,13 @@ public AsyncDataFetcher(DataFetcher<T> wrappedDataFetcher, Executor executor) {
 
     @Override
     public CompletableFuture<T> get(DataFetchingEnvironment environment) {
-        return CompletableFuture.supplyAsync(() -> wrappedDataFetcher.get(environment), executor);
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                return wrappedDataFetcher.get(environment);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
```
kaqqao: Should I maybe use a specific (internal) kind of exception here that can be unpacked by the error handler later?
Or perhaps sneaky-throw the original?

PR #1240, opened 2018-09-22
GraphQLSchema.Builder should expose Queries, Mutations and Subscriptions.

dnebing:
This is a PR for https://github.com/graphql-java/graphql-java/issues/1239


andimarek:
Thank you for your PR, but we will not that merge that: our Builders follow a common pattern and making some fields public is something in general we don't do.

PR #1244, opened 2018-09-24
Fix SchemaPrinter generating invalid values for directives

juhovh:
Steps to reproduce:

Add a `@deprecated` directive to a field or a value and then run it through both `SchemaParser` and `SchemaPrinter`, it will be transformed into `@deprecated(reason : "No longer supported" = "No longer supported")` which is invalid GraphQL according to the [spec](http://facebook.github.io/graphql/June2018/#sec-Language.Directives).


andimarek:
thanks @juhovh

PR #1249, opened 2018-09-27
Fix test failing due to date formatting problems (#1247)

felipe-gdr:
Test could potentially fail if Locale was different than en.


andimarek:
thanks @felipe-gdr

PR #1252, opened 2018-10-01
Improving complexity calculation

pete-proton:
no need for list of integers for storing complexity - integer as a sum is good enough

### DIFF
--- a/src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
+++ b/src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
@@ -7,13 +7,13 @@ import graphql.execution.instrumentation.SimpleInstrumentation;
 import graphql.execution.instrumentation.parameters.InstrumentationValidationParameters;
 import graphql.validation.ValidationError;
 
-import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
 import static graphql.Assert.assertNotNull;
 import static graphql.execution.instrumentation.SimpleInstrumentationContext.whenCompleted;
+import static java.util.Optional.ofNullable;
 
 /**
  * Prevents execution if the query complexity is greater than the specified maxComplexity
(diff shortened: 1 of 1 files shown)


src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
```diff
@@ -54,20 +54,21 @@ public MaxQueryComplexityInstrumentation(int maxComplexity, FieldComplexityCalcu
...
-                    if (valuesByParent.containsKey(env)) {
-                        childsComplexity = valuesByParent.get(env).stream().mapToInt(Integer::intValue).sum();
-                    }
+                    int childsComplexity = valuesByParent.getOrDefault(env, 0);
                     int value = calculateComplexity(env, childsComplexity);
-                    valuesByParent.putIfAbsent(env.getParentEnvironment(), new ArrayList<>());
-                    valuesByParent.get(env.getParentEnvironment()).add(value);
+
+                    valuesByParent.compute(env.getParentEnvironment(), (key, oldValue) ->
+                        ofNullable(oldValue).orElse(0) + value
```
bbakerman: This sets it to `oldValue + value` and not just `value` like it did before

pete-proton:
raised this small one 2 weeks ago - just one reply, which seemed to me like the PR was not looked at thoroughly. could you tell me if there is anything needs to be done from you PoV? 

TBH feel a little discouraged raising further PR's

andimarek:
@pete-proton I am sorry and I can understand your reaction, but this a private run project with limited time we can spend on it. This means unfortunately it can take quite some time until we get back to PRs, especially not urgent PRs.

We really appreciate your contribution and I hope we see more in the future.

Thanks,
Andi

src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
```diff
@@ -88,11 +89,11 @@ protected AbortExecutionException mkAbortException(int totalComplexity, int maxC
 
     QueryTraversal newQueryTraversal(InstrumentationValidationParameters parameters) {
         return QueryTraversal.newQueryTraversal()
-                .schema(parameters.getSchema())
-                .document(parameters.getDocument())
-                .operationName(parameters.getOperation())
-                .variables(parameters.getVariables())
-                .build();
+                             .schema(parameters.getSchema())
```
andimarek: can you make sure you are using the https://github.com/graphql-java/graphql-java/blob/master/graphql-java-code-style.xml codestyle? thanks

pete-proton:
thank you! do I need to reformat still? sorry for the late reply

bbakerman:
If you could then that would be great

pete-proton:
sure, sorry took a while, but finally created https://github.com/graphql-java/graphql-java/pull/1323

PR #1283, opened 2018-10-23
Pass wrappedType name through in GraphQLNonNull

timbonicus:
I was surprised to see GraphQLNonNull always returning null for getName(); it seems like this should pass through the name of the wrapped type.


andimarek:
My first reaction is that it is actually a mistake that there is getName(). 
I am not sure about the consequences if we change it, but If we do it, we should not just the wrapped type name, but add `[..]` or `!`

timbonicus:
Thanks @bbakerman, your explanation makes a lot of sense. I was upgrading from graphql-java v9 to v11 and this behavior changed. I'll switch to using GraphTypeUtil.unwrapAll instead!

PR #1286, opened 2018-10-25
Copy fieldDefinition when builder tries to create new DataFetchingEnv…

johnliucn925:
Copy fieldDefinition when builder tries to create new DataFetchingEnv from existing one.


andimarek:
@bbakerman we should backport that to 9.x I think. What do u think?

bbakerman:
Yes I think so

PR #1292, opened 2018-10-31
Immutable AbstractNode comments

michalbogdal:
Instead of creating new instance of ArrayList every time when getComments method is called we can return just immutable list


andimarek:
closing it now because of no feedback: please reopen it if needed....thx

PR #1327, opened 2018-11-25
Handle array values in SchemaPrinter

### LINKED ISSUE #1326: SchemaPrinter handles array values incorrectly
While printing `GraphQLList`, the values are are always expected to be `Iterable` instances, but arrays are also legal values in this case. As a result, if an array is provided, it will simply be `toString`ed, producing an invalid print-out.

kaqqao:
Make SchemaPrinter able to deal with array values for `GraphQLList`

Fixes #1326


kaqqao:
Didn't add any tests as it's kind of a trivial thing, but will do if you want.

Issue #1353, opened 2018-12-17
Allow instrumentation of the execution input once the parsed query is available


xtaixe:
Our use case is that we need to convert the values of some arguments and variables. To know which variables to convert we need the parsed query to know in which arguments they are used to check if they are eligible or not, so we can't use the current execution input instrumentation.

As discussed before, I looked into using our own scalar type to coerce the values for this, but I don't think that covers our case because:
1. We can't change the actual types, which should be IDs, due to compatibility with other systems and clients
2. Even if we replace the ID GraphQLScalarType with our own implementation, it can't get the necessary info about which values to convert and which not

I'll create a PR with a new instrumentation. Let me know if you think it's not a good idea or there are other alternatives

andimarek:
closing it for now: we provided a possibility to change the variables too: https://github.com/graphql-java/graphql-java/pull/1355

xtaixe:
Hi @andimarek 

As discussed with @bbakerman that PR doesn't work for all our cases. If the Document is cached or provided by the `preparsedDocumentProvider` the instrumentation won’t be called, but we still need to convert the variables in that case…  As @bbakerman pointed out thought, that instrumentation can't be done after the document caching because if you rewrote the document you would need to validate again.  So caching and document rewriting is opposed (while caching and variable rewriting is not).

So if we still needed to go the instrumentation route, we would still need an instrumentation just for the variables/execution input as I had in my PR (#1354)

That said, I also discussed with @bbakerman that the work in #1336 might be enough for us to do what we need with a bit of extra work on our side to get the info we need for the conversions, so we'll see how things go once we do the upgrade to version 12 (which presents some challenges due to the immutability changes I think)

PR #1354, opened 2018-12-17
Allow instrumentation of the execution input once the parsed query is available

xtaixe:
PR for #1353

The test shows how you could also use this to make queries more cacheable


andimarek:
@xtaixe thanks for the PR. We decided to go with a slightly different approach: https://github.com/graphql-java/graphql-java/pull/1355

PR #1385, opened 2019-01-11
Add ability to transform queries using QueryTraversal with full type information

tsroka:
@andimarek This is proof of concept, needs some additional test + most likely your feedback if this is good direction. 
While working on Nadel I found out that in order to use AstTransformation with full type info I would need to copy the code from QueryTravesal, so instead I changed graphql-java in a such way that it allows the transformation. 
Let me know if this is a good direction, or what the alternative approach should be. I tried to minimize code changes and retain full backward compatibility, but maybe to make code nicer we can just change QueryVisitor class to return TraversalControl (instead of introducing new interface QueryVisitorWithControl) 
Code is lacking more detailed unit tests which I will add once we decide if this implementation is way to go.
The main change here is to allow QueryVisitor to decide about traversal control (needed if node gets removed, since we do not want to visit children in that case) and also do not skip visiting if directives  says so (it will be up to the implementation of the visitor).


src/main/java/graphql/analysis/QueryTraversal.java
```diff
@@ -325,7 +381,6 @@ public static Builder newQueryTraversal() {
          * The schema used to identify the types of the query.
          *
          * @param schema the schema to use
-         *
```
andimarek: please use graphql-java code style

src/main/java/graphql/analysis/QueryTraversal.java
```diff
@@ -256,7 +310,9 @@ public TraversalControl visitFragmentSpread(FragmentSpread fragmentSpread, Trave
             QueryTraversalContext parentEnv = context.getVarFromParents(QueryTraversalContext.class);
 
             GraphQLCompositeType typeCondition = (GraphQLCompositeType) schema.getType(fragmentDefinition.getTypeCondition().getName());
-
+            Assert.assertTrue(typeCondition != null, "Invalid type condition '%s' in fragment '%s'", fragmentDefinition.getTypeCondition().getName(),
```
andimarek: we have `Assert.assertNonNull`

andimarek: but this assert is not really anyway, because we assume the query is a valid one, right?

src/main/java/graphql/util/DefaultTraverserContext.java
```diff
@@ -57,10 +57,7 @@ public DefaultTraverserContext(T curNode,
 
     @Override
     public T thisNode() {
-        assertFalse(this.nodeDeleted, "node is deleted");
```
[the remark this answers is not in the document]
tsroka: wow, that wasn't intended to go into the PR. good catch

src/main/java/graphql/analysis/QueryTraversal.java
tsroka: I am not sure. If you put wrong query (e.g. with fragment that is not suppose to be there) you will get weird NPE that will not tell you what's going on.

PR #1422, opened 2019-02-14
Ability to provide CacheControl instance via ExecutionInput

kaqqao:
Make CacheControl explicitly accessible from DataFetchingEnvironment, similar to how DataLoaderRegistry currently works


andimarek:
@kaqqao thanks for that.

PR #1445, opened 2019-03-05
Add support for visiting fragment definitions in QueryVisitor.

tsroka:


src/main/java/graphql/analysis/QueryVisitor.java
```diff
@@ -16,4 +16,8 @@
 
     void visitFragmentSpread(QueryVisitorFragmentSpreadEnvironment queryVisitorFragmentSpreadEnvironment);
 
+    default void visitFragment(QueryVisitorFragmentEnvironment queryVisitorFragmentEnvironment) {
```
andimarek: I think we should call it `visitFragmentDefinition`. `visitFragment` is not clear enough.

src/main/java/graphql/analysis/QueryVisitorFragmentEnvironment.java
```diff
@@ -0,0 +1,13 @@
+package graphql.analysis;
+
+import graphql.PublicApi;
+import graphql.language.FragmentDefinition;
+import graphql.language.Node;
+import graphql.util.TraverserContext;
+
+@PublicApi
+public interface QueryVisitorFragmentEnvironment {
```
andimarek: same as above: `QueryVisitorFragmentDefitintionEnvironment` is more accurate and more future proof.

src/main/java/graphql/analysis/QueryVisitorFragmentEnvironmentImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+import graphql.Internal;
+import graphql.language.FragmentDefinition;
+import graphql.language.Node;
+import graphql.util.TraverserContext;
+
+import java.util.Objects;
+
+@Internal
+public class QueryVisitorFragmentEnvironmentImpl implements QueryVisitorFragmentEnvironment {
```
andimarek: a toString would be nice, but is not strictly needed.

tsroka:
> nice ... see my comments about naming

Renamed as per your comments and added #toString

src/test/groovy/graphql/analysis/QueryTraversalTest.groovy
```diff
@@ -20,7 +21,7 @@ import static graphql.schema.GraphQLList.list
 import static graphql.schema.GraphQLNonNull.nonNull
 import static java.util.Collections.emptyMap
 
-class QueryTraversalTest extends Specification {
+class QueryTrqaversalTest extends Specification {
```
andimarek: typo :-)

Issue #1448, opened 2019-03-07
Getting error for valid Double value


josh00007777:
**Describe the bug**
For below schema --

type Employee {
    sal: Float   
}
type Query {
   getEmployee(sal:Float): Employee
}

When this query is passed, Invalid Syntax error is raised --

{
  getEmployee (sal:1.7976931348155E+308){
    sal
  }
}

If I remove + from "sal" value then it is working, as per java 1.7976931348155E+308 is valid double value.

andimarek:
merged the PR ... thanks

PR #1452, opened 2019-03-09
Exponents can have positive exponents

jadedevin13:
For #1448


andimarek:
@jadedevin13 thanks for that PR. I agree that we are not spec conform in this matter. Can you please add a unit test for that? That would be great, thanks.

andimarek:
never mind: @jadedevin13 I added the test myself

PR #1525, opened 2019-05-06
fixes #1524

### LINKED ISSUE #1524: ExecutionContext.addError can be very slow O(n^2)
addError takes O(n^2) time. This becomes extremely slow with a large volume of errors.

wasche:
before:
AddError.benchMarkAddError    ss    5  1.079 ± 0.064   s/op

after:
AddError.benchMarkAddError    ss    5  0.131 ± 0.059   s/op

### DIFF
--- a/src/main/java/graphql/execution/ExecutionContext.java
+++ b/src/main/java/graphql/execution/ExecutionContext.java
@@ -15,8 +15,10 @@ import graphql.schema.GraphQLSchema;
 import org.dataloader.DataLoaderRegistry;
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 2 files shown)


bbakerman:
Do you have a use case where you want to create 5000 errors and you have noticed an actual slow down because of the O(n) error lookup

These types of micro benchmarks have a history of not paying off in practice.

wasche:
The amount of throuput we do, we've run into this issue numerous times, and it effects us hitting our SLA. So yes, this has had a significant impact on performance. In an ideal world, there wouldn't be so many errors, but they are out of our control.

samkline:
Hi @bbakerman, thanks for looking at this. I was working with @wasche on this and at the time I had hooked up VisualVM to get some CPU profiles. The way I ran this test was to throw a bunch of queries at our GraphQL service. Each query fetched a few hundred items from one service, and then went to a few supplementary services to fill in the fields for those items. I took one of those supplementary services offline to see how the system responded.

Sadly I don't have the full results anymore but I do have this screenshot (below) showing that within my test, of all the time used by `ExecutionStrategy.completeField`, 93% of it was spent in `ExecutionContext.addError`. I had added a workaround in our service when we were running 9.2, but now that we're upgrading to 12.0, our workaround no longer works (we were relying on the public constructor of `ExecutionContext` to subclass it and override the `addError` behavior).

Screen Shot 2018-12-05 at 7 22 36 AM

bbakerman:
Ok thanks for the evidence... We will merge

PR #1543, opened 2019-05-18
Ability to merge GraphQLCodeRegistries

### LINKED ISSUE #1542: Ability to merge GraphQLCodeRegistries
I have a use case where types from multiple existing schemas are re-used in building a new schema. In v11 this was as easy as calling `GraphQLSchema.Builder#additionalTypes`.
As of v12 this is no longer enough as all the logic has now been moved from the fields and types themselves into the `GraphQLCodeRegistry`, meaning I need a way to copy the data fetchers and type resolvers from one code registry to another (in addition to copying the types over).

The existing `GraphQLCodeRegistry.builder(GraphQLCodeRegistry)` isn't enough as it only allows copying from one source, not multiple.

While it _is_ possible to work around this by recursing through the source schemas and copying the fechers one by one for each field on a copied type, allowing simple copying of fetchers directly between the registries is infinitely easier.

kaqqao:
Closes #1542

### DIFF
--- a/src/main/java/graphql/schema/GraphQLCodeRegistry.java
+++ b/src/main/java/graphql/schema/GraphQLCodeRegistry.java
@@ -315,6 +315,11 @@ public class GraphQLCodeRegistry {
             return this;
         }
 
+        public Builder dataFetchers(GraphQLCodeRegistry codeRegistry) {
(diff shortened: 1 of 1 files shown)


kaqqao:
Argh, sorry, totally forgot this one 😖
Thanks for taking care of it!

PR #1560, opened 2019-06-01
Retrieve DataFetcher from CodeRegistry using FieldCoordinates

dkuc84:
Currently GraphQLCodeRegistry allows registering DataFetchers using either FieldCoordinates or by explicitly providing parent and target field (which is then used to construct field coordinates) but it only allows retrieving DataFetcher for a given field based on the parent container field. This is problematic when attempting to configure directives while recursively generating the schemas (i.e. how it is done in graphql-kotlin) as the parent object is not yet available.

It appears that parent GraphQLFieldsContainer is only used to construct target field coordinates for retrieving the data fetcher. It would be great if we could retrieve the data fetchers by the coordinates as well as it is blocking us (graphql-kotlin) from upgrading to latest version of graphql-java.


bbakerman:
Thanks for this

dkuc84:
@bbakerman Thanks for the merge. By any chance do you know whether there is some high level ETA for the next release?

bbakerman:
We hope to release in the next two weeks

Issue #1579, opened 2019-06-24
Make internal SchemaGeneratorDirectiveHelper public


oliemansm:
As discussed on Spectrum will create a PR making SchemaGeneratorDirectiveHelper public instead of package private. We're using that class in [graphql-java-tools](https://github.com/graphql-java-kickstart/graphql-java-tools) to build the directives when generating the schema. Right now we can use it by having a helper class of our own in a package with the same name, but that will give problems when used in modular system, see also https://github.com/graphql-java-kickstart/graphql-java-tools/issues/282.

We do realize this is an internal class subject to change.

PR #1580, opened 2019-06-24
Make SchemaGeneratorDirectiveHelper public

oliemansm:


oliemansm:
As discussed on Spectrum and #1579

Issue #1581, opened 2019-06-26
Support additional Optional like types from other packages ( guava, vavr, etc )


sean-brandt:
**Describe the bug**
Recent changes moved unboxing of possible optionals to a package scoped class and static method with no 'hook' for extension or overriding.

**To Reproduce**
Attempt to support another optional type other than java's Optional

It's possible to support extension without needing to use either reflection or adding third party dependencies. PR to address this is inbound.

andimarek:
closing, because we now support custom unboxers.

PR #1582, opened 2019-06-26
Support custom Optional unboxers

sean-brandt:
Add support via the java ServiceLoader mechansim for providing custom
implementations and allowing for support of Guava, Vavr, etc.

No support for extension of the current mechanism, or overriding it

Addresses: #1581

### DIFF
--- a/src/main/java/graphql/execution/UnboxPossibleOptional.java
+++ b/src/main/java/graphql/execution/UnboxPossibleOptional.java
@@ -1,46 +1,102 @@
 package graphql.execution;
 
-import graphql.Internal;
-
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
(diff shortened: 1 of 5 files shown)


sean-brandt:
Thanks. I needed support for vavr's Option but knew/had seen the design goal of minimum dependencies. ServiceLoader is often overlooked but seemed to fit the bill here. :)

sean-brandt:
Just checking - this is planned for inclusion in the next release? Are there outstanding blockers preventing a merge at this point?

andimarek:
Hi @sean-brandt ... after giving it some thought we decided to create a similar but not quite the same mechanism #1684. Could you please review the other PR and let us know what you think?

The main reason why we are not completely happy with this PR is the ServiceLoader approach: on its own it is perfectly valid but we don't want to add another way of configuring/extending graphql java in order to keep the API surface small and consistent. 

Thanks a lot

Issue #1588, opened 2019-07-03
Validation of non-null input object fields ignores default values


craigkost:
**Describe the bug**

Request validation treats non-null input object fields with default values as missing. _Note:_ default values are handles correctly for query arguments.

**To Reproduce**

Given the following GraphQL schema (I apologize for its contrived nature)

```
input Locale {
    country: String! = "CA"
    language: String! = "en"
}

type Query {
    name (
        id: ID!
        locale: Locale! = {
            country: "US"
            language: "en"
        }
    ) : String
}
```

The expected behaviour for the query below is to pass validation and have locale set to `{country: "CA" language: "fr"}`; however, it fails stating country has not been provided despite it having a default value of "CA".

```
{
  name (id: "123" locale: {language: "fr"})
}
```

The follow, although unusual, should also succeed with local as `{country: "CA" language: "en"}`. It currently fails stating `country` and `language` are missing.

```
{
  name (id: "123" locale: {})
}
```

Related to https://github.com/facebook/graphql/pull/418

> NEW: Non-null arguments and **input object fields** are no longer required to be provided a value/variable if that argument also supplies a default value.

andimarek:
closing because the fix is merged

PR #1589, opened 2019-07-03
Non-null input object fields with default values should be valid

craigkost:
Update query validation to take default value for input object fields into account.

This should not break existing functionality but instead allow additional requests to pass validation. It should improve compliance with https://github.com/facebook/graphql/pull/418

Addresses: https://github.com/graphql-java/graphql-java/issues/1588


src/test/groovy/graphql/IssueNonNullDefaultAttribute.groovy
```diff
@@ -11,16 +11,28 @@ import static graphql.schema.idl.TypeRuntimeWiring.newTypeWiring
 // See https://github.com/facebook/graphql/pull/418
 class IssueNonNullDefaultAttribute extends Specification {
     def spec = '''
+            input Locale {
```
craigkost: Not sure if it is acceptable to extend this test or if you would prefer a new test class added.

PR #1594, opened 2019-07-09
add dependency graph

andimarek:
@gkesler @exbe here is my early PR for adding the dependency graph. Not really ready yet.

Inside `createDependencyGraph` it creates dot file representation of the graph which you can try out for example here:  https://dreampuf.github.io/GraphvizOnline/


src/main/java/graphql/execution/nextgen/depgraph/DependencyGraph.java
```diff
@@ -0,0 +1,123 @@
...
+        traverser.traverse(roots, new TraverserVisitorStub<MergedFieldWithType>() {
+            @Override
+            public TraversalControl enter(TraverserContext<MergedFieldWithType> context) {
+                MergedFieldWithType mergedFieldWithType = context.thisNode();
+                System.out.println(mergedFieldWithType.getName() + "" + mergedFieldWithType.getObjectType().getName());
+
+                FieldVertex fieldVertex = createFieldVertex(mergedFieldWithType, graphQLSchema);
+                FieldVertex parentVertex = context.getVarFromParents(FieldVertex.class);
+                fieldVertex.addDependency(parentVertex);
+                parentVertex.addDependsOnMe(fieldVertex);
```
gkesler: Perhaps, it would be easier for the client if fieldVertex.addDependency had recorded dependencies both ways. Conceptually "dependency" represents an edge between two vertices, so in my mind it would be more convenient if all necessary bookkeeping was done in one operation.

src/main/java/graphql/execution/nextgen/depgraph/DependencyGraph.java
```diff
@@ -0,0 +1,123 @@
...
+                context.setVar(FieldVertex.class, fieldVertex);
+
+                return TraversalControl.CONTINUE;
+            }
+
+        });
+
+        List<Set<FieldVertex>> listOfClosures = new ArrayList<>();
+        Set<FieldVertex> allClosed = new LinkedHashSet<>();
+        allClosed.add(rootVertex);
```
gkesler: looks redundant.
`allClosed.add(rootVertex)` will be anyways called on L.66 at the very first iteration of the while loop below.

src/main/java/graphql/execution/nextgen/depgraph/DependencyGraph.java
```diff
@@ -0,0 +1,123 @@
...
+        Set<FieldVertex> curSourceSet = new LinkedHashSet<>();
+        curSourceSet.add(rootVertex);
+
+        while (!curSourceSet.isEmpty()) {
```
gkesler: @andimarek 
This is a topological sorting algorithm.
Currently it is used to perform static analysis only, but it actually needs to be used to do the field resolution.
The main difference between these 2 cases is that during query execution the actual runtime data should be used to evaluate vertex predicate and if that fails, that vertex **and all its dependents** should be excluded from the topological sequence so they won't make their entries in the resulting object.

src/main/java/graphql/execution/nextgen/depgraph/DependencyGraph.java
```diff
@@ -0,0 +1,123 @@
...
+
+
+        Set<FieldVertex> allVertices = new LinkedHashSet<>();
+
+        Traverser<MergedFieldWithType> traverser = Traverser.depthFirst(getChildren);
+        FieldVertex rootVertex = new FieldVertex(null, null, null, null, null);
+        traverser.rootVar(FieldVertex.class, rootVertex);
+        allVertices.add(rootVertex);
+        List<MergedFieldWithType> roots = fieldCollector.collectFromOperation(parameters, operationDefinition, graphQLSchema.getQueryType());
+        traverser.traverse(roots, new TraverserVisitorStub<MergedFieldWithType>() {
```
gkesler: I think recursive collecting fields (graph traversal!) and then traversing collected fields again to build a graph is quite redundant. I would suggest to build the graph in 1 traverse cycle, incrementally with field collection.

andimarek:
outdated. We have now a `ExecutableNormalizedOperation` structure which can serve as basis for an execution plan

PR #1596, opened 2019-07-10
make sure variables map and query can't be null

andimarek:


grrttedwards:
Just out of curiosity, since I'm the author of that issue -- is this PR supposed to actually initialize that map? Or was that done somewhere else? You mentioned the "code path was not valid", so I assume the latter.

andimarek:
@grrttedwards sorry if it was not clear: we have always initiated the map with an empty map, but you were able to set it to `null`. The new PR makes sure it is never set to `null`.
With "code path not valid" I mean I was not able to reproduce the exact Exception because we changed things but I got another Exception. 
Does this help?

grrttedwards:
@andimarek Wow, I completely missed the root cause of my issue — that I was actually _setting_ it to null! Duh!

Thanks for the reply. Keep up the good work.

Issue #1504, opened 2019-04-16
SchemaDiff has no special/helpful handling for directives like @deprecated


grrttedwards:
Version `11.0`
Currently, SchemaDiff does not make any distinctions for changes which involve `@deprecated` directives. This may be intended, but I was considering a use case where this is important information for an automated service registry. For example, the following change generates no meaningful events:

```diff
type Query { foo: TypeA } 
-type TypeA { abc: String xyz: Int }
---
type Query { foo: TypeA } 
+type TypeA { abc: String @deprecated xyz: Int }
```
Output:
```
14:22:26.799 [main] INFO - Infos count: 9
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='Examining operation 'Query' ...', level=INFO, category=null, typeName='Query', typeKind=Operation, fieldName=null}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='Examining type 'Query' ...', level=INFO, category=null, typeName='Query', typeKind=Object, fieldName=null}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='	Examining field 'Query.foo' ...', level=INFO, category=null, typeName='Query', typeKind=Object, fieldName=foo}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='Examining type 'TypeA' ...', level=INFO, category=null, typeName='TypeA', typeKind=Object, fieldName=null}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='	Examining field 'TypeA.abc' ...', level=INFO, category=null, typeName='TypeA', typeKind=Object, fieldName=abc}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='	Examining field 'TypeA.xyz' ...', level=INFO, category=null, typeName='TypeA', typeKind=Object, fieldName=xyz}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='	Examining field 'TypeA.abc' ...', level=INFO, category=null, typeName='TypeA', typeKind=Object, fieldName=abc}
14:22:26.807 [main] INFO - DifferenceEvent{ reasonMsg='	Examining field 'TypeA.xyz' ...', level=INFO, category=null, typeName='TypeA', typeKind=Object, fieldName=xyz}
[...]

PR #1610, opened 2019-07-30
Get more information about @deprecated fields from SchemaDiff

grrttedwards:
Addresses #1504

Let me know your thoughts on this. My strategy was to just add another category, which can provide the extra information and keep the `Danger` and `Breaking` levels untouched.

The two new combinations of level and category are:
1. level: `DANGER`, category: `DEPRECATED` for fields that become deprecated
1. level: `BREAKING`, category: `DEPRECATED` for fields that were deprecated and now missing

Todo:
- [x] Detect fields that are `@deprecated` in the new schema
- [x] Detect removals that were `@deprecated` in the old schema
- [x] ???

Remarks:
It's actually interesting to note that `@deprecated` on anything but `FIELD_DEFINITION | ENUM_VALUE` is not currently a valid spec-compliant use of the directive.
See: https://graphql.github.io/graphql-spec/June2018/#sec--deprecated
and the corresponding implementation proposed on graphql-js: https://github.com/graphql/graphql-js/pull/1560/files

### DIFF
--- a/src/main/java/graphql/schema/diff/DiffCategory.java
+++ b/src/main/java/graphql/schema/diff/DiffCategory.java
@@ -26,5 +26,9 @@ public enum DiffCategory {
     /**
      * The new API has changed something compared to the old API
      */
-    DIFFERENT
+    DIFFERENT,
+    /**
+     * The new API has deprecated something or removed something deprecated from the old API
+     */
+    DEPRECATED
 }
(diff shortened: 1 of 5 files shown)


grrttedwards:
@andimarek @bbakerman Sorry for the bump, my organization has been looking for this feature for a while. Could you spend a moment taking a look at this?

PR #1670, opened 2019-10-07
Fixed build break when the project directory path contains spaces

dennisdietrich:


dennisdietrich:
Looks like graphql.util.TreeParallelTraverserTestgraphql.util.TreeParallelTraverserTest failed which I don't think has anything to do with my change. Can someone more familiar with the codebase please confirm?

PR #1684, opened 2019-10-25
add hook to allow for custom value unboxing

andimarek:
This PR adds a general ValuUnboxer hook which lets you implement a custom logic for unboxing values. The default impl unboxes only java.lang.Optional.

### DIFF
--- a/src/main/java/graphql/GraphQL.java
+++ b/src/main/java/graphql/GraphQL.java
@@ -7,6 +7,7 @@ import graphql.execution.Execution;
 import graphql.execution.ExecutionId;
 import graphql.execution.ExecutionIdProvider;
 import graphql.execution.ExecutionStrategy;
+import graphql.execution.PossibleOptionalUnboxer;
 import graphql.execution.SubscriptionExecutionStrategy;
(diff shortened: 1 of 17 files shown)


sean-brandt:
While a bit more intrusive a change than #1582 ( even considering that it is my PR ) obviously this works as well. 

I'm not personally convinced that the ServiceLoader approach is too much cognitive dissonance, but don't have any concerns that this will address the need I was trying to fill with #1582

andimarek:
@sean-brandt  you are right: it is a more intrusive change, but we are happy to pay this price for more consistency for our users.

kaqqao:
Just a thought, but why not go for a more general solution for converting/adapting values and then implement unboxing optionals in terms of that? Where I come from (generating GraphQL schemas dynamically), such a solution would be widely applicable. I have a similar custom solution in place already.
As it stands, the current approach solves a highly specialized problem only at the expense of a rather intrusive change...

andimarek:
this hook actually allows you to "unbox" any value. So it that sense it is a general solution. Will try to make the naming more clear

PR #1694, opened 2019-11-07
Correctly support additional optional arguments in fields defined by interfaces

### LINKED ISSUE #1595: Incorrect type validation for additional optional arguments
**Describe the bug**

The schema type checker checks if the argument sizes are equal: https://github.com/graphql-java/graphql-java/blob/828bf5c8d18815859a4aaa525139d99b12ffe3c4/src/main/java/graphql/schema/idl/SchemaTypeChecker.java#L503 

The [spec](https://github.com/graphql/graphql-spec/blob/master/spec/Section%203%20--%20Type%20System.md), however, says the following:

```
The object field may include additional arguments not defined in the interface field, but any additional argument must not be required, e.g. must not be of a non-nullable type.
```

This results in `does not have the same number of arguments as specified via interface`.

**Example**
```
interface a {
   field(a: string = 'hello'): string
}

type b implements a {
   field(a: string = 'hello', b: string = 'world'): string
}
```

danez:
This ensures that `MissingInterfaceFieldArgumentsError` is only thrown if the object type has less arguments than the interface type.

It also adds a new check to see if every additional argument in the object type is optional.

Ref: The [spec](https://graphql.github.io/graphql-spec/June2018/#sec-Objects) says:
```
The object field may include additional arguments not defined in the interface field, but any additional argument must not be required, e.g. must not be of a non‐nullable type.
```

Fixes #1595

### DIFF
--- a/src/main/java/graphql/schema/idl/SchemaTypeChecker.java
+++ b/src/main/java/graphql/schema/idl/SchemaTypeChecker.java
@@ -14,6 +14,7 @@ import graphql.language.InputObjectTypeDefinition;
 import graphql.language.InputValueDefinition;
 import graphql.language.InterfaceTypeDefinition;
 import graphql.language.Node;
+import graphql.language.NonNullType;
 import graphql.language.ObjectTypeDefinition;
 import graphql.language.ObjectTypeExtensionDefinition;
 import graphql.language.OperationTypeDefinition;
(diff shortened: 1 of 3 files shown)


danez:
Any news?

andimarek:
@danez sorry for the delay. It looks good overall. We will schedule it for the next release, which should happen soon.

PR #1711, opened 2019-12-09
Add transform-like methods to GraphQLSchema.Builder and GraphQLObjectType.Builder 

dimdimych:
In order to improve `schema.transform()` and `objectType.transform()` fluency, I suggest this pull request with the following changes:

Add the following methods to GraphQLSchema.Builder. 
* `GraphQLSchema.Builder query(Consumer<GraphQLObjectType.Builder> transformer)`
* `GraphQLSchema.Builder mutation(Consumer<GraphQLObjectType.Builder> transformer)`
* `GraphQLSchema.Builder subscription(Consumer<GraphQLObjectType.Builder> transformer)`
When invoked they transform existing operation type or create a new one 

Add `GraphQLObjectType.Builder.field(String name, <GraphQLFieldDefinition.Builder> transformer)`
When invoked transforms existing object type field or creates a new one

### DIFF
--- a/src/main/java/graphql/schema/GraphQLObjectType.java
+++ b/src/main/java/graphql/schema/GraphQLObjectType.java
@@ -309,6 +309,17 @@ public class GraphQLObjectType implements GraphQLNamedDescriptionType, GraphQLNa
             return field(builder.build());
         }
 
+        public Builder field(String name, Consumer<GraphQLFieldDefinition.Builder> transformer) {
+            assertNotNull(transformer, "builderFunction can't be null");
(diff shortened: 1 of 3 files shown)


andimarek:
thanks for the PR @dimdimych, but we will not merge it.

We recently added `SchemaTransformer` which allows you to transform the whole schema.

The individual `transform` methods are only used to transform one object. Have a look at https://github.com/graphql-java/graphql-java/blob/master/src/test/groovy/graphql/schema/SchemaTransformerTest.groovy#L54 to get an idea how the new SchemaTransformer works.

Best regards,
Andi

PR #1745, opened 2020-01-16
add new coercion method

andimarek:
add new internalInputValueToAst method to Coercing to convert internal intput value into ast.

See https://github.com/graphql/graphql-js/issues/1817 for the same GraphQL.js issue


andimarek:
This is outdated

PR #1754, opened 2020-01-17
Changed placement of string encoding Issue #1733

Sajan-99:
According to (https://github.com/graphql-java/graphql-java/issues/1733) the order of the encoding should be done later so this was moved to AstPrinter from AstValueHelper. The relevant tests were also moved to the AstPrinterTest file to test the string returned by StringValues in value(). Failing tests in the AstValueHelperTest file have been modified to ensure correctness with the issue.


src/main/java/graphql/language/AstValueHelper.java
```diff
@@ -103,7 +104,7 @@
             }
 
             // String types are just strings but JSON'ised
```
andimarek: we should remove this comment

PR #1756, opened 2020-01-17
Add new schema printer option includeDirectiveDefinitions

smyrick:
Apollo Federation requires that we return the schema as part of the _Service field to help the gateway understand all the fields. See 'Fetch Service Capabilities' https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#fetch-service-capabilities. 

This schema should not contain the directive definitions but should contain all the directives on the types and fields. This PR adds a new option to the `SchemaPrinter`, `includeDirectiveDefinitions`, which is on by default. Setting it to false will remove the definitions from the top of the schema

### DIFF
--- a/src/main/java/graphql/schema/idl/SchemaPrinter.java
+++ b/src/main/java/graphql/schema/idl/SchemaPrinter.java
@@ -89,6 +89,8 @@ public class SchemaPrinter {
 
         private final boolean includeSchemaDefinition;
 
+        private final boolean includeDirectiveDefinitions;
+
         private final boolean descriptionsAsHashComments;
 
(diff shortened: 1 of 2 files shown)


smyrick:
Related code in graphql-kotlin: https://github.com/ExpediaGroup/graphql-kotlin/blob/61326159d0c7938f41d98e2e3566aa327f672769/graphql-kotlin-federation/src/main/kotlin/com/expediagroup/graphql/federation/FederatedSchemaGeneratorHooks.kt#L73-L80

smyrick:
Resolved merged conflicts

PR #1761, opened 2020-01-19
failing test to show the problem with unvalidated default values

andimarek:


andimarek:
this is solved on master

PR #1769, opened 2020-01-22
Adding tests to prove scalars are omitted if not used

simeonleatherland:
- Validate schemas created by sdl and programmatically don't contain built in scalars and show custom scalars that aren't referenced.


simeonleatherland:
Commits include:
* Changing default of SchemaPrinter to print scalars and extended scalars as required for graphql spec
* Adding tests to conclude that schema printing through a programatically created schema doesn't print as expected for unused custom scalars
* It is expected that **4 tests will fail**

PR #1779, opened 2020-01-29
Issue #1768 ExecutionInput: allow to set value a top level

dimdimych:
Here's the push request it's a proposal for the issue #1768

### DIFF
--- a/src/test/groovy/graphql/Issue1768.groovy
+++ b/src/test/groovy/graphql/Issue1768.groovy
@@ -1,0 +1,81 @@
+package graphql
+
+import graphql.schema.DataFetcher
+import graphql.schema.GraphQLObjectType
+import graphql.schema.GraphQLSchema
+import graphql.schema.GraphQLSchemaElement
+import graphql.schema.GraphQLTypeVisitorStub
+import graphql.schema.SchemaTransformer
+import graphql.schema.idl.NaturalEnumValuesProvider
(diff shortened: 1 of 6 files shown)


dimdimych:
Hm... It looks like some tests not related to my fix are failing

andimarek:
Thanks for the PR @dimdimych. Test test are failing because your change, please have a look again. thanks

src/test/groovy/graphql/Issue1768.groovy
```diff
@@ -0,0 +1,81 @@
...
+        when:
+        ExecutionResult result = graphql.execute {
+            it.query(" { dummy } ")
+        }
+
+        then:
+        result.data.dummy == null
+
+    }
+    def "#1768 test if local context is set for top level"() {
```
andimarek: can you remove the reference to the issue number? thanks

src/test/groovy/graphql/Issue1768.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+        when:
+        ExecutionResult result = graphql.execute {
+            it.localContext("test").query(" { dummy } ")
+        }
+
+        then:
+        result.data.dummy == "test"
+    }
+    def "#1768 test if the top local context gets transferred to the next level"() {
```
andimarek: same here, please remove the reference to the issue

PR #1780, opened 2020-01-29
Removing java types as standard scalars

simeonleatherland:
* Removed flag for SchemaPrinting to include Extended Schema types
* Removed types from ScalarInfo so java types are treated as a custom scalar


src/test/groovy/graphql/schema/idl/SchemaParserTest.groovy
```diff
@@ -114,7 +114,7 @@ class SchemaParserTest extends Specification {
         typeExtensions.size() == 1
         typeExtensions.get("Query") != null
 
-        scalarTypes.size() == 12 // includes standard scalars
+        scalarTypes.size() == 6 // includes standard scalars
```
andimarek: maybe make this comment clear

PR #1797, opened 2020-02-25
add missing WiringFactory methods to CombinedWiringFactory

laszlovandenhoek:
The `WiringFactory` interface gained a couple of methods over time, but its implementor `CombinedWiringFactory` was not expanded accordingly to call those methods on its delegates. This appears to be an oversight which can lead to surprising behavior.


laszlovandenhoek:
I would be glad to provide a backport to the 14.x branch, once someone points me to it :-)

PR #1849, opened 2020-04-06
 #1806 input_object_field_uniqueness

dugenkui03:


dugenkui03:
RP releated to #1806 .

See [5.6.3 Input Object Field Uniqueness](https://spec.graphql.org/June2018/#sec-Input-Object-Field-Uniqueness).

PR #1862, opened 2020-04-13
Added test that shows NPE issue with query complexity 

paulbakker:
Calculating query complexity for the following query fails with a NullPointerException.

```
query Hello($name:String="Someone") {
    hello(name: $name)
}
```
The critical bit is the default value for the `$name` variable. Without a default value the NPE doesn't happen.

I've added a test that currently fails and shows the issue. The NPE is thrown in `NodeVisitorWithTypeTracking` (line 238). I don't understand the visitor code well enough to correctly fix the issue, but the test should be helpful doing so.


andimarek:
Hi @paulbakker, this is fixed on master and will be releases in 15.0. Thanks

PR #1867, opened 2020-04-17
1865: Do not visit a variable definition value node

### LINKED ISSUE #1865: Query complexity calculation fails with NPE
Calculating query complexity for the following query fails with a NullPointerException.

query Hello($name:String="Someone") {
    hello(name: $name)
}
The critical bit is the default value for the $name variable. Without a default value the NPE doesn't happen.

I've provided a test that shows the issue in PR #1862.

felipe-gdr:
There are no visitors for Variable Definitions, so trying to visit
its value to extract data in order to create an environment object
will throw an NPE.

fix #1865


andimarek:
thanks @felipe-gdr

PR #1871, opened 2020-04-19
added default value when using variable references. Fix #1870

### LINKED ISSUE #1870: Default value when using variable references
**Describe the bug**

Having the following type, and an associated data fetcher which returns `hello + name` 

```graphql
type Query {
   sayHello(name: String = "amigo"): String
}
```
when executing the following query:

```graphql
{
    sayHello
}
```

The result is the expected `hello amigo`, **but** when using variable references and not passing any variable values, like the following query:

```graphql
query SayHello($name: String) {
    sayHello(name: $name)
}
```

It returns `hello null`.

The  [GraphQL reference](https://spec.graphql.org/June2018/#sec-Execution) says:

*If no value is provided for a defined input object field and that field definition provides a default value, the default value should be used.*

So I think even with variable references the default value should be used.

**To Reproduce**

You can copy paste this test into **ValuesResolverTest**

```groovy
def "getArgumentValues: uses default value with null variable reference value"() {
        given: "schema defining input object"
        def inputObjectType = newInputObject()
                .name("inputObject")
                .build()

        def fieldArgument = new GraphQLArgument("arg", "", inputObjectType, "hello")
        def argument = new Argument("arg", new VariableReference("var"))

        when:
        def variables = [:]
        def values = resolver.getArgumentValues([fieldArgument], [argument], variables)

        then:
        values['arg'] == 'hello'
    }
```

mariogarcia:


andimarek:
Thanks. This is related to https://github.com/graphql-java/graphql-java/issues/1870 and we need check that this fix what it suppose to do.

One critical aspect is that we need to differentiate between `null` value and not provided value.

andimarek:
I took your commit and incorporated it into #1912

Issue #1879, opened 2020-04-25
validation error


dugenkui03:
**Describe the bug**
type definition:
```
type Query {
  fieldX(arg: [Long]): String
}
```

the query below is valid
```
query{
    fieldX (arg: 5 )
}
```

dugenkui03:
> **Describe the bug**
> type definition:
> 
> ```
> type Query {
>   fieldX(arg: [Long]): String
> }
> ```
> 
> the query below is valid
> 
> ```
> query{
>     fieldX (arg: 5 )
> }
> ```
match the specification
https://spec.graphql.org/draft/#sec-Type-System.List.Input-Coercion

PR #1880, opened 2020-04-25
issue1879-validate-array-argument

dugenkui03:
details in #1879


dugenkui03:
flow the specification, not bug :https://github.com/graphql-java/graphql-java/issues/1793

Issue #1765, opened 2020-01-20
Allow directives on variable definitions 


andimarek:
https://github.com/graphql/graphql-spec/pull/510 not sure how we handle this atm.

dugenkui03:
> [graphql/graphql-spec#510](https://github.com/graphql/graphql-spec/pull/510) not sure how we handle this atm.

I have done some work about **variable directive**, the main change as follows:

1. change the code in `GraphqlOperation.g4`、`GraphqlAntlrToLanguage`、`GraphQLDirective` 、`Introspection` to support 'variable directive';

2. add field `directives` in `VariableDefinition`;
3. change some code about **directive** in `KnownDirectives`、`AstPrinter`、`TestUtil.groovy`,；

4. add groovy test in `VariableDirectiveTest.groovy` to certify that:
   - the declared **variable directive** can't be used in wrong position other than variable;
   - program can get variable directive information from parsed Document.

Is that enough to support **variable directive**?

And should I  add method `instrumentVariables` in `Instrumatation`, which will be called before or around [Execution.execute.valuesResolver.coerceVariableValues](https://github.com/graphql-java/graphql-java/blob/c8de10ed221a4291d9c89455cb39dff56dc90720/src/main/java/graphql/execution/Execution.java#L73), to support the ability described in [spec](https://spec.graphql.org/draft/#sec-Language.Directives): Directives provide a way to describe **alternate runtime execution** and type validation behavior in a GraphQL document.

andimarek:
done

PR #1890, opened 2020-05-04
allow directives on variable definitions

dugenkui03:
Load to #1765 , Allow directives on variable definitions.

### DIFF
--- a/src/main/java/graphql/language/VariableDefinition.java
+++ b/src/main/java/graphql/language/VariableDefinition.java
@@ -22,6 +22,7 @@ public class VariableDefinition extends AbstractNode<VariableDefinition> impleme
     private final String name;
     private final Type type;
     private final Value defaultValue;
+    private final List<Directive> directives;
 
     public static final String CHILD_TYPE = "type";
     public static final String CHILD_DEFAULT_VALUE = "defaultValue";
@@ -30,6 +31,7 @@ public class VariableDefinition extends AbstractNode<VariableDefinition> impleme
     protected VariableDefinition(String name,
                                  Type type,
                                  Value defaultValue,
+                                 List<Directive> directives,
                                  SourceLocation sourceLocation,
                                  List<Comment> comments,
                                  IgnoredChars ignoredChars,
@@ -38,6 +40,7 @@ public class VariableDefinition extends AbstractNode<VariableDefinition> impleme
         this.name = name;
         this.type = type;
         this.defaultValue = defaultValue;
+        this.directives=directives;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/main/java/graphql/schema/GraphQLDirective.java
```diff
@@ -209,6 +224,7 @@ public Builder(GraphQLDirective existing) {
             this.onOperation = existing.isOnOperation();
             this.onFragment = existing.isOnFragment();
             this.onField = existing.isOnField();
+            this.onVariableDefinition=existing.isOnVariableDefinition();
```
andimarek, line 227: the formatting here is off a bit

src/main/java/graphql/schema/GraphQLDirective.java
```diff
@@ -51,8 +52,9 @@ public GraphQLDirective(String name,
                             List<GraphQLArgument> arguments,
                             boolean onOperation,
                             boolean onFragment,
-                            boolean onField) {
-        this(name, description, locations, arguments, onOperation, onFragment, onField, null);
+                            boolean onField,
+                            boolean onVariableDefinition) {
+        this(name, description, locations, arguments, onOperation, onFragment, onField,onVariableDefinition,null);
```
andimarek, line 57: we should not add new constructors.

src/main/java/graphql/language/VariableDefinition.java
```diff
@@ -50,7 +53,22 @@ protected VariableDefinition(String name,
...
+
+    /**
+     * alternative to using a Builder for convenience
+     *
+     * @param name         of the variable
+     * @param type         of the variable
+     * @param defaultValue of the variable
+     * @param directives   of the variable
+     */
+    public VariableDefinition(String name,
```
andimarek, line 67: we should not add any new constructors, but force new code to use the builder

src/main/java/graphql/schema/GraphQLDirective.java
```diff
@@ -342,6 +358,19 @@ public Builder onField(boolean onField) {
             return this;
         }
 
+        /**
+         * @param onVariableDefinition onVariableDefinition
+         *
+         * @return this builder
+         *
+         * @deprecated Use {@code graphql.schema.GraphQLDirective.Builder#validLocations(DirectiveLocation...)}
```
andimarek, line 366: lets not add new methods which are immediately deprecated.

src/main/java/graphql/schema/GraphQLDirective.java
dugenkui03: that is not a new constructor, but added new field(onVariableDefinition) in constructor

src/main/java/graphql/language/VariableDefinition.java
dugenkui03: thanks for you advice, I have delete this new constructor.

src/main/java/graphql/schema/GraphQLDirective.java
dugenkui03: copy that.

dugenkui03:
> thanks a lot.
> In general it looks fine, but I left a few comments.
> Also: inside the graphql java project is an IntelliJ formatter which we use to format. If you use intellij please apply it.
> 
> Thanks!

Thanks for your review and advice. I have update IntelliJ setting and this PR. 

ps: `isOnVariableDefinition` in `GraphQLDirective` haven't removed, because it is used in some method, such as `Introspection` and `GraphQLDirective` and `KnownDirectives`. should I replace it with other methods.

andimarek:
Thanks @dugenkui03: I removed a not needed boolean and merged it.

Issue #1897, opened 2020-05-16
Wrong assumption on argument count for implementation of interfaces


cjanietz:
**Describe the bug**
Currently this library assumes and validates that the amount of arguments a field has on the interface and implementing object has to be the same. However this assumption is not true according to GraphQL Spec (see: https://github.com/graphql/graphql-spec/blob/master/spec/Section%203%20--%20Type%20System.md - IsValidImplementation(type, implementedType)). An implementing type may add optional arguments to the field. The library will then produce an error as:
object type 'Person' does not implement interface 'User' because field 'groups' has a different number of arguments

**To Reproduce**
```
interface User {
  id: String!
  groups: UserGroupList!
}

type Person implements User {
  id: String!
  groups(name: StringFieldFilter): UserGroupList!
}
```

cjanietz:
The implementation in  #1884 seems to fix this, waiting for it to be merged

felipe-gdr:
@cjanietz , #1884 actually doesn't fix this. 
I'll look into fixing this one as well.

PR #1920, opened 2020-05-24
1897: Fix field argument validation in SchemaValidator

felipe-gdr:
Implementing type **can** declare more field arguments than in the implemented interface given that the extra arguments are optional (aka not non-null).

fixes: #1897

### DIFF
--- a/src/test/groovy/graphql/schema/validation/TypesImplementInterfacesTest.groovy
+++ b/src/test/groovy/graphql/schema/validation/TypesImplementInterfacesTest.groovy
@@ -2,9 +2,12 @@ package graphql.schema.validation
 
 import graphql.TypeResolutionEnvironment
 import graphql.schema.GraphQLInterfaceType
+import graphql.schema.GraphQLNonNull
 import graphql.schema.GraphQLObjectType
 import graphql.schema.TypeResolver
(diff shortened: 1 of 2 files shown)


src/test/groovy/graphql/schema/validation/TypesImplementInterfacesTest.groovy
```diff
@@ -317,4 +320,82 @@ class TypesImplementInterfacesTest extends Specification {
         goodErrorCollector.getErrors().isEmpty()
     }
 
+    def "type can declare extra optional field arguments"() {
```
andimarek, line 323: we should add also an SDL based test, because we are currently not always consistent.

felipe-gdr:
fixes: #1897

src/test/groovy/graphql/schema/validation/TypesImplementInterfacesTest.groovy
felipe-gdr: There are current tests that cover that for the SDL already in `graphql.schema.idl.SchemaTypeCheckerTest`.

- `test field arguments on object can contain additional optional arguments`

Currently, the SDL validation doesn't reproduce the error described in the ticket.

PR #1923, opened 2020-05-24
fix variable directive

dugenkui03:
make directives to be children of VariableDefinition.

### DIFF
--- a/src/main/java/graphql/language/VariableDefinition.java
+++ b/src/main/java/graphql/language/VariableDefinition.java
@@ -17,7 +17,7 @@ import static graphql.language.NodeChildrenContainer.newNodeChildrenContainer;
 import static java.util.Collections.emptyMap;
 
 @PublicApi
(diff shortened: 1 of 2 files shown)


andimarek:
thanks, you are right, this is missing at the moment.

Can you please add  directives to the `getNamedChildren` method in `VariableDefinition`? thanks

dugenkui03:
> thanks, you are right, this is missing at the moment.
> 
> Can you please add directives to the `getNamedChildren` method in `VariableDefinition`? thanks

Thanks for point out the omission.

dugenkui03:
The RP haven been change and squashed into a single commit.

dugenkui03:
> I think the Copy Constructor of the Builder needs also to deal with directives.

The RP haven been change  and make VariableDefinition implement DirectivesContainer.

andimarek:
thanks a lot

PR #1926, opened 2020-05-26
1897: Fix argument implementation in SchemaValidator

felipe-gdr:
I've missed one scenario on the my previous PR (#1920).

Since now we're no longer relying on argument positions to do the validation, it doesn't make sense to compare argument count anymore.


felipe-gdr:
@andimarek it would be nice if we could add this fix to the v15 release.

Issue #1944, opened 2020-06-08
GraphQL-java uses toString instead of ConnectionCursor getValue


pal-thomassen:
**Describe the bug**
When I make an implementation of the ConnectionCursor interface, I implement the `getValue`-method for the interface. However, GraphQL java is using toString() for the implementation instead of the `getValue`-method from the interface.  

**To Reproduce**
See Pull-request https://github.com/graphql-java/graphql-java/pull/1945 where this issue is hopefully resolved.

PR #1945, opened 2020-06-08
Coerce the ConnectionCursor by using the interface

pal-thomassen:
The DefaultConnectionCursor had a convenient toString()-method which used the value from
the getValue method in the interface ConnectionCursor. This made the ScalarsIDTest to
pass by accident. This commit will change the default toString()-method to be more aligned
with the other relay implementations and add a check for ConnectionCursor interface in the
ScalarID coercing code.

I'm not sure if this is the correct solution so a discussion is welcome.  I stumpled upon this issue #1944


andimarek:
thanks for that @pal-thomassen 
Unfortunately I am not convinced this the way we should go: graphql java should not have any specific dependency on the relay package in my opinion.

I think the bigger question is here how/if objects should influence the behaviour of build in Scalars.

PR #1955, opened 2020-06-13
New validation rules about type, field, argument

dugenkui03:
The validation about GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType, GraphQLEnumType, GraphQLInputObjectType, GraphQLScalarType.

- Types must define one or more fields;
- Enum type must define one or more enum values;
- Union type must include one or more unique member types;
- The member types of a Union type must all be Object base types;
- Non‐Null type must not wrap another Non‐Null type;
- Invalid name begin with "__" (two underscores).
 
Details in [specification: TypeSystem](https://spec.graphql.org/June2018/#sec-Type-System)


dugenkui03:
The validation about GraphQLObjectType, GraphQLInterfaceType, GraphQLUnionType, GraphQLEnumType, GraphQLInputObjectType, GraphQLScalarType.

- Types must define one or more fields;
- Enum type must define one or more enum values;
- Union type must include one or more unique member types;
- The member types of a Union type must all be Object base types;
- Non‐Null type must not wrap another Non‐Null type;
- Invalid name begin with "__" (two underscores).
 
Details in [specification: TypeSystem](https://spec.graphql.org/June2018/#sec-Type-System)

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+        if (graphQLNamedTypes == null || graphQLNamedTypes.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        return graphQLNamedTypes.stream().filter(type -> !isScalarOrIntrospectionTypes(type)).collect(Collectors.toList());
+    }
+
+    private boolean isScalarOrIntrospectionTypes(GraphQLNamedType type) {
+        if (type instanceof GraphQLScalarType && (
+                type.getName().equals("Int") || type.getName().equals("Float") || type.getName().equals("String") ||
```
andimarek, line 245: we should only treat the build-in scalar types (Int, Float, String, Boolean and ID) in a special way. The others are not really special and should be treated as every other custom scalar.

We also have a helper method in Scalars (I believe) to determine if the type is a build-in one or not

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private void validateUnion(GraphQLUnionType type, SchemaValidationErrorCollector errorCollector) {
+        assertTypeName(type.getName(), errorCollector);
+
+        UnionTypeDefinition definition = type.getDefinition();
+        if (definition == null) {
+            return;
+        }
+
+        List<Type> memberTypes = definition.getMemberTypes();
```
andimarek, line 134: we should not need to check the AST definition here. Just getting the members of the type should be enough.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+        } else if (type instanceof GraphQLEnumType) {
+            validateEnum((GraphQLEnumType) type, errorCollector);
+        } else if (type instanceof GraphQLInputObjectType) {
+            validateInputObject((GraphQLInputObjectType) type, errorCollector);
+        } else if (type instanceof GraphQLScalarType) {
+            validateScalar((GraphQLScalarType) type, errorCollector);
+        }
+    }
+
+    private void validateObject(GraphQLObjectType type, SchemaValidationErrorCollector errorCollector) {
```
andimarek, line 85: can we combine validateObject and validateInterface?
I would also suggest a more prescriptive name like `validateContainsField` or so.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+                        type.getName().equals("BigInteger") || type.getName().equals("BigDecimal") || type.getName().equals("Char")
+        )) {
+            return true;
+        }
+
+        if (type instanceof GraphQLEnumType && (type.getName().equals("__DirectiveLocation") || type.getName().equals("__TypeKind"))) {
+            return true;
+        }
+
+        if (type instanceof GraphQLObjectType && (
```
andimarek, line 256: IntrospectionQuery is not a special type at all (it is just a name).

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private List<GraphQLNamedType> filterScalarAndIntrospectionTypes(List<GraphQLNamedType> graphQLNamedTypes) {
+        if (graphQLNamedTypes == null || graphQLNamedTypes.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        return graphQLNamedTypes.stream().filter(type -> !isScalarOrIntrospectionTypes(type)).collect(Collectors.toList());
+    }
+
+    private boolean isScalarOrIntrospectionTypes(GraphQLNamedType type) {
```
andimarek, line 243: could this whole method just check for `__` prefix?

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+            errorCollector.addError(schemaValidationError);
+        }
+    }
+
+    private List<GraphQLNamedType> filterScalarAndIntrospectionTypes(List<GraphQLNamedType> graphQLNamedTypes) {
+        if (graphQLNamedTypes == null || graphQLNamedTypes.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        return graphQLNamedTypes.stream().filter(type -> !isScalarOrIntrospectionTypes(type)).collect(Collectors.toList());
```
andimarek, line 240: We have static util methods  (in `FpKit` I believe), please use them instead of using directly `filter` etc.

src/test/groovy/graphql/schema/validation/TypeAndFieldRuleTest.groovy
```diff
@@ -0,0 +1,123 @@
...
+
+import static graphql.Scalars.GraphQLString
+import static graphql.schema.GraphQLFieldDefinition.newFieldDefinition
+import static graphql.schema.GraphQLObjectType.newObject
+import static graphql.schema.GraphQLSchema.newSchema
+
+class TypeAndFieldRuleTest extends Specification {
+
+
+    def "type must define one or more fields."() {
```
andimarek, line 18: we should add tests for Interfaces here as well.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        List<Type> memberTypes = definition.getMemberTypes();
+        if (memberTypes == null || memberTypes.size() == 0) {
+            SchemaValidationError validationError =
+                    new SchemaValidationError(SchemaValidationErrorType.UnionTypeLackOfTypeError, String.format("Union type \"%s\" must include one or more unique member types.", type.getName()));
+            errorCollector.addError(validationError);
+        }
+
+        List<String> typeNames = new ArrayList<>();
```
andimarek, line 141: I think a Set is better here as track just the names we already saw.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: graphql-java framework contains a type named [IntrospectionQuery](https://github.com/graphql-java/graphql-java/blob/10eeacc3979624964ef10ea1078ef84534d9c5cc/src/main/java/graphql/introspection/Introspection.java#L556), which contain field start with '__'.

src/test/groovy/graphql/schema/validation/TypeAndFieldRuleTest.groovy
dugenkui03: done

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: graphql-java framework contains a type named [IntrospectionQuery](https://github.com/graphql-java/graphql-java/blob/10eeacc3979624964ef10ea1078ef84534d9c5cc/src/main/java/graphql/introspection/Introspection.java#L556), which contains fields start with '__'.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: Thanks for your suggestion.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: Thanks for your suggestion.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: Thanks for your suggestion.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: `types` in `GraphQLUnionType.Builder` is a Map([code link](https://github.com/graphql-java/graphql-java/blob/10eeacc3979624964ef10ea1078ef84534d9c5cc/src/main/java/graphql/schema/GraphQLUnionType.java#L195)), which will lost the type with duplicated name. 

Maybe it will be better to move this check to `SchemaTypeChecker`.

dugenkui03: I will followed your suggestion, because types definied in GraphQLUnionType is list. 

Then I will add `unique member check` to `SchemaTypeChecker` in another PR.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: Replace it by `filterBuiltInTypes`, which aimed to filter Built-In Types.

src/test/groovy/graphql/schema/validation/TypeAndFieldRuleTest.groovy
dugenkui03: Thanks for your suggestion.

src/main/java/graphql/schema/validation/TypeAndFieldRule.java
dugenkui03: Thanks for your suggestion.

dugenkui03:
@andimarek make some changes and rebase to second commit:
more test about rule; remove code where used AST definition.

dugenkui03:
hi @andimarek ,  appreciate your time for this. is there any thing else I need to do for this PR?

I think is is pretty important to improve schema validation in future release version.

tfij:
Hi, is there any way to help in review of this PR?
Correct schema validation is very much expected by my.

PR #1957, opened 2020-06-14
Bugfix: Transforming an argument should not set a default value

### LINKED ISSUE #1956: Transforming GraphQLArgument accidentally sets a default value
**Describe the bug**
`GraphQLArgument#transform` accidentally sets the default value to `null` when no default value already exists.

**To Reproduce**
```groovy
        def startingArgument = GraphQLArgument.newArgument().name("A1")
                .type(GraphQLInt)
                .build()
        when:
        def transformedArgument = startingArgument.transform({
            it.name("A2")
        })

        then:

        transformedArgument.name == "A2"
        !transformedArgument.hasSetDefaultValue()
```

kaqqao:
Fixes #1956


andimarek:
Thanks for that

PR #1959, opened 2020-06-15
[NOT READY] Distinguish null default from no default in input fields

### LINKED ISSUE #1958: GraphQLInputObjectField does not distinguish between a null default value and no default value
**Describe the bug**
Unlike `GraphQLArgument`, `GraphQLInputObjectField` does not distinguish between a `null` default value and no default value. I.e. `GraphQLArgument#hasSetDefaultValue` exists but `GraphQLInputObjectField#hasSetDefaultValue` doesn't.

This is a surprisingly inconsistent behavior, so probably unintentional.

kaqqao:
Fixes #1958

Not ready because I'm not 100% sure if there's more places in `ValuesResolver`  that need changing (although I don't think so), or whether e.g. [this line](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/SchemaPrinter.java#L589) needs anything (again, I don't think so).

### DIFF
--- a/src/main/java/graphql/execution/ValuesResolver.java
+++ b/src/main/java/graphql/execution/ValuesResolver.java
@@ -210,7 +210,7 @@ public class ValuesResolver {
     }
 
     private boolean alwaysHasValue(GraphQLInputObjectField inputField) {
-        return inputField.getDefaultValue() != null
+        return inputField.hasSetDefaultValue()
                 || isNonNull(inputField.getType());
     }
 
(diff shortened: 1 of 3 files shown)


andimarek:
Thanks @kaqqao for this.

I think the line you mentioned needs some tweaking: the `non null` check is used as a way to ask "is there a default value". So I think we should use the new method `hasDefaultValue`.

andimarek:
One concrete use case that should be fixed with this PR is null default values in Input object `null` default values:
```
type Query {
  foo(arg: Arg): String
}
input Arg {
  field: String = null;
}
```
This should successfully be parsed and then printed again with your fix. I assume this is currently not working correctly.

andimarek:
This is solved and will be releases with 17

PR #1962, opened 2020-06-20
Rename ExecutionPath to ResultPath

andimarek:
ExecutionPath is not a great name: the path identifies one "node" in the result. ResultPath is better.

This is a breaking change, but should be minor:
- Normally you don't import ExecutionPath directly, therefore it will not break anything.
- If you import ExecutionPath directly the compiler will tell you that something is wrong.


williamboman-pp:
> Normally you don't import ExecutionPath directly, therefore it will not break anything.

I think this might be more common than one would expect, due to people subclassing `ExceptionWhileDataFetching`. Would be nice if this was included in the changelogs (although as you mention compiler does all the work in locating the issue, but I had to refer to graphql-java's git history to understand what it had been renamed to)

PR #1966, opened 2020-06-24
union type check

dugenkui03:
UnionType check, details in https://spec.graphql.org/June2018/#sec-Type-System. 
Especially in https://spec.graphql.org/June2018/#sec-Unions
        
           
Invalid name begin with "__" (two underscores);
Union type must include one or more member types;
The member types of a Union type must all be Object base types;
The member types of a Union type must be unique.


dugenkui03:
wrong source branch

PR #1967, opened 2020-06-24
add union type checker

dugenkui03:
union type check, details in https://spec.graphql.org/June2018/#sec-Type-System. 
especially in https://spec.graphql.org/June2018/#sec-Unions
        
           
Invalid name begin with "__" (two underscores);
Union type must include one or more member types;
The member types of a Union type must all be Object base types;
The member types of a Union type must be unique.


bbakerman (approved):
Thanks for excellent PR.  Great code and great tests


Sorry for the long delay in us reviewing this PR / issue.  Work, home and covid have conspired against us time wise and we haven't put in the attention that the graphql-java project needs.

PR #1984, opened 2020-07-22
Include class loader in caching key

### LINKED ISSUE #1983: Need to include class loader in caching keys
`PropertyDataFetcherHelper` performs caching for properties, but the caching key only consists of class name and property name: https://github.com/graphql-java/graphql-java/blob/v15.0/src/main/java/graphql/schema/PropertyDataFetcherHelper.java#L269 . This becomes a problem in application servers, where multiple GraphQL applications might run within one JVM, in case that they share some model classes - if these application classes are then loaded by different class loaders, their caching keys will clash, and when one application retrieves a property that was cached for another application, it will run into `java.lang.IllegalArgumentException: object is not an instance of declaring class` when executing queries.
A possible solution is to include the hashCode of the owning class loader for caching keys.

jmartisk:
Fixes #1983

### DIFF
--- a/src/main/java/graphql/schema/PropertyDataFetcherHelper.java
+++ b/src/main/java/graphql/schema/PropertyDataFetcherHelper.java
@@ -266,7 +266,12 @@ public class PropertyDataFetcherHelper {
     }
 
     private static String mkKey(Class<?> clazz, String propertyName) {
-        return clazz.getName() + "__" + propertyName;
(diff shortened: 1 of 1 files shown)


bbakerman (approved):
I am happy to accept this change as is because its not a bad caching key like this.

I was wary of some "in memory class loader" cache item but this works for me

PR #2002, opened 2020-08-04
Proposed fix for #2001 Allow non-list values on schema directive arguments of list type

dimdimych:
This is proposed fix for #2001


dimdimych:
Ah crap! Hold on, fixing it!

bbakerman (approved):
Thanks for this PR

Sorry for the long delay in us reviewing this PR / issue.  Work, home and covid have conspired against us time wise and we haven't put in the attention that the graphql-java project needs.

PR #2009, opened 2020-08-20
Make UnknownOperationException implement GraphQLError

gnawf:
Useful if we are catching `GraphQLErrors` from execution and putting them as errors in the GraphQL response. With this change in place we don't have to have custom handling.


bbakerman (approved):
Looks good

PR #2011, opened 2020-08-25
upgrade to gradle 6.6 and a bunch of other dependencies

andimarek:
Note: the build in OSGI plugin doesn't work anymore with gradle 6.6.


build.gradle
```diff
@@ -51,30 +54,30 @@ jar {
 }
 
 dependencies {
-    compile 'org.antlr:antlr4-runtime:4.7.2'
-    compile 'org.slf4j:slf4j-api:' + slf4jVersion
-    compile 'com.graphql-java:java-dataloader:2.2.3'
-    compile 'org.reactivestreams:reactive-streams:' + reactiveStreamsVersion
+    implementation 'org.antlr:antlr4-runtime:4.8'
```
andimarek, line 57: @bbakerman this is now explicitly API vs implementation detail. I am not completely sure this is correct.

src/test/groovy/graphql/analysis/QueryTraverserTest.groovy
```diff
@@ -1407,9 +1407,9 @@ class QueryTraverserTest extends Specification {
             """)
         QueryTraverser queryTraversal = createQueryTraversal(query, schema)
         QueryVisitorFieldEnvironment env
-        1 * visitor.visitField({ QueryVisitorFieldEnvironmentImpl it ->
-            env = it
-        })
+        1 * visitor.visitField(_) >> { args ->
```
andimarek, line 1410: this is a change in spock (or maybe groovy) which doesn't allow assignment anymore in the argument matcher closure. Therefore we need to capture the arguments in the implementation closure

bbakerman (approved):

Issue #1999, opened 2020-08-03
Fix for getArgumentValues


dugenkui03:
**Describe the bug**
[CoerceArgumentValues()](https://spec.graphql.org/draft/#CoerceArgumentValues()) defined in spec. 

![31598547672_ pic_hd](https://user-images.githubusercontent.com/18216266/91470570-75295380-e8c7-11ea-8c89-0442d608b621.jpg)

In some cases, null value should be used:
1. use null when reference value in variables map is null(<key,null>): **f.III**;
2. use null when argumentValue defined in `arguments` of `Field` is `NullValue`:  **g -> j.I** .

That is the reason for the changes of `getArgumentValuesImpl()`.
```java
            if (value == null
                    // use null when argumentValue defined in `arguments` of `Field` is NullValue
                    && !(argument != null && argument.getValue() instanceof NullValue)
                    // use null when reference value in variables map is null(<key,null>)
                    && !(argument != null && argument.getValue() instanceof VariableReference // hasValue is true
                    && variables.containsKey(((VariableReference) argument.getValue()).getName())) // the value exists,but null
            )
```

Note: defaultValue defined in typeSystem should be used when `hasValue` is false:
1. argument is not provided in Field;
2. variables map doesn't provided value for argument.

**To Reproduce**
Details in #2014 .

PR #2014, opened 2020-08-28
Fix getArgumentValuesImpl according by spec.CoerceArgumentValues()

dugenkui03:
Details in #1999.


dugenkui03:
hi @bbakerman , very appreciate when you have time to spend on this pr review.

the reason for this change have updated on issue, I have try hard to make it clear.

andimarek:
thanks for that!

Issue #2029, opened 2020-09-09
replace isEqualTo with Objects.equals(a,b)


dugenkui03:
I notice that the method `isEqualTo` in `NodeUtil` is equivalent to `Objects.equals(a,b)`,
will you agree to replace `isEqualTo` with `Objects.equals(a,b)`.

Code linkage:     https://github.com/graphql-java/graphql-java/blob/7ee760264b0a596639ee2c34f9e79a4e9e0ad07f/src/main/java/graphql/language/NodeUtil.java#L20

PR #2030, opened 2020-09-09
replace isEqualTo with Objects.equals

dugenkui03:
The reason for this change is for making the code more readable. #2029


bbakerman (approved):
This was written before Java 7 and hence we didnt have this.

Can you please remove the NodeUtil.isEqualTo method since its no longer needed

dugenkui03:
it was remove already :)

PR #2038, opened 2020-09-14
New validation rule: providing the explicit value 'null' to non-null argument is invalid

dugenkui03:
[required Arguments rule](https://spec.graphql.org/draft/#sel-HALTHHFTCAACCnCjmR)
>Providing the explicit value null is also not valid since required arguments always have a non‐null type.

[demo](https://spec.graphql.org/draft/#example-0bc81):
```sql
fragment missingRequiredArg on Arguments {
  nonNullBooleanArgField(nonNullBooleanArg: null)
}
```


bbakerman (approved):
Thanks for this

bbakerman (approved):
Thank you for this PR

PR #2071, opened 2020-10-13
Remove unused import; Remove redundant method

dugenkui03:
- Remove unused import;
- Remove redundant method.


andimarek:
thanks

Issue #2072, opened 2020-10-14
TypesImplementInterfaces check fails for interfaces implementing interfaces


KammererTob:
With v15.0 and the support for interfaces implementing interfaces it seems like the TypesImplementInterfaces fails in some scenarios. This line:

https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/validation/TypesImplementInterfaces.java#L175

only works when one is of type `GraphQLObjectType` and the other of `GraphQLInterfaceType`. There can be cases where both are interfaces now.

```
interface Trait {
    name: String
}

interface MammalTrait implements Trait {
    name: String
}

interface Animal {
    traits: [Trait]
}

type Dog implements Animal {
    traits: [MammalTrait]
}
```

This leads to:
_Caused by: graphql.schema.validation.InvalidSchemaException: invalid schema:
object type 'Dog' does not implement interface 'Animal' because field 'traits' is defined as '[MammalTrait]' type and not as '[Trait]' type_

PR #2073, opened 2020-10-14
Fix TypesImplementInterfaces.check() for interfaces implementing interface #2072 

KammererTob:
This is a fix for #2072.
It adds another check for the case where both types are of `GraphQLInterfaceType` and then checks if the implementing type actually has the correct interface, similar to how `objectImplementsInterface` works. It also includes a test for the case mentioned in the issue.


felipe-gdr (approved):
Thanks for fixing it!

andimarek:
@felipe-gdr do you think this should be backported to 14.x?

felipe-gdr:
Yes @andimarek , let's do that.

PR #2074, opened 2020-10-15
avoiding allocations in FieldCollectorParameters and MergedField

dfa1:
ExecutionContext uses immutable maps for variables and fragments, that could be shared freely.

And, since this is an internal class, we could change a bit the contract and avoid some allocations of LinkedHashMap:

                .variables(executionContext.getVariables())

### DIFF
--- a/src/main/java/graphql/execution/FieldCollectorParameters.java
+++ b/src/main/java/graphql/execution/FieldCollectorParameters.java
@@ -6,7 +6,6 @@ import graphql.language.FragmentDefinition;
 import graphql.schema.GraphQLObjectType;
 import graphql.schema.GraphQLSchema;
 
-import java.util.LinkedHashMap;
 import java.util.Map;
 
 /**
@@ -48,8 +47,8 @@ public class FieldCollectorParameters {
 
     public static class Builder {
         private GraphQLSchema graphQLSchema;
-        private final Map<String, FragmentDefinition> fragmentsByName = new LinkedHashMap<>();
-        private final Map<String, Object> variables = new LinkedHashMap<>();
+        private Map<String, FragmentDefinition> fragmentsByName;
+        private Map<String, Object> variables;
         private GraphQLObjectType objectType;
 
         /**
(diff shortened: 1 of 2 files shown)


dfa1:
For MergedField:

image

src/main/java/graphql/execution/MergedField.java
```diff
@@ -155,7 +156,7 @@ private Builder(MergedField existing) {
         }
 
         public Builder fields(List<Field> fields) {
-            this.fields = fields;
+            this.fields.addAll(fields);
             return this;
```
dfa1, line 160: used only in tests, could be a breaking change.

Maybe this is the root cause of the additional copy that I've just removed in the constructor.

src/main/java/graphql/execution/FieldCollectorParameters.java
```diff
@@ -70,12 +69,12 @@ public Builder objectType(GraphQLObjectType objectType) {
...
         public Builder fragments(Map<String, FragmentDefinition> fragmentsByName) {
-            this.fragmentsByName.putAll(fragmentsByName);
+            this.fragmentsByName = fragmentsByName;
             return this;
         }
 
         public Builder variables(Map<String, Object> variables) {
-            this.variables.putAll(variables);
+            this.variables = variables;
             return this;
```
bbakerman, line 78: This is a tricky one - generally we try for more immutable classes and hence by "copying" we gain some control in mutable Java land

However as this class is repeatedly called, we probably need to change this general pattern if we see evidence that it costs us.

bbakerman (approved):

src/main/java/graphql/execution/FieldCollectorParameters.java
dfa1: yes, I'm not sure too. I did that because the class is marked as @Internal (so we may have more freedom) but I know that it is against the pattern mutable/builder -> immutable/built object. 

It looks like it is always called with **ExecutionContext** data, that is unmodifiable.

dfa1:
Thank you @bbakerman!

I just integrated this new version and run it under the memory profiler. Results are a bit better (1-2%):
image

PR #2076, opened 2020-10-15
Add schema validator to check if object type is not empty

tfij:
Documentation of graphql doesn't allow to define object with no fields : http://spec.graphql.org/draft/#sel-HAHZhCFBABABuBq9O


juriad:
Can you do the same for interfaces and input objects?

tfij:
yes, I fix all tests and do the same for interfaces and input objects

dugenkui03:
With the same purpose as #1955.

PR #2077, opened 2020-10-15
Update antlr to 4.8

phillip-kruger:
I am not sure if there is a reason that antlr4-runtime and antlt4 is on different versions ? In case this is ok to make them the same, here is a PR for that.

We get 

> ANTLR Tool version 4.7.2 used for code generation does not match the current runtime version 4.8

in SmallRye (and Quarkus). I am not sure if this is related.

Thanks :)


andimarek:
I don't think there is a specific reason ... thanks for that.

PR #2083, opened 2020-10-27
Fix for coerceVariableValues

dugenkui03:
Fix for [coerceVariableValues](https://github.com/graphql-java/graphql-java/blob/0de364ae0fdc75126b0217ba6ad61253ec4007c0/src/main/java/graphql/execution/ValuesResolver.java#L54) to ensure that:
1. use null when variable defined in variableValuesMap is null;
2. if variableType is a Non‐Nullable type and value is null, throw a query error.

Linkage to [algorithm defined in spec](https://spec.graphql.org/draft/#sec-Coercing-Variable-Values)

### DIFF
--- a/src/main/java/graphql/execution/ValuesResolver.java
+++ b/src/main/java/graphql/execution/ValuesResolver.java
@@ -68,21 +68,12 @@ public class ValuesResolver {
                 }
             } else {
                 Object value = variableValues.get(variableName);
-                Object coercedValue = getVariableValue(fieldVisibility, variableDefinition, variableType, value);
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/execution/ValuesResolver.java
```diff
@@ -68,23 +68,14 @@
...
                 coercedValues.put(variableName, coercedValue);
             }
         }
 
         return coercedValues;
     }
 
-    private Object getVariableValue(GraphqlFieldVisibility fieldVisibility, VariableDefinition variableDefinition, GraphQLType variableType, Object value) {
-
-        if (value == null && variableDefinition.getDefaultValue() != null) {
```
dugenkui03, line 81: use null when variable defined in variableValuesMap is null

PR #2091, opened 2020-11-02
Minor fix: replace `tracingMap` with `withTracingExt`

dugenkui03:
I think `withTracingExt` is more reasonable and semantically correct,
tracingMap can indicate the result of `tracingSupport.snapshotTracingData()`.


bbakerman (approved):
Ok

PR #2122, opened 2020-12-02
upgrading jmh to 1.25

dfa1:
Following instructions here:
https://github.com/artyushov/idea-jmh-plugin/issues/13#issuecomment-687805558


andimarek:
thanks

PR #2125, opened 2020-12-03
Sort GraphQLSchemaElement by name or toString(), avoid cast error

dugenkui03:


andimarek:
Thanks for that.
The challenge here is that we rely now on `.toString()`. 

We should unwrap the types until we get a named element and then sort by this name.

dugenkui03:
@andimarek This PR was updated for unwrapping GraphQLType:
```
            if (element instanceof GraphQLType) {
                element = unwrapAll((GraphQLType) element);
            }
```

`toString()` seems to be redundant, but I think it is safer than cast GraphQLSchemaElement to GraphQLNamedSchemaElement without any check.

bbakerman (approved):

PR #2126, opened 2020-12-07
Ensure NormalizedQueryTreeFactory handles mutations and subscriptions

### LINKED ISSUE #2117: DataFetchingEnvironment.getSelectionSet().getFields() is broken from Mutations after upgrade to graphql-java 16
**Describe the bug**
Trying to invoke `dataFetchingEnvironment.getSelectionSet().getFields()` from a Mutation results in `graphql.AssertException: no field with name mutationFieldName found in object Query` exception being thrown out.

This seems be caused by https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java#L55 that assume that root is always a Query.

This worked fine in graphql-java 15, and is broken after upgrade to graphql-java 16.

**To Reproduce**
From a DataFetcher that's registered to resolve a Mutation field invoke something like below:

```
DataFetchingFieldSelectionSet selectionSet = environment.getSelectionSet();
if (selectionSet.getFields().stream().allMatch(field -> SUPPORTED_FIELDS.contains(field.getName()))) { return "xyz"; }
```

skevy:
Fixes #2117

### DIFF
--- a/src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java
+++ b/src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java
@@ -9,6 +9,7 @@ import graphql.language.NodeUtil;
 import graphql.language.OperationDefinition;
 import graphql.normalized.FieldCollectorNormalizedQuery.CollectFieldResult;
 import graphql.schema.FieldCoordinates;
+import graphql.schema.GraphQLObjectType;
 import graphql.schema.GraphQLSchema;
 
 import java.util.ArrayList;
@@ -52,7 +53,18 @@ public class NormalizedQueryTreeFactory {
                 .variables(variables)
                 .build();
 
-        CollectFieldResult topLevelFields = fieldCollector.collectFromOperation(parameters, operationDefinition, graphQLSchema.getQueryType());
+        GraphQLObjectType rootType;
+        if (operationDefinition.getOperation() == OperationDefinition.Operation.QUERY) {
+            rootType = graphQLSchema.getQueryType();
+        } else if (operationDefinition.getOperation() == OperationDefinition.Operation.MUTATION) {
+            rootType = graphQLSchema.getMutationType();
+        } else if (operationDefinition.getOperation() == OperationDefinition.Operation.SUBSCRIPTION) {
+            rootType = graphQLSchema.getSubscriptionType();
+        } else {
+            throw new IllegalStateException("Operation type is not valid.");
+        }
+
+        CollectFieldResult topLevelFields = fieldCollector.collectFromOperation(parameters, operationDefinition, rootType);
 
         Map<Field, List<NormalizedField>> fieldToNormalizedField = new LinkedHashMap<>();
         Map<NormalizedField, MergedField> normalizedFieldToMergedField = new LinkedHashMap<>();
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java
```diff
@@ -52,7 +53,18 @@ private NormalizedQueryTree createNormalizedQueryImpl(GraphQLSchema graphQLSchem
...
-        CollectFieldResult topLevelFields = fieldCollector.collectFromOperation(parameters, operationDefinition, graphQLSchema.getQueryType());
+        GraphQLObjectType rootType;
+        if (operationDefinition.getOperation() == OperationDefinition.Operation.QUERY) {
+            rootType = graphQLSchema.getQueryType();
+        } else if (operationDefinition.getOperation() == OperationDefinition.Operation.MUTATION) {
+            rootType = graphQLSchema.getMutationType();
+        } else if (operationDefinition.getOperation() == OperationDefinition.Operation.SUBSCRIPTION) {
+            rootType = graphQLSchema.getSubscriptionType();
+        } else {
+            throw new IllegalStateException("Operation type is not valid.");
```
andimarek, line 64: We use normally `Assert.assertShouldNeverHappen` here.

src/test/groovy/graphql/normalized/NormalizedQueryTreeFactoryTest.groovy
```diff
@@ -853,6 +853,113 @@ type Dog implements Animal{
         coordinatesToNormalizedFields[coordinates("Foo", "subFoo")].size() == 2
     }
 
+    def "handles mutations"() {
```
andimarek, line 856: Another test for Subscriptions would be awesome, but Mutation is good enough.

andimarek (changes requested):
Thanks a lot .. great PR. Please have a look at the comments.

src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java
artem-ag: Hello Andreas, 
`Assert.assertShouldNeverHappen` seems to not work here as compiler complains that `rootType` is not initialized. And it looks like a good compile time check.
Can we commit this (as original author seems to be not available) as is? And replace this exception with something like
```
// Don't use Assert.assertShouldNeverHappen to make sure compiler does not force
// rootType above to be initialized. That way if code structure changes, compiler will
// catch that. Otherwise, we'd get NullPointerException at runtime.
throw new AssertException(
        String.format(
                "Internal error: should never happen: Operation type '%s' is not valid.",
                operationDefinition.getOperation()));
```
in a follow up PR?

artem-ag: @andimarek thoughts?

olim7t: +1 to commit, I'm hitting this issue as well. How about using a return statement to get rid of the initialization issue:
```java
return Assert.assertShouldNeverHappen("Operation type is not valid.");
```

andimarek:
superseded by #2173

PR #2127, opened 2020-12-07
Ensure GraphQLSchema.transform works properly

skevy:
`transform` needs to pass `true` (signifying that the schema is being transformed) to `buildImpl` so that it doesn't throw errors when it finds types in the transformed schema that were previously defined.

Without this, transforming the schema throws an error claiming that "All types within a GraphQL schema must have unique names".


skevy:
I'm still working out a test on this one -- it was easy to reproduce in my schema, but not sure what's actually triggering it yet.

andimarek:
ping @skevy : if you need specific help or have a question ask here.

PR #2129, opened 2020-12-09
Add anonymizer 

andimarek:
New class which takes a schema and queries matching that schema and returns an anonymous version of both. 

All type names, field names, fragment names etc are replaced with generic names like `Fragment1` or `Object10`

### DIFF
--- a/src/main/java/graphql/Directives.java
+++ b/src/main/java/graphql/Directives.java
@@ -119,4 +119,12 @@ public class Directives {
         return new Description(s, null, false);
     }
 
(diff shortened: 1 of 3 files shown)


andimarek:
ping @skevy

skevy:
Nice! I’ll play with this tomorrow. 

Thanks for working on this!

andimarek:
@skevy It is merged so it will be available on the next dev build (that makes the testing easier)

cc @dfa1

PR #2140, opened 2020-12-18
Add subscription to __DirectiveLocation enum

helfer:
Without this, schema introspection on directives will fail if any directive is defined as valid on `SUBSCRIPTION`.


dugenkui03 (approved):

andimarek:
thanks for that

PR #2144, opened 2020-12-18
introducing Field.getResultKey()

dfa1:
This method has several definitions in a lot of places. It is also a good API for advanced users.

@andimarek  @bbakerman 
I'm not sure about the name: any suggestion is welcome :-)


andimarek:
thanks for that @dfa1

PR #2153, opened 2020-12-30
Fix ClassCastException in TypeAndFieldRule

tinnou:
### Description 
This PR fixes two instances of ClassCastException in `TypeAndFieldRule` when the following conditions are met:
- union member types are not object types
- union member types are repeated.

### Testing
It was quite hard to add a unit test since it looks like these two rules are already present in the [UnionTypesChecker](https://github.com/graphql-java/graphql-java/blob/7a8e9038df64ddba27a8aa4040ecdeb60fe387e5/src/main/java/graphql/schema/idl/UnionTypesChecker.java#L69).

Should we remove these two altogether?

### DIFF
--- a/src/main/java/graphql/schema/validation/InvalidSchemaException.java
+++ b/src/main/java/graphql/schema/validation/InvalidSchemaException.java
@@ -2,14 +2,23 @@ package graphql.schema.validation;
 
 import graphql.GraphQLException;
 import graphql.Internal;
+import graphql.VisibleForTesting;
 
 import java.util.Collection;
 
 @Internal
 public class InvalidSchemaException extends GraphQLException {
 
+    private final Collection<SchemaValidationError> errors;
+
     public InvalidSchemaException(Collection<SchemaValidationError> errors) {
         super(buildErrorMsg(errors));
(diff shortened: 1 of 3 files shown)


dugenkui03:
I think that, `SchemaTypeChecker` does some basic validation, and `SchemaValidator` validate whether the schema is valid or not. There isn't specific boundary, and I implement rules by spec as comprehensive as possible.

There are indeed some repeated rule between `SchemaTypeChecker` and `SchemaValidator`. I think it is ok to remove the duplicate rule in  `SchemaTypeChecker/UnionTypesChecker`.

tinnou:
Ok so I did a little bit of digging and we can't remove the rule from `UnionTypesChecker` since it's necessary to actually build the `GraphQLSchema` instance, (otherwise it would throw a ClassCastException while [building](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java#L882-L882) the schema.

What about we leave the fix as is?  The rule is over-defensive but I don't see that as hurting.

andimarek:
thanks

PR #2157, opened 2021-01-07
Fix nested detached types were not added to additionalTypes

tinnou:
### Context
The `GraphQLSchema` holds the list of types. 
For the purpose of this PR, we will differentiate types into two categories. **connected** and **detached** types.
**Connected** types are types that are connected to the root operation types, meaning there is at least one path that leads from the root operation types entry points (Query, Mutation, Subscription) to each of the connected types. 
**Detached** types are types that are not connected to the root operations types.

All the detached types are held in the `GraphQLSchema.additionalTypes` field.
All the connected types AND the additional types are held in the `GraphQLSchema.typeMap` field.

## Bug description
I noticed that in certain cases, **detached** types are not detected properly and end up only in the `GraphQLSchema.typeMap` instead of also being in the `GraphQLSchema.additionalTypes` set.
Specifically, this happens for types that are **nested** within a detached type. 

Here is a simplified example of a test I wrote that showcases the issue:
```groovy
def "nested detached types should be detected as additional types"() {
         def spec = """      
            type Query {
              fieldA : ReferencedA
            }
            
            type ReferencedA {
              field : String
            }

            type UnReferencedA {
              field : UnReferencedNestedE
            }
            
            type UnReferencedNestedE {
                field: String
            }
        """

        def schema = schema(spec)

        expect: "unreferenced types should all be additional types"
        def namedTypes = schema.getAdditionalTypes() as Set<GraphQLNamedType>
        namedTypes.name.toSet() == [
                   "UnReferencedA", 
                   "UnReferencedNestedE"  // fail because this type is not be part of the additionalTypes but it is be part of the schema typeMap
         ].toSet()
}
```

## Root cause and fix
When additionalTypes are being built inside the `SchemaGenerator`, we build them in the same way we build other types. We **recursively** build the types as we traverse their definition using the `BuildContext`. However, in case of **nested** detached types, there is a chance they are registered first on the buildContext [but never added to the additionalTypes collection.](https://github.com/graphql-java/graphql-java/blob/a12f84b98fc6fc592b22c3c9ccb186efe6427648/src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java#L1162-L1162). As we recurse [the types are built and added to the buildContext](http://github.com/graphql-java/graphql-java/blob/a12f84b98fc6fc592b22c3c9ccb186efe6427648/src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java#L982-L982).

In order to fix this bug, we record the additional types even when they already exist on the context.

## Uncovering another bug in [FieldVisibilitySchemaTransformation](FieldVisibilitySchemaTransformation)
After this was fixed, it unearthed a bug in `FieldVisibilitySchemaTransformation`, specifically when a field is marked with `@private` and the field is on a detached type that happens to reference a detached type, it was not properly removing the subsequent type. 

Here is the test that showcases the issue:
```groovy
def "unreferenced types can have fields removed, and the referenced types must be removed as well if they are not used"() {
        given:
        GraphQLSchema schema = TestUtil.schema("""

        directive @private on FIELD_DEFINITION
        
        type Query {
            foo: Foo 
        }
        
        type Foo {
            id: ID
        }
        
        type Bar {
            baz: String
            bing: Bing @private
        }
        
        type Bing { ## Should be removed entirely since the only reference to it was from the Bar:bing field
            id: ID
        }
        """)

        when:
        GraphQLSchema restrictedSchema = visibilitySchemaTransformation.apply(schema)

        then: "Bar.bing field must have been removed"
        (restrictedSchema.getType("Bar") as GraphQLObjectType).getFieldDefinition("bing") == null

        and: "since Bing is not used anywhere else, it should be removed"
        restrictedSchema.getType("Bing") == null //this was failing as the type was still present
    }
```
Looking at the [FieldVisibilitySchemaTransformation](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/transform/FieldVisibilitySchemaTransformation.java#L154) implementation, it seems it might not have been finished and has benefited from the previous bug for the tests to pass. The fix here consists of detecting if there is any other elements still referencing the detached type and perform an extra pass to remove them from the visibility altered schema.

### DIFF
--- a/src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java
+++ b/src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java
@@ -73,6 +73,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Deque;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


bbakerman (approved):
Wow - this PR is a master class in explanation, code and testing.

Thanks for this!

PR #2161, opened 2021-01-15
Use bndtool plugin to generate OSGi manifest headers

benfortuna:
This PR restores support for OSGi using the bndtool gradle plugin


andimarek:
thanks

PR #2162, opened 2021-01-17
Throw an exception when try to register or unregister data loader when no DataLoaderRegistry provided

donbeave:
Hello!

The original implementation is throwing an exception only when the `register` method was invoked but in my reality, I'm using `computeIfAbsent` method to register a new data loader (kind of lazy initialization). Without such changes, if no data loader registry was provided it will not throw any exception and hangs on waiting for the data from the data loader, which in the first place is not very intuitive behavior. I think it's better to throw an exception for all three methods `register`, `computeIfAbsent` and `unregister`, not only one `register`.


andimarek:
thanks

PR #2173, opened 2021-02-01
Ensure NormalizedQueryTreeFactory handles mutations and subscriptions

### LINKED ISSUE #2117: DataFetchingEnvironment.getSelectionSet().getFields() is broken from Mutations after upgrade to graphql-java 16

austinarbor:
Credit to @skevy for original fix here https://github.com/graphql-java/graphql-java/pull/2126 but it seems like the PR has stalled on the assertion error

Creating new PR here with suggested assertion changes

Fixes #2117


src/main/java/graphql/normalized/NormalizedQueryTreeFactory.java
```diff
@@ -1,14 +1,12 @@
...
 
+import graphql.Assert;
 import graphql.Internal;
 import graphql.execution.MergedField;
-import graphql.language.Document;
-import graphql.language.Field;
-import graphql.language.FragmentDefinition;
-import graphql.language.NodeUtil;
-import graphql.language.OperationDefinition;
+import graphql.language.*;
```
artem-ag, line 6: Revert a change for wildcard import?

austinarbor: done

skevy:
Sorry about the stall. Thank you for picking it  up and fixing it!

PR #2184, opened 2021-02-04
fix #2131: less allocations in SimpleInstrumentation

### LINKED ISSUE #2131: Small performance optimization: SimpleInstrumentation should use SimpleInstrumentationContext.noOp()
Hi,
The SimpleInstrumentation is creating a lot of useless SimpleInstrumentationContext objects. I suggest using the already existing SimpleInstrumentationContext.noOp() instead to save some memory allocations and GC overhead.

dfa1:
Since SimpleInstrumentation is often used as base class, it would be
possible to skip some allocations in case the method is not overridden.


bbakerman (approved):
Thanks

Issue #1896, opened 2020-05-15
MaxQueryComplexityInstrumentation leaks exceptions when input fields are null


alexgenco:
**Describe the bug**

See comments in reproduction code, kinda hard to describe outright.

**To Reproduce**

graphql-java version: ~13.0~ 14.0
graphql-java-tools version: ~5.6.1~ 6.0.2 (just for the schema builder helpers)

```java
public class ReproComplexityBug {
  @Test
  public void repro() {
    GraphQLSchema schema = new SchemaParserBuilder()
      .schemaString("type Query {\n" +
                    "  echo(input: EchoInput!): String!\n" +
                    "}\n" +
                    "input EchoInput {\n" +
                    "  msg: String!\n" +
                    "}")
      .resolvers(new MyResolver())
      .build()
      .makeExecutableSchema();

    Instrumentation maxComplexity = new MaxQueryComplexityInstrumentation(10);

    GraphQL graphql = new GraphQL.Builder(schema)
      .instrumentation(maxComplexity) // If you get rid of this line, the test passes
      .build();

    // This line throws `NonNullableValueCoercedAsNullException`, instead of returning an `ExecutionResult` with errors.
    ExecutionResult result =
      graphql.execute("query Q($foo: String!) { echo(input: {msg: $foo}) }");

    // Executing the following instead causes the test to pass. Only difference is top-level input is null vs. nested input field is null.
    // ExecutionResult result = graphql.execute("query Q($input: EchoInput!) { echo(input: $input) }");

    assertEquals(1, result.getErrors().size());
    assertTrue(result.getErrors().get(0).getMessage().contains("NonNull"));
  }

  private static class MyResolver implements GraphQLQueryResolver {
    public String echo(EchoInput input) {
      return input.msg;
    }
  }

  private static class EchoInput {
    String msg;
  }
}
```

alexgenco:
Update: just tried on 14.0 and encountered the same error.

alexgenco:
Any thoughts on this?

lkorth:
This is now fixed and can be closed https://github.com/graphql-java/graphql-java/pull/2234#issuecomment-1128395971

PR #2234, opened 2021-03-02
Coercion exception catching tests for MaxComplexityInstrumentation

zkry:
This PR includes two tests which reproduce the issue described in issue #1896 with the MaxQueryComplexityInstrumentation. 

Between versions 14.1 and 16.2 it seems and additional regression was introduced which caused this for non-nested inputs as well. Between these two versions it also started to become a problem for scalar coercion as well.


dondonz:
Hi, this is fixed in the current version, closing this PR

lkorth:
@dondonz thank you for the update, I verified the fix with this test and a few others we had.

PR #2252, opened 2021-03-19
Build: Export all graphql packages in OSGi bundle.

### LINKED ISSUE #2251: Version 16.0 and later is no OSGi bundle anymore
**Describe the bug**
Since version 16.0 the graphql-java.jar is no valid OSGi bundle anymore.
The MANIFEST.MF file does not contain the required entries anymore.

Version 15.0 was a valid bundle and was running fine in an OSGi runtime.

**To Reproduce**
Try to load graphql-java in an OSGi runtime, like Eclipse or Apache Felix.

**Expected** 
Valid OSGi bundle similar to version 15.0. With valid import/export. 
Version 15 appear to have used Bnd to generate the correct MANIFEST.MF entries.

AndreasTu:
Fixes #2251

This also removes the Guava imports, because it is relocated inside the bundle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,6 @@ jar {
     manifest {
         attributes('Automatic-Module-Name': 'com.graphql-java')
     }
-	bnd('-exportcontents': 'graphql.*')
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


AndreasTu:
@andimarek I have changed this PullRequest to exclude Guava from the OSGi MANIFEST.MF.

AndreasTu:
@andimarek Can I help with something or shall I change something to get this merged?
I would need this fix, that a can migrate to the new version, because I am using the library in an OSGi environment.

andimarek:
Can you add a comment to explain in more detail what this config does and why? This would help us to maintain that better in the future. Thanks

AndreasTu:
@andimarek I have added comments to explain what the bnd configuration does.
BTW: Thanks for the great library!

bbakerman (approved):

PR #2253, opened 2021-03-19
OSGi: Allow Guava Library in a wider range from 30.0 up until excl. 32.

AndreasTu:
This allows the usage of the currently newest Guava version 31.1 in an OSGi environment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,7 +59,10 @@ jar {
     manifest {
         attributes('Automatic-Module-Name': 'com.graphql-java')
     }
-	bnd('-exportcontents': 'graphql.*')
+	bnd('''
+-exportcontents: graphql.*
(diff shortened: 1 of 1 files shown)


andimarek:
Google guava is not imported from another Module: we actually rewrite all imports and remove the dependency in the POM https://github.com/graphql-java/graphql-java/blob/master/build.gradle#L203.

So from my understanding this means it is a purely implementation detail and OSG should not be concerned with it at all.

AndreasTu:
So this means the shadowJar is released instead of the jar? 
This may also explain why the released jar is no OSGi bundle anymore:  Issue #2251
In this case this PR does not make sense. And I will also change PR #2252, which does not fix the OSGi problem.

andimarek:
@AndreasTu yes we release only a shadow jar

PR #2323, opened 2021-05-07
Updated antlr to version 4.9.2

kai-baumann:
Updated the antlr runtime and the antlr generator to version 4.9.2


andimarek:
Thanks for that @kai-baumann

PR #2329, opened 2021-05-11
Add a failing test to show NPE when calculating query complexity

srinivasankavitha:
This occurs for queries that take in a list input as a variable with default value.


andimarek:
This will be fixed in 17.0

Issue #2336, opened 2021-05-13
Inline fragments don't work on fields declared with interface types (v16.2)


aleksandarsusnjar:
**Description**
Specifying an inline fragment for a concrete (sub)type of a field declared with that same type seems to work. However, field fetchers aren't even called when the same is done for a field declared using an interface type, whether that declared interface type, concrete object type or any other type in between is used.

For more details see:
https://github.com/graphql-java/graphql-java/discussions/2331

**To Reproduce**
Create a schema as follows:
```GraphQL
interface Base {
    id: ID
}

type Concrete implements Base {
    id: ID
    name: String
}

type Test {
    base: Base
    concrete: Concrete
}

type Query {
    test: Test
}   
```

Set it up so that you can return/fetch some data. Perhaps return the exact same object/instance for both `base` and `concrete` fields of `Test` to make it more interesting (both need to be of type `Concrete` anyway).  Issue the following query:

```GraphQL
query {
    test {
        base {
            __typename
            ... on Base {
                id
            }
            ... on Concrete {
                name
            }
        }
        concrete {
            __typename
            ... on Concrete {
                id
                name
            }
        }
    }
}
```

Fields `base` and `concrete` should include exactly the same sub-fields: `__typename`, `id` and `name`. They do not.

aleksandarsusnjar:
I've been tracing this... in `ExecutionStrategy`, method:

```java
protected CompletableFuture<ExecutionResult> completeValueForObject(ExecutionContext executionContext, ExecutionStrategyParameters parameters, GraphQLObjectType resolvedObjectType, Object result) {
```
... line:
```java
MergedSelectionSet subFields = fieldCollector.collectFields(collectorParameters, parameters.getField());
```
yields `subFields` without field from inline fragments. `collectorParameters` has:

1. `graphQLSchema`:  non-null 
2. `fragmentsByName`: empty, size = 0
3. `variables`: empty, size = 0
4. `objectType`: correct object type and its `fieldDefinitionsByName` includes all the fields.

I can see that `fieldCollector` has a non-null `conditionalNodes` reference which has a non-null `valuesResolver` reference.

aleksandarsusnjar:
SelectionSet for the field in question (for which subFields are being determined) prints as:
```
[...]

PR #2339, opened 2021-05-14
Fix for issue 2336 using alternate type equality via isEquivalentTo method (master branch version)

aleksandarsusnjar:
See https://github.com/graphql-java/graphql-java/issues/2336


aleksandarsusnjar (commented):
Something is off... This should have been "clean" but isn't It has too many changes, not mine.

PR #2343, opened 2021-05-14
Fix for issue 2336 using alternate type equality via isEquivalentTo method (master version)

aleksandarsusnjar:
See #2336 and #2331


aleksandarsusnjar:
See #2336. Not needed.

PR #2364, opened 2021-05-30
add automated performance tests

andimarek:
This PR adds automated performance tests.

Every JMH test is executed on a fresh AWS EC2 instance and then the results are compared to previous results to make sure GraphQL Java didn't get slower.

If it got slower, the build will fail.


dondonz:
Closing this PR as the code here has been recycled into a new JMH pipeline.

PR #2376, opened 2021-06-06
Anonymizer default values fixes

tinnou:
### Description
This PR revises anonymizer's handling of default values. Specifically it adds support for correctly anonymizing default values of type enum, object, and list in both schema and queries.

This PR also fixes a bug where the anonymizer would produces a schema that fails GraphQL spec schema validation because fields on implementing types would have different anonymized default values as the default values on the fields of the implemented interface. 

Example:
```
type Human implements Character {
  id: String
  friends(limit: Int = 100): [Character]
}

interface Character {
  id: String
  friends(limit: Int = 100): [Character]
}
```

would anonymize to:
```
type Object1 implements Interface1{
  field1: String
  field2(arg1: Int = 1): [Interface1]
}

interface Interface1 {
  field1: String
  field2(arg1: Int = 2): [Interface1]
}
```

and above would fail validation since `Object1.field2` definition is different from `Interface1.field2` definition.

### DIFF
--- a/src/main/java/graphql/util/Anonymizer.java
+++ b/src/main/java/graphql/util/Anonymizer.java
@@ -15,19 +15,26 @@ import graphql.analysis.QueryVisitorInlineFragmentEnvironment;
 import graphql.execution.ValuesResolver;
 import graphql.introspection.Introspection;
 import graphql.language.Argument;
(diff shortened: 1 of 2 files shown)


andimarek:
thanks a lot @tinnou

PR #2392, opened 2021-06-21
Fix absent and null variable values

gnawf:


src/test/groovy/graphql/normalized/NormalizedQueryFactoryTest.groovy
```diff
@@ -1103,7 +1103,93 @@ schema {
...
 
+        then:
+        def topLevelField = tree.getTopLevelFields().get(0)
+        def arg = topLevelField.getNormalizedArgument("arg")
+        def otherArg = topLevelField.getNormalizedArgument("otherArg")
+
+        arg == new NormalizedInputValue(
+                "Arg",
+                [
+                        ids: new NormalizedInputValue(
```
andimarek, line 1141: they should not be `null`: a missing var value is different from a `null` one.  It should be an empty map.

andimarek (approved):

PR #2417, opened 2021-07-02
Support mutations & subscriptions in NQ -> AST compiler

gnawf:


temaEmelyan (approved):

src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -11,36 +12,62 @@
...
-        Document document = Document.newDocument().definition(newOperationDefinition()
-                .operation(QUERY)
-                .selectionSet(selectionSet)
-                .build())
-                .build();
-        return document;
+
+        Map<OperationDefinition.Operation, ImmutableList<Selection<?>>> selectionsByOperationKind = FpKit.groupingBy(selections, (selection) -> {
+            if (!(selection instanceof InlineFragment)) {
+                throw Assert.<RuntimeException>assertShouldNeverHappen("Top level selection must be an inline fragment");
```
andimarek, line 45: this looks like a non needed type Qualifer <RuntimeException>

src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -11,36 +12,62 @@
...
-import static graphql.language.OperationDefinition.newOperationDefinition;
+import static graphql.language.OperationDefinition.Operation.SUBSCRIPTION;
 import static graphql.language.SelectionSet.newSelectionSet;
 import static graphql.language.TypeName.newTypeName;
 
 @Internal
 public class ExecutableNormalizedOperationToAstCompiler {
-
-    public static Document compileToDocument(List<ExecutableNormalizedField> topLevelFields) {
+    public static Document compileToDocument(GraphQLSchema schema, List<ExecutableNormalizedField> topLevelFields) {
```
andimarek, line 40: We should not need a schema to print a NormalizedOperation. I think we need to add the operation to the `ExecutableNormalizedOperation` itself or pass it here into the function.

andimarek (changes requested):
I think this is not the right approach as it now requires a Schema object.

andimarek (approved):

PR #2558, opened 2021-09-21
Add InstrumentationValidationParameters to QueryComplexityInfo

stevenleeDB:
Addressing #2547


bbakerman (approved):
Thanks for this PR.

I had a concern that the raw parameters would to "wide" but given that `MaxQueryComplexityInstrumentation` is an instrumentation anyway, giving access to the validtion parameters is ok

stevenleeDB:
@bbakerman thanks! Would you be able to merge this for me?

softprops:
any word on when the next release will be that would include these changes?

bbakerman:
Every merge to master results in a timestamped release of graphql-java in
Maven central

So you can update if you need to.

As for the release, we havent convened to decide when - Typically its every
6 months or so but...that is not set in stone

Issue #2560, opened 2021-09-24
When sending an invalid enum value the error is not as expected


haneev:
**Describe the bug**
When an invalid enum-value is send-in the error results in a weird chain of invalid value. Like this:

```
{
  "errors" : [ {
    "message" : "invalid value : invalid value : invalid value : Invalid input for Enum 'PositionType'. No value found for name 'UNKNOWN_POSITION'",
    "extensions" : {
      "classification" : "ValidationError"
    }
  } ]
}```

**To Reproduce**

With the following schema:

```
enum PositionType {
MANAGER
DEVELOPER
}
input PersonInput {
name: String
position: PositionType
}

type Mutation {
  updatePerson(input: PersonInput!): Boolean
}
```

and request
```
mutation Mutation($input: PersonInput!) {
    updatePerson(data: $input)
}

vars
{
 "input": {
"name": "Name",
"position": "UNKNOWN_POSITION"
}
}
```

This is caused by the change related to default-values and a change in the ValuesResolver.java
Each type in the tree will add a `invalid value : ` to the error message. So if you have a depth of 3 it will be three invalid values.

Another issue is that you don't know which field is actually invalid. The previous version of the code had the name of the field in the error message, which was way more convenient. 

Previous message was: `Variable 'position' has an invalid value: ...`

I also encounter this for other types as well, same for sending "maybe" to a Boolean type. So probably other types will have the same issue as well?

**Comment by dondonz** [comment]
Thanks @haneev! The fix has been merged in. 

The error message will now appear as
```
Variable 'input' has an invalid value: Invalid input for Enum 'PositionType'. No value found for name 'UNKNOWN_POSITION''
```

juriad:
I just found this issue during our upgrade from 15.0 to 17.3. Some of our tests started failing during the deserialization of a bad UUID (0000-0000) for a custom non-null scalar variable.

Thanks for the fix, it looks good. How likely is it that there is going to be release 17.4?

PR #2570, opened 2021-10-03
WIP: Fix triple nested enum validation error

dondonz:
WIP, please come back later!

Working on a fix for incorrect enum validation message, as reported in #2560.

### DIFF
--- a/src/test/groovy/graphql/Issue2560.groovy
+++ b/src/test/groovy/graphql/Issue2560.groovy
@@ -1,0 +1,49 @@
+package graphql
+
+import spock.lang.Specification
+
+class Issue2560 extends Specification {
+    def "invalid enum error message is nested and source location missing"() {
+        when:
+        def mutation = '''
+            mutation UpdatePerson($input: PersonInput!) {
+                updatePerson(input: $input)
+            }
+        '''
+
+        def graphQL = TestUtil.graphQL('''
+            enum PositionType {
+                MANAGER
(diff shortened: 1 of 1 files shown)


src/main/java/graphql/execution/ValuesResolver.java
```diff
@@ -388,30 +388,38 @@ private Object externalValueToLiteralForObject(GraphqlFieldVisibility fieldVisib
...
-                coercedValues.put(variableName, coercedDefaultValue);
-            } else if (isNonNull(variableType) && (!hasValue || value == null)) {
-                throw new NonNullableValueCoercedAsNullException(variableDefinition, variableType);
-            } else if (hasValue) {
-                if (value == null) {
-                    coercedValues.put(variableName, null);
-                } else {
-                    Object coercedValue = externalValueToInternalValue(fieldVisibility, variableType, value);
-                    coercedValues.put(variableName, coercedValue);
+            try {
```
dondonz, line 395: The fundamental change is to move the `CoercingParseValueException` catch one function earlier in the chain, to `externalValueToInternalValueForVariables`.

Although it appears many lines are changing, it's mostly only indentation for the try catch block.

dondonz: The rationale for moving the catch: the source location is available here, whereas it is not passed down to the next function `externalValueToInternalValue`.

Issue #2524, opened 2021-08-27
RuntimeException in ValuesResolver thrown without any further information


belks:
https://github.com/graphql-java/graphql-java/blob/27b11d99d4885e81b0047a58ea88dd99b5a5b40f/src/main/java/graphql/execution/ValuesResolver.java#L455

This is a nasty code smell and makes debugging very painful. Please throw a meaningful exception here with a detailed error message.

PR #2572, opened 2021-10-04
Improve non-nullable argument exception in ValuesResolver

dondonz:
Addresses #2524.

Previously, an non-nullable argument with a coerced null value raised a generic `RuntimeException`. This PR updates the exception to be more meaningful.


bbakerman (approved):
Thanks

PR #2585, opened 2021-10-15
Allow QueryTraverser and QueryTransformer to complete without required variables

schenkman:
Hello from Twitter!

### What
This PR adds a new method to `QueryTraverser.Builder`, `allowMissingVariables(boolean allow)` to relax the requirement during traversal that all required variables are present for the traversal to complete. The implication is this state variable needs to eventually be passed down into the execution package `ValuesResolver` class. The intent of this change is to simply relax the requirement for all required variables be present during traversal, and that's it. If there's a better way to implement this functionality, I'm open to any suggestions.

### Why
Our workflow includes other teams uploading queries as strings to be stored in a database. At the time the query is uploaded, we perform static analysis on the query to determine certain statistics of execution based on some directives we add to various fields in the schema. This means that at the time the query document string is uploaded, we do not have the required non-null variables for the query.

Digging into the `QueryTraverser`, we noticed that the `NodeVisitorWithTypeTracking` creates its own `ValuesResolver` instance which expects a fully validated query (otherwise we get the `NonNullableValueCoercedAsNullException`), which isn't necessarily what we care about when we're traversing for static analysis.

Additionally, `NodeVisitorWithTypeTracking` is also used by `QueryTransformer`, and this functionality also seems to make sense to enable there, if you want to transform a query without necessarily having all the variables. In this case, `QueryTransformer` will always traverse into all parts of the query regardless of variables and conditions.

If you have thoughts on a better way of doing this kind of analysis, please let us know. Thanks!

### DIFF
--- a/src/main/java/graphql/analysis/NodeVisitorWithTypeTracking.java
+++ b/src/main/java/graphql/analysis/NodeVisitorWithTypeTracking.java
@@ -61,6 +61,15 @@ public class NodeVisitorWithTypeTracking extends NodeVisitorStub {
         this.fragmentsByName = fragmentsByName;
     }
 
+    public NodeVisitorWithTypeTracking(QueryVisitor preOrderCallback, QueryVisitor postOrderCallback, Map<String, Object> variables, GraphQLSchema schema, Map<String, FragmentDefinition> fragmentsByName, boolean allowMissingVariables) {
+        this.preOrderCallback = preOrderCallback;
+        this.postOrderCallback = postOrderCallback;
+        this.variables = variables;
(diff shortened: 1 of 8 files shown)


schenkman:
Hey @andimarek and @bbakerman I discussed with @jbellenger that we can cancel this PR as we'd prefer to work tooling code into the newer `normalized` package that you wrote about in the blog post. We may still need to discuss how to manage the state to allow missing variables for relaxed static analysis, but we can revisit it once we get back to looking at this change. It's lower priority on our end for now, but it is on our radar.

PR #2621, opened 2021-11-29
Avoid printing inline fragment if there is no conditional type

gnawf:
My IDE keeps freezing when I open GraphQL Java, so I used the IDE without the smart features to write this code… felt like I was sent 10 years back in time. Let me know if something is wrong.

### DIFF
--- a/src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
+++ b/src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
@@ -2,6 +2,7 @@ package graphql.normalized;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import graphql.Assert;
 import graphql.Internal;
 import graphql.language.Argument;
 import graphql.language.ArrayValue;
@@ -16,6 +17,8 @@ import graphql.language.Selection;
 import graphql.language.SelectionSet;
 import graphql.language.TypeName;
 import graphql.language.Value;
+import graphql.schema.GraphQLSchema;
+import graphql.util.FpKit;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -26,71 +29,81 @@
...
 @Internal
 public class ExecutableNormalizedOperationToAstCompiler {
-    public static Document compileToDocument(
-            OperationDefinition.Operation operationKind,
-            String operationName,
-            List<ExecutableNormalizedField> topLevelFields
-    ) {
-        List<Selection<?>> selections = selectionsForNormalizedFields(topLevelFields);
+    public static Document compileToDocument(GraphQLSchema schema,
+                                             OperationDefinition.Operation operationKind,
```
gnawf, line 41: Ok now that we have the `schema` again, we can avoid explicitly specifying the `operationKind` and derive it from the `ENF` and `GraphQLSchema` but as I said in the PR desc, my IDE is broken, and that approach required some refactoring which I wasn't prepared to do without the IDE. What a massive pain…

src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -26,71 +29,81 @@
...
-                List<Field> fieldList = overallGroupedFields.computeIfAbsent(objectTypeName, ignored -> new ArrayList<>());
-                fieldList.addAll(fields);
-            });
+        // All conditional fields go here instead of directly to selections so they can be grouped together
+        // in the same inline fragement in the output
+        Map<String, List<Field>> conditionalFieldsByObjectTypeName = new LinkedHashMap<>();
 
+        for (ExecutableNormalizedField nf : executableNormalizedFields) {
+            Map<String, List<Field>> groupFieldsForChild = selectionForNormalizedField(schema, nf);
+            if (nf.isConditional(schema)) {
```
gnawf, line 66: This is the change, we check if the NF is conditional, if it is then we add it to the conditional field `Map` to group the type conditions, otherwise we add it straight to the output as a raw `Field`.

src/main/java/graphql/normalized/ExecutableNormalizedOperationToAstCompiler.java
```diff
@@ -26,71 +29,81 @@
...
+                    fieldList.addAll(fields);
+                });
+            } else {
+                List<Field> fields = groupFieldsForChild.values().iterator().next();
+                selections.addAll(fields);
+            }
         }
 
-        overallGroupedFields.forEach((objectTypeName, fields) -> {
+        conditionalFieldsByObjectTypeName.forEach((objectTypeName, fields) -> {
```
gnawf, line 77: As usual, we group the condition fields by object type conditions so that they can be grouped together in the same inline fragment.

andimarek (approved):

Issue #2649, opened 2021-12-18
ANTLR grammar doesn't match GraphQL spec for directive locations


sachindshinde:
**Describe the bug**

Looking at the [2018 GraphQL spec](https://spec.graphql.org/June2018/#sec-Type-System.Directives), it has for a while allowed for an optional leading `|` in directive locations. E.g., the following is valid GraphQL:
```graphql
directive @foo on
    | OBJECT
    | INTERFACE
```

However, trying to parse this with `graphql-java` throws, because the [ANTLR grammar](https://github.com/graphql-java/graphql-java/blob/master/src/main/antlr/GraphqlSDL.g4#L128) doesn't allow the optional leading `|`.

**To Reproduce**

This code snippet throws:
```java
new SchemaParser().parse("directive @foo on | OBJECT");
```

jord1e:
`graphql-js` does not seem to have test cases describing this. It does have them for unions, weird

PR #2650, opened 2021-12-19
Allow leading pipe in directive SDL, and add leading union pipe tests

jord1e:
Fixes graphql-java/graphql-java#2649

Represented as |opt in the GraphQL specification

Link to union tests in graphql-js: https://github.com/graphql/graphql-js/blob/3ab82f4981970f50dd3a96e22a11839e4ec5bb7b/src/language/__tests__/schema-parser-test.ts#L892-L962

Spec blame: https://github.com/graphql/graphql-spec/blame/c18590c477d6817d73e19b6c248c9fcc62b03017/spec/Appendix%20B%20--%20Grammar%20Summary.md#L327

graphql-js does not seem to provide directive tests for this specific circumstance (I should really open a PR over there)

### DIFF
--- a/src/main/antlr/GraphqlSDL.g4
+++ b/src/main/antlr/GraphqlSDL.g4
@@ -126,6 +126,6 @@ directiveDefinition : description? DIRECTIVE '@' name argumentsDefinition? REPEA
 directiveLocation : name;
 
 directiveLocations :
-directiveLocation |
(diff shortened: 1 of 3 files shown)


bbakerman (approved):
Thanks very much for this PR.

andimarek (approved):

PR #2672, opened 2022-01-11
Fix AstPrinter trimming blank string values

gnawf:
Let me know if we should add any other tests.


src/main/java/graphql/language/AstPrinter.java
```diff
@@ -482,7 +482,7 @@ private String value(Value value) {
         } else if (value instanceof FloatValue) {
             return valueOf(((FloatValue) value).getValue());
         } else if (value instanceof StringValue) {
-            return wrap("\"", escapeJsonString(((StringValue) value).getValue()), "\"");
+            return "\"" + escapeJsonString(((StringValue) value).getValue()) + "\"";
```
gnawf, line 485: Decided to just switch to plain old concat instead of using the `wrap` method.

bbakerman (approved):
Nice

PR #2705, opened 2022-01-31
Fix wrong newline on Windows in `SchemaPrinter` (`\r\n -> \n`)

jord1e:
The printer was printing `\r\n` sequences instead of `\n`, because it was using `PrintWriter#println`

Fixes the `"can print a schema as AST elements"` test in `SchemaPrinterTest.groovy` on Windows

See
https://docs.oracle.com/javase/8/docs/api/java/io/PrintWriter.html#println--

> The line separator string is defined by the system property line.separator, **and is not necessarily a single newline character ('\n')**.

### DIFF
--- a/src/main/java/graphql/schema/idl/SchemaPrinter.java
+++ b/src/main/java/graphql/schema/idl/SchemaPrinter.java
@@ -636,7 +636,7 @@ public class SchemaPrinter {
                 out.printf("\n%s\n", AstPrinter.printAst(extension));
             }
         }
-        out.println();
+        out.print('\n');
     }
 
     private static String printAst(InputValueWithState value, GraphQLInputType type) {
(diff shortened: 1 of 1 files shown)


bbakerman:
But why is `\r\n` wrong in Windows terms - isnt that what Windows programs want / handle?

jord1e:
Yes, for text programs, but the test "can print a schema as AST elements"
https://github.com/graphql-java/graphql-java/blob/4a82b0bbd1483d148f3abf575fe705be6d2b0f70/src/test/groovy/graphql/schema/idl/SchemaPrinterTest.groovy#L1264
fails because Groovy multiline strings use `\n`

You can see this when debugging and extracting the string to be compared to a variable
afbeelding
`result` is before the patch and contains `\r` at 9 places (coincidentally there are 9 `extend` types, this is how I found the bug)
`expected` is how the test is compiled by groovy (without `\r`'s)

Notepad and other software correctly recognize `\n` line feeds
afbeelding

afbeelding

This behaviour is *only* present when using `extend` types. The regular SchemaPrinter and AstPrinter use `\n` everywhere, which is consistent across all operating systems

jord1e:
Now that I am reworking the CI with Windows support you can view the error:

https://github.com/jord1e/graphql-java/actions/runs/1773061590
`gradle-test-summary-b8-t8-temurin-windows-latest`, and open `index.html`

bbakerman:
thanks for the detailed explanation

PR #2788, opened 2022-04-09
Add ParserOption to ignore single-line comments

jord1e:
This commit adds the ability to ignore single-line comments during AST conversion.

Fixes graphql-java/graphql-java#2767

Always glad to help ❤️


bbakerman (approved):
Thanks you very much for this.  At some point I think we would flip the default to be "dont capture comments" but having it on for now is a sensible starting place

PR #2793, opened 2022-04-12
Use method-based reflection on JDK 16+ records instead of field-based reflection

jord1e:
See title.

Records were introduced in JDK 16 by [JEP 395](https://openjdk.java.net/jeps/395). When the field (constructor parameter) is called `a`, the method will be called `a()` - even when it is a boolean.

I tested it manually. Adding a JDK 16+ testing target will be a little bit harder.

Migrating to [Gradle JVM Test Suites](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html) might help in the future.

Kotlin can also compile `data class`es to records by adding a special `@JvmRecord` annotation: https://kotlinlang.org/docs/jvm-records.html#declare-records-in-kotlin=

### DIFF
--- a/src/main/java/graphql/schema/PropertyFetchingImpl.java
+++ b/src/main/java/graphql/schema/PropertyFetchingImpl.java
@@ -125,6 +125,13 @@ public class PropertyFetchingImpl {
     }
 
     private Object getPropertyViaGetterMethod(Object object, String propertyName, GraphQLType graphQLType, MethodFinder methodFinder, Object singleArgumentValue) throws NoSuchMethodException {
+        if (isRecord(object.getClass())) {
+            try {
+                return getPropertyViaGetterWithoutPrefix(object, propertyName, methodFinder, singleArgumentValue);
+            } catch (NoSuchMethodException ignored) {
+                // Records can still have getX and isX methods...
+            }
+        }
         if (isBooleanProperty(graphQLType)) {
             try {
                 return getPropertyViaGetterUsingPrefix(object, propertyName, "is", methodFinder, singleArgumentValue);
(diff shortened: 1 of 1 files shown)


bbakerman:
Do you have any insights into whether this is faster?

We cache the `Method`s we find via reflection and hence it might not really end up in significantly faster calls.  But maybe I am wrong

In general I like the idea - testing it is of course problematic - but it might be worth having even without a specific test in the code base.

jord1e:
Method access seems to be a little bit faster (and with less variance, so a more constant performance profile):

This probably has to do something with the fact that record fields are compiled to private fields, and thus access checks take more time (even when setAccessible is set).

```
Benchmark                                                       Mode  Cnt        Score        Error  Units
RecordPropertyFetcherBenchMark.benchMarkRecordViaFieldAccess   thrpt   15  6030381,447 ± 688112,314  ops/s
RecordPropertyFetcherBenchMark.benchMarkRecordViaMethodAccess  thrpt   15  6329489,223 ± 310016,693  ops/s
```

afbeelding

Benchmarks: https://github.com/jord1e/graphql-java/pull/2/files

The JEP also makes clear that the default generated 'getters' (e.g. `a()`) can also be overriden with custom behaviour, as they are not marked as final:
> For each component in the header, two members: a ***`public` accessor method*** with the same name and return type as the component, and a `private final` field with the same type as the component;

In the current situation possible side-effects (e.g. logging) will not be called for methods overriding generated record getters, whilst they will be recorded when the record manually has a `getX` method. (not that I am advocating for side-effect riddled code in records)

I think this is a good change

jord1e:
Compilers can also add custom behaviour, the Kotlin compiler adds non-null checks and @NotNull annotations to `@JvmRecord data class`'es methods:

afbeelding

A custom data fetcher implementation may hypothetically check for the presence of these annotations.

andimarek:
Thanks a lot @jord1e , but we decided not to adopt it.

If you look at the JMH output and take the error range into considerations it is not clear that this faster.

Additionally we are concerned with the introduction of JVM version specific code. I know you have another PR open which might help us to get there in the future, but would need to make sure that we test against Java 8 (at a minimum) and 16 with every build. This is a big step up in terms of maintenance and commitment for us, which only want to take if really needed.

Thanks a lot!

PR #2798, opened 2022-04-16
Fix instances of incorrectly sized Maps and Sets

kilink:
Fix a few occurrences where a Map or Set was initialized with an initial capacity
value N, and then immediately filled with N items, which would always trigger
a resize.


bbakerman (approved):
Thanks

PR #2803, opened 2022-04-18
Make some static fields final

kilink:
Make some static fields final, and avoid unnecessary boxing in UnicodeUtils.


bbakerman (approved):
Thanks

PR #2818, opened 2022-05-11
Fix silent thread leak for chained instrumentation

KammererTob:
This PR is related to another PR i did some time ago, which tried to address a very specific case of thread leakage. 
I am now revisiting this issue since i noticed that the merged PR did not apply it's fix to chained instrumentation, which this PR tries to remedy.

Issue: https://github.com/graphql-java/graphql-java/issues/2068
PR: https://github.com/graphql-java/graphql-java/pull/2359


bbakerman (approved):
Thanks

PR #2846, opened 2022-06-04
Add subscription root validation

dondonz:
Adding validation to enforce single root fields in subscriptions & no introspection root https://spec.graphql.org/draft/#sec-Subscription-Operation-Definitions


bbakerman (approved):
Nice

Issue #2809, opened 2022-04-22
antlr4 v4.10.x is out - any plans to use / adopt to it?


aamotharald:
Are there any pland to move to antlr 4.10.x ?
This implies re-generating the scanners, lexers and parsers based on a grammar. 4.9.x (compile-time) and 4.10.x (runtime) does not work.
It's just a technicality to upgrade, but it has to be done in one step.

andimarek:
We don't have specific plans, but we are open for a pull request if you would like to contribute.

Is there a specific reason that you ask? 

Thanks!

aamotharald:
Hi @andimarek ,
there is no specific reason. My project uses the renovate bot that informed us about the new antlr4 version with a PR.
We noticed that we can only move to this version when all other (reuse)libraries that use antlr4 are also on that version. So graphql-java is one of these. It's a bigger coordination effort for us to move to antlr4 v 4.9.x and would have rather expected this to be version 5.x.x . In other words: if I start a contribution for graphql-java a major version bump should be considered to not cause runtime issues by your consumers. I am still not sure how I should deal with this, but I am happy to make a contribution there if the question about the semantiv versioning for graphql-java would be discussed out..

dondonz:
Thanks for raising the issue! We're now shading `antlr-runtime`, which will resolve this version clash problem.

PR #2852, opened 2022-06-11
Upgrade ANTLR to 4.10.1 WIP

dondonz:
Update to the latest ANTLR version and unblock others relying on this dependency update #2809 and https://github.com/openrewrite/rewrite/issues/1615

The key change with this update is that the lexers/parsers etc needed to be regenerated in order to be compatible with the new ANTLR version. 

In case you're wondering why those files don't appear in the diff - we regenerate these files on every build with the `generateGrammarSource` task, we do not check in these files into the repo.


dondonz:
Closing this PR and I'll shortly open a new one where `antlr-runtime` is shaded in, which will also solve the dependency issues discussed

PR #2789, opened 2022-04-09
Continuous integration improvements

jord1e:
Fixes #2676

Does the following:
1. Updates GitHub Actions dependencies
2. Run tests on HotSpot, OpenJ9 VM,
3. Run tests on Ubuntu and Windows
4. Run tests on Java 8 (latest), 8.0.282 (earliest supported), 11 (LTS), 17 (LTS) and 18 (latest)
5. Compile with Java 8 and test on Java 17 (HotSpot Ubuntu)
6. Remove Java 8 requirement for `sourceCompatibility` and `targetCompatibility` (now resides in gradle.properties and just takes whatever J8 you have installed, or installs the latest J8)

Does not touch deployment of the jars on master, I found it kind of unnecessary because everything is done in PRs anyway and it complicates the build scripts.

This needs to be squashed before merging.

PR #2854, opened 2022-06-12
Shade antlr-runtime

dondonz:
This PR shades `antlr-runtime` and removes it off the pom.

`antlr-runtime` version clashes can be a blocker as described in https://github.com/graphql-java/graphql-java/issues/2809 and https://github.com/openrewrite/rewrite/issues/1615

### DIFF
--- a/src/main/java/graphql/execution/ExecutionContext.java
+++ b/src/main/java/graphql/execution/ExecutionContext.java
@@ -116,6 +116,8 @@ public class ExecutionContext {
     }
 
     /**
+     * @return map of coerced variables
+     *
      * @deprecated use {@link #getCoercedVariables()} instead
      */
     @Deprecated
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/execution/ExecutionContext.java
```diff
@@ -116,6 +116,8 @@ public OperationDefinition getOperationDefinition() {
     }
 
     /**
+     * @return map of coerced variables
```
dondonz, line 119: Fixing up javadoc warnings, unrelated to antlr

jord1e:
Just noting that #2789 will allow running ANTLR 4.10.x generation with JDK11 whilst preserving JDK8 source compatibilty
I will merge the current featureset into #2789 again today.

PR #2871, opened 2022-06-23
create jmh jar

adarsh-jaiswal:
In order ro run benchmark tests by the test runner we need the jmh jar. To create the jmh jar I have added the jmh plugin in this PR. Also, added exception to be thrown by the benchmark methods.

### DIFF
--- a/src/test/java/benchmark/AddError.java
+++ b/src/test/java/benchmark/AddError.java
@@ -28,7 +28,7 @@ public class AddError {
     @BenchmarkMode(Mode.SingleShotTime)
     @Warmup(iterations = 1, batchSize = 50000)
     @Measurement(iterations = 1, batchSize = 5000)
-    public ExecutionContext benchMarkAddError() {
+    public ExecutionContext benchMarkAddError() throws InterruptedException {
         context.addError(
                 new SchemaMissingError(),
                 ResultPath.fromList(Collections.singletonList(x++))
--- a/src/test/java/benchmark/AddError.java
+++ b/src/test/java/benchmark/AddError.java
@@ -28,7 +28,7 @@ public class AddError {
     @BenchmarkMode(Mode.SingleShotTime)
     @Warmup(iterations = 1, batchSize = 50000)
     @Measurement(iterations = 1, batchSize = 5000)
-    public ExecutionContext benchMarkAddError() throws InterruptedException {
+    public ExecutionContext benchMarkAddError() {
         context.addError(
                 new SchemaMissingError(),
                 ResultPath.fromList(Collections.singletonList(x++))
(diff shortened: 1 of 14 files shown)


src/test/java/benchmark/AddError.java
```diff
@@ -28,7 +28,7 @@ public class AddError {
     @BenchmarkMode(Mode.SingleShotTime)
     @Warmup(iterations = 1, batchSize = 50000)
     @Measurement(iterations = 1, batchSize = 5000)
-    public ExecutionContext benchMarkAddError() {
+    public ExecutionContext benchMarkAddError() throws InterruptedException {
```
chriswr95, line 31: why do we need to add these exceptions to every method signature?

chriswr95:
what is meaningfully different after this PR? Can I now run some command that I couldn't previously?

src/test/java/benchmark/AddError.java
adarsh-jaiswal: The generated jmh test class uses these methods and it is wrapped inside a try statement which catches InterruptedException. When I build the jmh test jar it throws compilation error. In order to fix it I had to add it to all the methods.

adarsh-jaiswal:
> what is meaningfully different after this PR? Can I now run some command that I couldn't previously?

we can now run the bollowing commands -

1.  to build jmh jar - `RELEASE_VERSION=test-runner-jmh ./gradlew jmhJar`
2.  to run test and produce test result in json - `java -jar build/libs/graphql-java-test-runner-jmh.jar -rf json`

andimarek:
I needed to a bit more config last time I did this: https://github.com/graphql-java/graphql-java/pull/2364/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R86
and
https://github.com/graphql-java/graphql-java/pull/2364/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R109

We don't need this?

src/test/java/benchmark/AddError.java
andimarek: I didn't need to do that last time. Maybe you need the additional gradle changes I mentioned here: https://github.com/graphql-java/graphql-java/pull/2871#issuecomment-1164909737

andimarek:
@adarsh-jaiswal I was looking into this a bit more and it seemed that the jmh plugin is a bit outdated. We need to update the jmh version the plugin is using with:

```
dependencies {
    testImplementation 'org.openjdk.jmh:jmh-core:1.35'
    testAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.35'
    jmh 'org.openjdk.jmh:jmh-core:1.35'
    jmh 'org.openjdk.jmh:jmh-generator-annprocess:1.35'
}
```
The first two dependencies are already there, but I upgraded to 1.35 jmh.
The last two are special instructions for the jmh plugin

After this it worked for me without adding any `throws InterrruptedException`.

Issue #2883, opened 2022-07-13
ClassCastException when type extension implements interface and placed before it base


tinnou:
## Issue
If a type extension implementing an interface is placed before its base implementation and there is a field returning the same type.

## Steps to reproduce
The test below showcases the issue by attempting create a schema.

```groovy
def "classCastException when interface extension is before base and has recursion"() {
        given:
        def spec = '''
        # order is important, moving extension below type Foo will fix the issue
        extend type Foo implements HasFoo {
          foo: Foo
        }
        
        type Query {
          test: ID
        }
        
        interface HasFoo {
          foo: Foo
        }
        
        type Foo {
          id: ID
        }
        '''

        when:
        TestUtil.schema(spec)

        then:
        noExceptionThrown()
    }
```
will throw 

```
Expected no exception to be thrown, but got 'java.lang.ClassCastException'

	at spock.lang.Specification.noExceptionThrown(Specification.java:118)
	at graphql.schema.idl.SchemaGeneratorTest.classCastException when interface extension is before base and has recursion(SchemaGeneratorTest.groovy:2534)
Caused by: java.lang.ClassCastException: graphql.schema.GraphQLTypeReference cannot be cast to graphql.schema.GraphQLInterfaceType
	at graphql.schema.idl.SchemaGeneratorHelper.lambda$null$26(SchemaGeneratorHelper.java:661)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:422)
	at graphql.schema.idl.SchemaGeneratorHelper.lambda$buildObjectTypeInterfaces$27(SchemaGeneratorHelper.java:660)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at graphql.schema.idl.SchemaGeneratorHelper.buildObjectTypeInterfaces(SchemaGeneratorHelper.java:660)
	at graphql.schema.idl.SchemaGeneratorHelper.buildObjectType(SchemaGeneratorHelper.java:645)
	at graphql.schema.idl.SchemaGeneratorHelper.buildOutputType(SchemaGeneratorHelper.java:755)
	at graphql.schema.idl.SchemaGeneratorHelper.buildField(SchemaGeneratorHelper.java:795)
	at graphql.schema.idl.SchemaGeneratorHelper.lambda$buildInterfaceType$19(SchemaGeneratorHelper.java:593)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:422)
[...]

PR #2884, opened 2022-07-13
#2883 Fix classCastException when using an extension 

tinnou:
Fix for #2883.

Callers of `buildOutputType` were wrongly expecting result could be cast to `GraphQLInterfaceType`.  Using `GraphQLNamedOutputType` is a wider type that is compatible.


bbakerman (approved):
Thanks

PR #2885, opened 2022-07-14
Make ValuesResolver methods static

dondonz:
Make every method in `ValuesResolver` static and save some object allocations. It's a util class that doesn't need to have instances.


andimarek:
@dondonz Lets add also a private constructor to ValuesResolver to make sure nobody can instantiate it.

bbakerman (approved):
Nice one

PR #2889, opened 2022-07-19
Added workflow and gcp JS client library script with package json

DiegoManzanarezx:
The purpose of this draft PR is to add a Github actions workflow that will be executed every time there is a push to this repository or when manually triggering it through Github UI.
The workflow will run a Javascript Google Cloud Client library file that will create and send a task to a Google Cloud Tasks Queue and will then forward it to a Google Cloud Workflow that will execute the test runner.

This draft PR consists on adding 3 new files:

1. A Github actions yaml file to install all the dependencies needed to successfully run a Javascript file that implements Google Cloud Tasks Queue Client library.
2. A Javascript file that create and send a task to a Queue, using Google Cloud Tasks Queue Client library.
3. A package.json file that stores all the dependencies needed for the JS file.

*To verify the reliability of the workflow we have first built and tested a fully functional prototype that works as expected.

Considerations:
a) In order to execute the workflow we need credentials and other information (i.e. the name of our project in Google Cloud Platform) that we will provide, but must remain confidential, therefore, those should be stored as Github Actions secrets and are the following:
GOOGLE_APPLICATION_CREDENTIALS -> This is a JSON file that will authenticate our GCP account at every execution.
LOCATION -> Location where the Queue is running.
PROJECT_ID -> The name of the GCP project where the Queue is stored.
QUEUE_ID -> The name of the Queue where the task will be sent.
SERVICE_ACCOUNT_ROLE -> The GCP account that contains the permissions and roles to manage the Queue.
WORKFLOW_URL -> The endpoint to which the Queue will send an http request to forward the task to a GCP Workflow.

b) The purpose of triggering manually the workflow is to run personalized tests, so 3 inputs will we requested:
     - Commit hash. (required)
     - Classes to test. (optional)
     - Pull request number. (optional)
    *If the optional fields are left empty, all tests will be executed on that specific commit.
    We will comment a link to the tests results on this pull request.

c) Since we are still working on the test runner this draft PR is intended just for code reviews purposes. We plan on landing the PR once the project is on a working state, probably about four weeks time.


.github/workflows/gql-perf-arch-github_workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+  workflow_dispatch:
+    inputs:
+      COMMIT_HASH_INPUT:
+        description: 'Commit hash'
+        required: true
+      CLASSES_TO_EXECUTE_INPUT:
+        description: 'Classes to test'
+        required: false
+      PULL_REQUEST_NUMBER_INPUT:
+        description: 'Pull request number'
```
chriswr95, line 15: can you the usage of each of these parameters to the description? for example `if this is specified, we will comment a link to the tests results on this pull request`.

.github/workflows/gql-perf-arch-github_workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+        required: false
+env:
+  #Actions secrets (keys)
+  PROJECT_ID: ${{ secrets.PROJECT_ID }}
+  QUEUE_ID: ${{ secrets.QUEUE_ID }}
+  LOCATION: ${{ secrets.LOCATION }}
```
chriswr95, line 21: what is this?

.github/workflows/gql-perf-arch-github_workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+  WORKFLOW_URL: ${{ secrets.WORKFLOW_URL }}
+  SERVICE_ACCOUNT_EMAIL: ${{ secrets.SERVICE_ACCOUNT_EMAIL }}
```
chriswr95, line 23: why is this needed?

google_cloud_http_target_task.js
```diff
@@ -0,0 +1,63 @@
...
+  const serviceAccountEmail = parsedDocument.SERVICE_ACCOUNT_EMAIL;
+
+  // Construct payload
+  const payloadStructure =  {
+                                 "jobId": jobId,
+                                 "commitHash": commitHash,
+                                 "classes": classes,
+                                 "pullRequest": pullRequestNumber,
+                            }
+  //Format payload
```
chriswr95, line 35: I prefer to have comments only for un-intuitive bits of code. Where you have a straight forward comment like this, I think really what you're after is a helper method definition. For example

const formatPayload = () => {
   // whatever you need to do to format the payload.
}

google_cloud_http_target_task.js
```diff
@@ -0,0 +1,63 @@
...
+  const parsedPayload = JSON.stringify(JSON.stringify(payloadStructure));
+  const payload = `{"argument": ${parsedPayload}}`;
+
+  console.log(`Payload: ${payload}`);
+
+  // Construct the fully qualified queue name
+  const parent = client.queuePath(project, location, queue);
```
chriswr95, line 42: why is this variable named parent? should it instead be named `fullyQualifiedQueueName`?

google_cloud_http_target_task.js
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  console.log(`Task: ${task}`);
+  const request = {parent: parent, task: task};
+  const [response] = await client.createTask(request);
+  const name = response.name;
+  console.log(`Created task ${name}`);
+  console.log(`Your job id is: ${jobId}`);
+}
+createHttpTaskWithToken();
```
chriswr95, line 63: let's chat in person about refactoring this script just a bit, overall it looks pretty good.

.github/workflows/gql-perf-arch-github_workflow.yml
```diff
@@ -0,0 +1,70 @@
...
+  push_action:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14'
+      - run: npm install
+      - run: npm i yenv
```
chriswr95, line 38: if you add `yenv` to your package.json, can you skip this extra step, as it will be installed during `npm install`?

package.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "dependencies": {
+    "@google-cloud/tasks": "^3.0.0",
+    "body-parser": "^1.18.3",
+    "express": "^4.16.3"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "mocha": "^8.0.0",
+    "uuid": "^8.0.0"
```
chriswr95, line 21: is uuid only a dev dependency? I think it should be a regular dependency. Similarly, do you need mocha and chai? I we don't have any tests for this code.

google_cloud_http_target_task.js
```diff
@@ -0,0 +1,63 @@
+const {CloudTasksClient} = require('@google-cloud/tasks');
```
chriswr95, line 1: this probably shouldn't live at the root level, maybe instead at `.github/workflow-scripts`?

google_cloud_http_target_task.js
```diff
@@ -0,0 +1,63 @@
...
+const yenv = require('yenv')
+const { v4: uuidv4 } = require('uuid');
+const fs = require('fs')
+const os = require('os');
```
chriswr95, line 5: is this used anywhere?

DiegoManzanarezx: No, I have already fixed it in the new commit.

google_cloud_http_target_task.js
DiegoManzanarezx: Done, I tried to reduce the work load of the main function by creating multiple self documented functions.

.github/workflows/gql-perf-arch-github_workflow.yml
DiegoManzanarezx: I already updated the description of the draft PR to explain why we need each of those variables.

.github/workflows/google_cloud_http_target_task.js
```diff
@@ -0,0 +1,66 @@
+const { CloudTasksClient } = require('@google-cloud/tasks');
+const yenv = require('yenv')
+const { v4: uuidv4 } = require('uuid');
+
+const parsedDocument = yenv('.github/workflows/gql-perf-arch-github_workflow.yml')
```
adarsh-jaiswal, line 5: Is it possible to use process.env.COMMIT_HASH and likewise for all other env variables? we might not need yenv dependency.

adarsh-jaiswal: cc

package.json
DiegoManzanarezx: Yes, I have already updated this file. We don't need mocha an chai.

.github/workflows/gql-perf-arch-github_workflow.yml
DiegoManzanarezx: I added it to the description.

google_cloud_http_target_task.js
DiegoManzanarezx: Done. I added it inside .github/workflow

.github/workflows/google_cloud_http_target_task.js
DiegoManzanarezx: Yes, you were right. The change is done.

.github/workflows/google_cloud_http_target_task.js
```diff
@@ -0,0 +1,66 @@
...
+  const serviceAccountEmail = parsedDocument.SERVICE_ACCOUNT_EMAIL;
+  const requestBody = {
+      "fullyQualifiedQueueName": client.queuePath(project, location, queue),
+      "task": constructTask(serviceAccountEmail, payload, url)
+  }
+  return requestBody;
+}
+
+async function createHttpTaskWithToken() {
+  const payloadStructure = constructPayload();
```
chriswr95, line 58: you can wrap lines 58 - 61 into yet another helper function `createRequest()` which returns the request.

.github/workflows/gql-perf-arch-github_workflow.yml
```diff
@@ -0,0 +1,50 @@
...
+        description: 'Commit hash'
+        required: true
+      CLASSES_TO_EXECUTE_INPUT:
+        description: 'Classes to test'
+        required: false
+      PULL_REQUEST_NUMBER_INPUT:
+        description: 'Pull request number'
+        required: false
+env:
+  #Actions secrets (keys)
```
chriswr95, line 18: can remove this comment.

chriswr95 (approved):
This code looks much tighter. 

NIT: For your PR description, you are now a little bit **too** verbose. Aim for clear and succinct language. You should spend 20 minutes writing the description so that a reviewer can read it in one minute. Using Markdown also helps to structure longer messages. No need to change anything, just leaving this for the future :)

Approved pending a final nit comment.

package.json
andimarek: I am not really comfortable with having this in the root folder. Can we move this file into the .github/workflow directory? 
The reason is security and maintenance: this is not s JS project and we don't depend on JS libraries for the actual lib we publish. This is should be clearly only used for the github workflow

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "files": [
+    "*.js"
+  ],
+  "dependencies": {
+    "@google-cloud/tasks": "^3.0.0",
+    "body-parser": "^1.18.3",
+    "express": "^4.16.3",
```
andimarek, line 16: do we need express here? I don't think we are running any server, right?

package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "appengine-cloudtasks",
+  "description": "Google App Engine Cloud Tasks example.",
+  "license": "Apache-2.0",
+  "author": "Google Inc.",
```
andimarek, line 5: I don't think this is right. We could just remove author, license, descp etc as we don;t plan to publish this

DiegoManzanarezx: True, I removed those fields.

package.json
DiegoManzanarezx: True, we don't need express. Already removed it.

package.json
DiegoManzanarezx: Thanks for that Andi! I moved it into **workflows** folder.

PR #2942, opened 2022-09-05
Deprecation test cleanup: GraphQLFieldDefinition datafetcher builder

dondonz:
Cleaning up tests to use `GraphQLCodeRegistry` for datafetchers, rather than directly on the `GraphQLFieldDefinition`.

You might be thinking, that's so many lines of code for only one builder cleanup! We were using it everywhere.


bbakerman (approved):
Wow there is heaps

PR #2971, opened 2022-09-23
Patch SchemaDiff to apply respective nullability change validation for input vs. output types

bspeth:
## Description
SchemaDiff had a single method `checkTypeWithNonNullAndList` that always applied the rule "non-null to nullable is safe; nullable to non-null is breaking."
This is true for input types and arguments, but the opposite should be true for fields on output types -- "nullable to non-null is safe; non-null to nullable is breaking."
A consumer operating on a nullable field should already be checking for presence, and adding an additional guarantee of non-null is safe.
A consumer operating on a non-null field may not be checking for presence, and removing the guarantee of non-null may break when that field is null.

## Implementation
Two methods are defined -- one for input and argument types, and one for output types.
The methods are rewritten in a style of handling three top-level cases for the old type (non-null, list, and neither), and within each case handling the cases for the new type.
Calls to the previous method are updated to their respective new versions.

graphql-js implementation for reference:
[isChangeSafeForObjectOrInterfaceField](https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/utilities/findBreakingChanges.ts#L445)
[isChangeSafeForInputObjectFieldOrFieldArg](https://github.com/graphql/graphql-js/blob/a24a9f35b876bdd0d5050eca34d3020bd0db9a29/src/utilities/findBreakingChanges.ts#L480)

## Testing
Unit tests for the new methods are updated and added:
- change_in_null_ness_object_or_interface
- change_in_null_ness_input_or_arg
- change_in_list_ness_object_or_interface
- change_in_list_ness_input_or_arg

schema_ABaseLine:
- Added Query.allCharactersByTemperament : [[Character]] to test a complex wrapped type.
- Made Questor.beingID non-null to be able to test "non-null to nullable" for input.

schema_changed_input_object_fields:
- Added case Query.being( .. , newArg: String!) -> stricter (moved from schema_changed_object_fields to here)
- Added case Questor.beingID "required to optional" -> safe
- Added case Questor.nestedInput "optional to required" -> stricter

schema_changed_object_fields:
- Moved case Query.being( .. , newArg: String!) to schema_changed_object_fields
- Added case Query.allCharactersByTemperament : [[Character!]]! to test complex wrapped type "optional to required" -> safe
- Added case Being.ID "optional to required" -> safe
- Added case Istari.temperament "required to optional" -> stricter

### DIFF
--- a/src/main/java/graphql/schema/diff/SchemaDiff.java
+++ b/src/main/java/graphql/schema/diff/SchemaDiff.java
@@ -854,7 +854,7 @@ public class SchemaDiff {
                     // if they're both lists, compare the unwrapped types
                     oldTypeInfo = oldTypeInfo.unwrapOne();
                     newTypeInfo = newTypeInfo.unwrapOne();
-                } else if (newTypeInfo.isNonNull()){
+                } else if (newTypeInfo.isNonNull()) {
                     // nullable to non-null creates a stricter input requirement for clients to specify
                     return DiffCategory.STRICTER;
                 } else {
@@ -869,7 +869,7 @@ public class SchemaDiff {
                     // non-list to list is not valid
                     return DiffCategory.INVALID;
                 } else {
-                   return null;
+                    return null;
                 }
             }
         }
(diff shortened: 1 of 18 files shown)


bspeth:
Hi @bbakerman, is this something you can help review? I see you originally wrote SchemaDiff and have looked at a couple other changes to it this year.

bspeth:
Hey @bbakerman, is there anything I can/should do to help with respect to this change?

bbakerman (approved):
Thanks for this excellent and well tested change

PR #2977, opened 2022-10-03
Update GraphQL Instrospection Spec Link

cookieMr:


cookieMr:
The old link no longer works. And it took me a while to find the
proper link with all the info i needed.

PR #2988, opened 2022-10-17
Add missing equals/hashcode methods to relay classes

pgr0ss:
Add equals/hashcode methods to DefaultConnection, DefaultEdge, DefaultPageInfo.

As discussed: https://github.com/graphql-java/graphql-java/discussions/2986

This is my first time using Spock, so please let me know if there's a different or better way to do this testing. I used the tables to try to reduce the test noise.


bbakerman (approved):
Thanks

The tests look spot on

PR #3014, opened 2022-11-11
Bugfix for SDL check if an Interface is implemented correctly

andimarek:
Applied directives on arguments are not relevant when checked if an interface is implemented correctly.

For example this SDL should be valid, but it was rejected:
```graphql
directive @d on ARGUMENT_DEFINITION
type Query implements I {
  foo(arg: String @d): String
}
interface I {
  foo(arg: String): String
}
```
This fixes this bug.


bbakerman (commented):
Nice

PR #3042, opened 2022-12-05
parseValue coercion alignment for Boolean, Float, and Int

dondonz:
Aligning `parseValue` coercion to be closer to the JavaScript reference implementation https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts

Key change is that a string will no longer be accepted as an input, e.g. `"true"` is no longer accepted as a `parseValue` input for Boolean.

There is no change corresponding change for ID as coercion for ID was already aligned.

I tried to make coercion as close as possible, keeping in mind that JavaScript uses the floating point `Number` type to represent integer values.

### DIFF
--- a/src/main/java/graphql/scalar/GraphqlIntCoercing.java
+++ b/src/main/java/graphql/scalar/GraphqlIntCoercing.java
@@ -63,7 +63,13 @@ public class GraphqlIntCoercing implements Coercing<Integer, Integer> {
     }
 
     @NotNull
-    private Integer parseValueImpl(Object input, @NotNull Locale locale) {
+    private Integer parseValueImpl(@NotNull Object input, @NotNull Locale locale) {
+        if (!(input instanceof Number)) {
+            throw new CoercingParseValueException(
+                    i18nMsg(locale, "Int.notInt", typeName(input))
+            );
+        }
+
         Integer result = convertImpl(input);
         if (result == null) {
             throw new CoercingParseValueException(
--- a/src/main/java/graphql/scalar/GraphqlIntCoercing.java
+++ b/src/main/java/graphql/scalar/GraphqlIntCoercing.java
@@ -64,19 +64,45 @@ public class GraphqlIntCoercing implements Coercing<Integer, Integer> {
 
     @NotNull
     private Integer parseValueImpl(@NotNull Object input, @NotNull Locale locale) {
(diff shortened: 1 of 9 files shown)


src/main/java/graphql/scalar/GraphqlIntCoercing.java
```diff
@@ -63,14 +63,46 @@ private Integer serialiseImpl(Object input, @NotNull Locale locale) {
...
         if (result == null) {
             throw new CoercingParseValueException(
                     i18nMsg(locale, "Int.notInt", typeName(input))
             );
         }
-        return result;
+
+        if (result.compareTo(INT_MIN) < 0 || result.compareTo(INT_MAX) > 0) {
+            throw new CoercingParseValueException(
+                    i18nMsg(locale, "Int.outsideRange", result.toString())
```
dondonz, line 86: For values outside of the integer range, we previously threw an exception with the same `Int.notInt` message. We ought to use `Int.outsideRange` for this case, as we do in `parseLiteral`

src/test/groovy/graphql/ScalarsBooleanTest.groovy
```diff
@@ -53,7 +53,6 @@ class ScalarsBooleanTest extends Specification {
     def "Boolean serialize #value into #result (#result.class)"() {
         expect:
         Scalars.GraphQLBoolean.getCoercing().serialize(value, GraphQLContext.default, Locale.default) == result
-        Scalars.GraphQLBoolean.getCoercing().parseValue(value, GraphQLContext.default, Locale.default) == result
```
dondonz, line 56: Moved into separate test below

src/test/groovy/graphql/ScalarsFloatTest.groovy
```diff
@@ -58,7 +58,6 @@ class ScalarsFloatTest extends Specification {
     def "Float serialize #value into #result (#result.class)"() {
         expect:
         Scalars.GraphQLFloat.getCoercing().serialize(value, GraphQLContext.default, Locale.default) == result
-        Scalars.GraphQLFloat.getCoercing().parseValue(value, GraphQLContext.default, Locale.default) == result
```
dondonz, line 61: Moved into separate test below

bbakerman (approved):

Issue #3065, opened 2023-01-03
FieldValidationInstrumentation.beginExecuteOperation does not forward the state and throws assertionerror


kfwerf:
Hi good folks of graphql-java!

We are trying to switch from version 19.* to version 20.* of the library. However, in doing so we ran into some issues with our local validator instrumentation. I didn't see a bug ticket for this yet, so please correct me if I somehow missed it!

**Describe the bug**
In version 20.0 we cannot use the `FieldValidationInstrumentation` anymore. This is due to it using a method that throws a deprecated warning. Considering `FieldValidationInstrumentation` not deprecated, we think this was a mistake. The `FieldValidationInstrumentation` fails to forward the `state` to the `super.beginExecuteOperation(parameters)`, which should be `super.beginExecuteOperation(parameters, state)`. Due to this it gives an assertion error;

```
graphql.AssertException: Internal error: should never happen: The deprecated beginExecuteOperation was calledgraphql.AssertException: Internal error: should never happen: The deprecated beginExecuteOperation was called at graphql.Assert.assertShouldNeverHappen(Assert.java:53) at graphql.execution.instrumentation.SimplePerformantInstrumentation.beginExecuteOperation(SimplePerformantInstrumentation.java:91) at graphql.execution.instrumentation.fieldvalidation.FieldValidationInstrumentation.beginExecuteOperation(FieldValidationInstrumentation.java:46) at graphql.execution.instrumentation.ChainedInstrumentation.lambda$beginExecuteOperation$3(ChainedInstrumentation.java:143) 
```

*TL;DR;*
`FieldValidationInstrumentation.beginExecuteOperation` should forward the `state` in the method -> `super.beginExecuteOperation(parameters, state)` 

**To Reproduce**
If you run something like this; with a active query it will result in a assertion error.

```
import graphql.GraphQLError;
import graphql.execution.instrumentation.ChainedInstrumentation;
import graphql.execution.instrumentation.Instrumentation;
import graphql.execution.instrumentation.fieldvalidation.FieldValidation;
import graphql.execution.instrumentation.fieldvalidation.FieldValidationEnvironment;
import graphql.execution.instrumentation.fieldvalidation.FieldValidationInstrumentation;
import java.util.ArrayList;
import java.util.List;

class Scratch {

  public static void main(String[] args) {
    final var instrumentations = new ArrayList<Instrumentation>();
    instrumentations.add(new FieldValidationInstrumentation(new FieldValidation() {
      @Override
[...]

PR #3066, opened 2023-01-03
Update FieldValidationInstrumentation.java

kfwerf:
Fix FieldValidationInstrumentation to forward the `state`

Fixes https://github.com/graphql-java/graphql-java/issues/3065

### DIFF
--- a/src/main/java/graphql/execution/instrumentation/fieldvalidation/FieldValidationInstrumentation.java
+++ b/src/main/java/graphql/execution/instrumentation/fieldvalidation/FieldValidationInstrumentation.java
@@ -43,6 +43,6 @@ public class FieldValidationInstrumentation extends SimplePerformantInstrumentat
         if (errors != null && !errors.isEmpty()) {
             throw new AbortExecutionException(errors);
         }
-        return super.beginExecuteOperation(parameters);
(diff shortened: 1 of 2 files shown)


dondonz:
Thanks! Could you also add a test that would have failed with the old behaviour, and would be fixed with your change?

kfwerf:
Added a test for chained instrumentation

kfwerf:
> Thanks! Could you also add a test that would have failed with the old behaviour, and would be fixed with your change?

Done! Added one for ChainedInstrumentation

Aynemayne (approved):

kfwerf:
@dondonz any thoughts on this PR? We are blocked with upgrading until this gets resolved

Ekryd (approved):

dondonz (approved):
Thanks for finding the bug & writing the fix!

dondonz:
The master build version containing this fix will be available in Maven in a few minutes. Keep in mind one change between v20 and current master was the change from Java 8 to Java 11.

FYI we usually bundle up a few PRs together for a small release, so this fix will be included in the next release. In the meantime, you can use the master build version to get unblocked now.

kfwerf:
cool, thank you @dondonz ! Appreciate the quickness on all of this :)

PR #3071, opened 2023-01-14
Fix extend schema directives ANTLR rule

dondonz:
Thanks @act1on3 and Daniela for identifying that an ANTLR rule could be approved & writing the fix

I am raising this PR on their behalf. @act1on3 I've added you as co-author, I didn't have Daniela's GitHub account or email.


dondonz:
Force pushed to add co-author credit

PR #3073, opened 2023-01-22
Add missing getter and fix name consistency

gnawf:


src/main/java/graphql/schema/diffing/ana/SchemaDifference.java
```diff
@@ -97,7 +97,7 @@ public String getNewName() {
             return newName;
         }
 
-        public boolean isRenamed() {
+        public boolean isNameChanged() {
             return renamed;
```
andimarek, line 101: lets fix the property too?

PR #3097, opened 2023-02-21
Ensured that the MANIFEST.MF files is the first entry in the JAR File

schaefa:
as it is required for java.util.jar.JarFile to obtain the manfiest entry which is essential for Sling to work as discussed in #3089. This will fail to work with Sling's bundle installer as it is using JarFile to obtain the manifest and that is not found if it is not the first entry.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,20 +160,27 @@ Import-Package: !com.google.*,!org.checkerframework.*,!javax.annotation.*,!graph
 }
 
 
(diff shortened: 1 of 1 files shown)


dondonz:
Hello, thanks so much for opening a PR, I will take a look

bbakerman (approved):

dondonz:
Thanks so much @schaefa for reporting the issues and to top it off, opening a PR with the fix!

PR #3099, opened 2023-02-27
Backported the fix to remove sun.misc

schaefa:
and make sure MANIFEST.MF is the first entry in the JAR file.
Due to a class loading issue in 20.0 I am going to upgrade to 19.x in Sling Graphql Core and so I back ported the issues fixed in master.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -155,7 +155,7 @@ shadowJar {
     bnd('''
 -exportcontents: graphql.*
 -removeheaders: Private-Package
-Import-Package: !com.google.*,!org.checkerframework.*,!javax.annotation.*,!graphql.com.google.*,!org.antlr.*,!graphql.org.antlr.*,*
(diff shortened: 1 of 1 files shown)


dondonz:
Hello, thanks for the PR. As a heads up, we haven't yet decided what will be backported or if we will do another v19 release. We typically backport critical security fixes and we decide other fixes on a case by case basis. I will take this PR into consideration at the next release point.

schaefa:
Right now we (Sling) is having a problem that we cannot use graphql-java out of the box on any versions with the security fix in 17.4 to 20.0 because of that. In addition master has the issue with class loading that prevents us from using 20.x.
So this is, at least for us, a security issue.

dondonz:
To follow up, we will include the fix for the class loading problem #3095 in the next release. Hopefully you'll be able to use 20.1 when it's released

PR #3115, opened 2023-03-14
Add dependabot configuration

yeikel:
See https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github_actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github_actions"
```
dondonz, line 7:
```suggestion
  - package-ecosystem: "github-actions"
```
Could you please change it to a dash? From the documentation: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

dondonz (changes requested):
Great idea to add Dependabot, thanks

.github/dependabot.yml
yeikel: Thank you for the catch. I confused them because that's the name internally. Thanks for the review!

dondonz (approved):
Thanks so much for adding Dependabot!

And thanks for your first PR on this repo (hopefully, first of many to come!)

PR #3119, opened 2023-03-16
Add manual stop on algorithm

gnawf:
Adds `.stop()` function to `SchemaDiffing.java` and also checks for `Thread.interrupted()`

### DIFF
--- a/src/main/java/graphql/schema/diffing/SchemaDiffingCancelledException.java
+++ b/src/main/java/graphql/schema/diffing/SchemaDiffingCancelledException.java
@@ -1,0 +1,7 @@
+package graphql.schema.diffing;
+
+public class SchemaDiffingCancelledException extends RuntimeException {
+    SchemaDiffingCancelledException(boolean byInterrupt) {
+        super("Schema diffing job was cancelled by " + (byInterrupt ? "thread interrupt" : "stop call"));
+    }
+}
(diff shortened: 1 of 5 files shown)


dondonz (approved):

src/main/java/graphql/schema/diffing/SchemaDiffingCancelledException.java
```diff
@@ -0,0 +1,7 @@
+package graphql.schema.diffing;
+
```
andimarek, line 2: lets make this `Internal`

src/main/java/graphql/schema/diffing/SchemaDiffingRunningCheck.java
```diff
@@ -0,0 +1,20 @@
+package graphql.schema.diffing;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+class SchemaDiffingRunningCheck {
+    private final AtomicBoolean wasStopped = new AtomicBoolean(false);
+
+    void check() {
+        if (Thread.interrupted()) {
```
andimarek, line 9: Lets use only the boolean flag

andimarek (changes requested):
Lets not check the Thread flag at all and only use the provided boolean.

Also: mark all new classes as `@Internal`

PR #3146, opened 2023-03-23
Diffing fixes 2

gnawf:


src/main/java/graphql/schema/diffing/DiffImpl.java
```diff
@@ -390,6 +394,7 @@ private double calcLowerBoundMappingCost(Vertex v,
                 anchoredVerticesCost++;
             }
 
+            runningCheck.check();
```
gnawf, line 397: I added more checks to places that the algorithm would get stuck on when the changes were large.

bbakerman (approved):

PR #3168, opened 2023-04-03
ci: use temurin

yeikel:
Host runners include Temurin by default as part of the [hosted tool cache](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#hosted-tool-cache). Using Temurin speeds up builds as there is no need to download and configure the Java SDK with every build.

Before: 

image

After : 

image

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -19,6 +19,6 @@ jobs:
         uses: actions/setup-java@v3
         with:
           java-version: '8'
-          distribution: 'corretto'
+          distribution: 'temurin'
       - name: build test and publish
         run: ./gradlew assemble && ./gradlew check --info && ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository -x check --info --stacktrace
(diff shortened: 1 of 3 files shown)


dondonz:
Hello, thanks for PR and spending the time to research the build.

The maintainer team had a discussion and we'd like to use Amazon's Corretto (in a forthcoming PR), as it's widely used and build speeds aren't too slow at present.

yeikel:
> Hello, thanks for PR and spending the time to research the build.
> 
> The maintainer team had a discussion and we'd like to use Amazon's Corretto, as it's widely used and build speeds aren't too slow at present.

I respect your decision. 

Temurin/Adopt is very popular. Based on the number of github.com search results, I assume that is more popular in OSS than corretto and it is widely used in many popular OSS projects

If you're curious : 

https://github.com/search?q=temurin+path%3A.github%2Fworkflows&type=code -> 34.6k files
https://github.com/search?q=adopt+path%3A.github%2Fworkflows&type=code  -> 40.8k files  (these repos should migrate to temurin at some point as adopt is no longer maintained)

https://github.com/search?q=corretto+path%3A.github%2Fworkflows&type=code -> 1.8k files

PR #3186, opened 2023-04-14
Revert stricter scalar parseValue coercion

dondonz:
Reverts stricter `parseValue` coercion for built-in scalars String #3030 and Boolean, Float, and Int #3042. 

We've received feedback that this `parseValue` breaking change is causing an issue with clients expecting the old  behaviour. Whilst this change was to align `parseValue` coercion with the JS reference implementation, we want to temporarily revert this change while we work on a better migration solution.

The stricter parseValue coercion was previously released with [GraphQL Java 20.0](https://github.com/graphql-java/graphql-java/releases/tag/v20.0).

JS implementation: https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts

### DIFF
--- a/src/test/groovy/graphql/ScalarsIntTest.groovy
+++ b/src/test/groovy/graphql/ScalarsIntTest.groovy
@@ -137,12 +137,15 @@ class ScalarsIntTest extends Specification {
         new Short("42")       | 42
         1234567l              | 1234567
         new AtomicInteger(42) | 42
-        Integer.MAX_VALUE     | Integer.MAX_VALUE
-        Integer.MIN_VALUE     | Integer.MIN_VALUE
         42.0000d              | 42
         new BigDecimal("42")  | 42
         42.0f                 | 42
         42.0d                 | 42
+        Integer.MAX_VALUE     | Integer.MAX_VALUE
+        Integer.MIN_VALUE     | Integer.MIN_VALUE
+        "42"                  | 42
+        "42.0000"             | 42
+        "-1"                  | -1
     }
 
     @Unroll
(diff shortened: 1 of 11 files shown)


src/test/groovy/graphql/ScalarsIntTest.groovy
```diff
@@ -137,12 +137,15 @@ class ScalarsIntTest extends Specification {
         new Short("42")       | 42
         1234567l              | 1234567
         new AtomicInteger(42) | 42
-        Integer.MAX_VALUE     | Integer.MAX_VALUE
-        Integer.MIN_VALUE     | Integer.MIN_VALUE
```
dondonz, line 141: Some reorganising so these two adjacent tests are ordered in the same way

src/test/groovy/graphql/ScalarsIntTest.groovy
```diff
@@ -152,18 +155,21 @@ class ScalarsIntTest extends Specification {
 
         where:
         value                 | result
-        42.0000d              | 42
         new Integer(42)       | 42
         new BigInteger("42")  | 42
-        new BigDecimal("42")  | 42
-        42.0f                 | 42
-        42.0d                 | 42
```
dondonz, line 160: Reorganising

src/test/groovy/graphql/ScalarsStringTest.groovy
```diff
@@ -86,24 +85,15 @@ class ScalarsStringTest extends Specification {
...
 
         where:
-        literal      | _
-        123          | _
-        true         | _
-        customObject | _
+        value        | result
+        123          | "123"
+        true         | "true"
+        customObject | "foo"
```
dondonz, line 96: As all Java objects must have a `toString` method, there will always be a result for `parseValue` with the looser coercion.

bbakerman (approved):

jbellenger:
👋 howdy! I've recently been looking into upgrading a v18.1 service to v20.2 and have been looking into how to safely migrate to this (now reverted) part of the update. 

For example, an early attempt at this was to try modifying scalars like Boolean to use a new Coercing that could log compatibility issues. I couldn't get this to work, as the standard Boolean and its Coercing are baked into skip/include directives in a way that makes them difficult to replace.

It would be great to get this un-reverted eventually, but for now I'm interested in following any thinking about what a migration solution might look like. Is there a way to keep tabs on this?

andimarek:
hey @jbellenger ... so we decided to reverted it, because even if we think this is the right behaviour there is a chance of some existing clients could break.

Have a look at https://github.com/graphql-java/graphql-java/pull/3188 and maybe comment if that would help you: we plan to introduce a temporary internal only API for one or two releases that will allow everybody to migrate more safely.

dondonz:
You can keep track of the progress on #3191, I'll continue to add changes related to `parseValue` coercion to that issue

PR #3192, opened 2023-04-18
Shortcut diffing execution if it is known to take long

gnawf:
So I'm not too sure how to introduce this. I mean we're the only consumers of this so it should be fine. But the size of the input graphs is also a contributing factor. But our input graphs are mainly static, so for us simply checking these sizes are fine…

### DIFF
--- a/src/main/java/graphql/schema/diffing/ana/EditOperationAnalyzer.java
+++ b/src/main/java/graphql/schema/diffing/ana/EditOperationAnalyzer.java
@@ -786,11 +786,17 @@ public class EditOperationAnalyzer {
 
         if (fieldsContainerForField.isOfType(SchemaGraph.OBJECT)) {
             Vertex object = fieldsContainerForField;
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/schema/diffing/ana/EditOperationAnalyzer.java
```diff
@@ -786,11 +786,17 @@ private void fieldChanged(EditOperation editOperation) {
 
         if (fieldsContainerForField.isOfType(SchemaGraph.OBJECT)) {
             Vertex object = fieldsContainerForField;
+            if (isObjectAdded(object.getName())) {
+                return;
+            }
```
gnawf, line 791: Ran into this bug when I ran `shortcuts execution on big changes` test without the shortcut code.

gnawf:
Closing this as it's no longer needed. If we need it in the future we can revive it.

PR #3204, opened 2023-05-02
Sort diff edits before analysis

gnawf:
Ok so I tried to do some 2nd or 3rd loops but there's no easy way to get like Object -> Fields -> Arguments.

So I just decided to go with the sorting option.


andimarek (changes requested):
Can we add a test that shows this works as intended?

gnawf:
Added test. Also manually verified it works with some past commits.

Before

Screenshot 2023-05-03 at 3 38 41 PM

After

Screenshot 2023-05-03 at 3 42 19 PM

PR #3206, opened 2023-05-02
Add map function to DataFetcherResult 

konnik:
This PR adds a map function to DataFetcherResult (i.e. makes it a functor) to support a functional style of programming.


bbakerman (approved):
Thanks for this PR.  I think this is a great edition for those using DataFetcherResult

PR #3265, opened 2023-07-07
Use the non deprecated method in InstrumentationExamples.java

appreciated:
Currently, a deprecated method is still being used in InstrumentationExamples.java. This causes an NPE during the execution of `beginExecution`. Instead, use the non deprecated method in the example.


dondonz:
Thanks!

PR #3298, opened 2023-08-23
Schema Printer - add deprecation reason printing and fix missing deprecated directive on argument

vadimofd:
This PR addresses two schema printer items:
1. Adds deprecation reason printing to better support programmatically provided deprecation reason which is required by schema linters such as GraphQL-ESLint schema recommended rule [require-deprecation-reason](https://the-guild.dev/graphql/eslint/rules/require-deprecation-reason).
2. Fixes `@deprecated` directive missing from printed arguments.


bbakerman (approved):
Thanks you for this.  The simplifies the code in many ways.

That `DeprecatedAppliedDirective4Printing` was a lot of a hack.  This is much nicer

PR #3299, opened 2023-08-24
WIP Edge case where schema diffing has NPE

dondonz:
WIP, this is a failing test for now

There are a number of schema changes, this particular combination is causing an NPE because Mappings is null


src/test/groovy/graphql/schema/diffing/ana/EditOperationAnalyzerAppliedDirectivesTest.groovy
```diff
@@ -32,6 +32,1183 @@ import static graphql.schema.diffing.ana.SchemaDifference.UnionModification
 
 class EditOperationAnalyzerAppliedDirectivesTest extends Specification {
```
dondonz, line 33: PS this is a quick reproduction, I will move the test to another file later

dondonz:
Fixed by #3304

PR #3304, opened 2023-08-27
Permit parent restricted nodes to map to isolated nodes

gnawf:


bbakerman (approved):
I have to trust this - I have no background on this code

PR #3336, opened 2023-09-21
Use string concat instead of string builder when building field glob name

tylerscoville:
Using `+` can be expensive with a sufficiently large schema and query. This is a micro-optimization to swap `+`/`StringBuilder` for `.concat`

### DIFF
--- a/src/main/java/graphql/schema/DataFetchingFieldSelectionSetImpl.java
+++ b/src/main/java/graphql/schema/DataFetchingFieldSelectionSetImpl.java
@@ -264,7 +264,11 @@ public class DataFetchingFieldSelectionSetImpl implements DataFetchingFieldSelec
     }
 
     private static String mkFieldGlobName(String fieldPrefix, String fieldName) {
-        return (!fieldPrefix.isEmpty() ? fieldPrefix + SEP : "") + fieldName;
+        if (fieldPrefix.isEmpty()) {
+            return fieldName;
+        }
+
+        return fieldPrefix.concat(SEP).concat(fieldName);
     }
 
     private static PathMatcher globMatcher(String fieldGlobPattern) {


bbakerman:
As I understand it the `+` operator in Java gets changed by the compiler to to a StringBuilder at runtime

With large enough strings, StringBuilder is significantly faster than String.concat() which allocates a String for each call.

https://www.baeldung.com/java-string-concatenation-methods


That all said - this is mute because the strings in this case are going to be small and probably called a small number of times with a small number bytes

I would love to know more if you saw this as a performance hotspot in your code?

tylerscoville:
This is likely a minuscule performance improvement. It's so tiny I considered not opening this PR. It did pop up on our profiling and I added a picture below. Our use case is that we're checking if the selection set contains certain strings which executes the code paths in the flame graph. I wound up opening this up because I thought even the tiniest of wins was still a win. 

image

lqc:
Just out of curiosity, can you tell if ```SEP.join(fieldPrefix, fieldName)``` would be better or worse than ```concat()``` ? If it's similar, then at least it doesn't look that magical, IMHO.

bbakerman:
OK one of the challenges with performance related changes is that they need to demonstrate that they are faster.

In this PR you have (excellently) showed evidence of how the old code shows up in profiling via the flame graph image above but you  have not showed the same thing after the change.

The problem I is that String.concat() allocates Strings anyway and I have evidence from others that its slower then `+` which is really a StringBuilder via the compiler.

So I am loathe to accept this PR - bit because its bad code but because it's anecdotally no better without support evidence

Now I don't say this from some high handed position.  We do really appreciate those who make contributions and we thank you.

tylerscoville:
That's fair if I find more evidence, I'll be sure to report it. Thanks for your consideration.

PR #3353, opened 2023-10-20
@oneOf handles non-null inputs

felipe-gdr:
Non-null inputs are not going through `@oneOf` validation.

For example, the query bellow will not blow up, even though both fields under the oneOf type are being passed.

```
  # schema
  type Query {
    search(by: SearchInput!): String
  }
  input SearchInput @oneOf {
    name: String
    id: ID
  }

  # query
  query myQuery {
    search(by: {name: "Figaro", id: "1001"})
  }
```


dondonz:
Nice! Thanks

dondonz (approved):

Issue #2709, opened 2022-02-01
`MultiSourceReader` has breaking behavioural changes when running on Java 16+


jord1e:
**Describe the bug**
Because of [`JDK-8241020` "LineNumberReader.getLineNumber() behavior is inconsistent with respect to EOF"](https://bugs.openjdk.java.net/browse/JDK-8241020), the `MultiSourceReader` class behaves incorrectly when running on Java 16+ VMs, even when compiling with Java 8. This causes breaking changes.

The following test cases are failing:
![afbeelding](https://user-images.githubusercontent.com/30464310/151972502-9eb01957-cbdf-4ff7-8c80-14294bbdbb9b.png)

**To Reproduce**

Run the tests on JDK 16, or higher
Same issue occurs when you compile on JDK 8 and run tests on JDK 16+ (see results below)

For ease of testing you may temporarily clone https://github.com/jord1e/graphql-java/tree/ci-improvements and change the versions in `gradle.properties` (or checkout https://github.com/jord1e/graphql-java/commit/a1f4f8455c21a1beb60d38d2baff6dabf93be719)

Results can be downloaded here:
https://github.com/jord1e/graphql-java/actions/runs/1778419182

Found whilst working on #2676 

Can't figure this one out on myself quickly...

jord1e:
.

dondonz:
Fixed by #3670

PR #3359, opened 2023-10-23
Add GitHub action to manage stale PRs and issues

dondonz:
Add a GitHub action to automatically comment on open PRs and issues, to help the maintainers close out older items.

Both PRs and issues will be closed after 90 days of inactivity.

This GitHub action runs once per day.

Based off Gradle's config https://github.com/gradle/gradle/blob/master/.github/workflows/stale-issue.yml

### DIFF
--- a/.github/workflows/stale-pr-issue.yml
+++ b/.github/workflows/stale-pr-issue.yml
@@ -1,0 +1,44 @@
+# Mark inactive issues and PRs as stale
+# GitHub action based on https://github.com/actions/stale
+
+name: 'Close stale issues and PRs'
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


jord1e:
Could you add https://github.com/actions/stale?tab=readme-ov-file#exempt-pr-labels

for issues like #2709 that are super low priority but will not go “stale”? As it is a bug

i think this could apply to at least some issues

dondonz:
@jord1e thanks for the suggestion

Although the bot will close issues and PRs, it won't delete any work so we can always refer back to items. We're going to trial the 90 day time period, to also give us some more incentive to close out PRs sooner

PR #3361, opened 2023-10-24
Earlier validation that document operations exist in the schema

dondonz:
_Update: Dec 2024_

This PR brings forward validation to check an operation exists in the schema.

Previously, this was checked at execution time. Now it will be checked at documentation validation time.

The motivation for this change was a change in the specification, which is currently in RFC2 and is about to be merged in. See https://github.com/graphql/graphql-spec/pull/955

We also had folks from the community request this. Originally from #2741 and #2740.

I've aligned names with the reference implementation https://github.com/graphql/graphql-js/commit/50d555c400b8c3d487dce5250825d901536d8f07


dondonz:
To do: check the reference implementation graphql-js

dondonz:
Interesting, this is a gap in the specification, which our .NET implementation friends have also found.

This is a proposal to formalise that mutation and subscription operations exist in the schema before execution. 

Issue link: https://github.com/graphql/graphql-spec/issues/1097
Spec proposal link: https://github.com/graphql/graphql-spec/pull/1098

dondonz:
Coming back to this PR, a new spec change is landing soon which adds validation for this failing test case, where an operation doesn't exist in the schema.

This PR has all the changes: https://github.com/graphql/graphql-spec/pull/955
This is another PR for context, but all changes were merged into the first one https://github.com/graphql/graphql-spec/pull/1098

dondonz:
3 days ago this RFC was promoted to the RFC 2 Draft stage https://github.com/graphql/graphql-spec/pull/955

src/test/groovy/graphql/GraphQLTest.groovy
```diff
@@ -370,7 +370,7 @@ class GraphQLTest extends Specification {
 
         then:
         result.errors.size() == 1
-        result.errors[0].class == MissingRootTypeException
+        ((ValidationError) result.errors[0]).validationErrorType == ValidationErrorType.UnknownOperation
```
dondonz, line 373: I'd like to discuss this

Previously, the same sort of problem would be caught at execution time, throwing a `MissingRootTypeException` if the operation could not be executed. Given that I've now moved forward validation, it's very unlikely the old exception will ever be triggered. However for safety I'd like to be defensive and leave in the old code. See `SchemaUtil` and `getOperationRootType`.

src/test/groovy/graphql/validation/rules/KnownDirectivesTest.groovy
```diff
@@ -246,6 +246,10 @@ class KnownDirectivesTest extends Specification {
             field: String
         }
         
+        type Subscription {
+            field: String
+        }
+
```
dondonz, line 252: This test needed an adjustment to be valid

src/main/java/graphql/validation/rules/KnownOperationTypes.java
```diff
@@ -0,0 +1,48 @@
...
+                && graphQLSchema.getQueryType() == null) {
+            // This is unlikely to happen, as a validated GraphQLSchema must have a Query type by definition
+            String message = i18n(UnknownOperation, "KnownOperationTypes.noOperation", formatOperation(documentOperation));
+            addError(UnknownOperation, operationDefinition.getSourceLocation(), message);
+        }
+    }
+
+    private String formatOperation(OperationDefinition.Operation operation) {
+        return StringKit.capitalize(operation.name().toLowerCase());
+    }
```
dondonz, line 47: Enums are ALL CAPS by default and I thought that was too shouty

src/main/java/graphql/validation/rules/KnownOperationTypes.java
```diff
@@ -0,0 +1,48 @@
...
+                && graphQLSchema.getMutationType() == null) {
+            String message = i18n(UnknownOperation, "KnownOperationTypes.noOperation", formatOperation(documentOperation));
+            addError(UnknownOperation, operationDefinition.getSourceLocation(), message);
+        } else if (documentOperation == OperationDefinition.Operation.SUBSCRIPTION
+                && graphQLSchema.getSubscriptionType() == null) {
+            String message = i18n(UnknownOperation, "KnownOperationTypes.noOperation", formatOperation(documentOperation));
+            addError(UnknownOperation, operationDefinition.getSourceLocation(), message);
+        } else if (documentOperation == OperationDefinition.Operation.QUERY
+                && graphQLSchema.getQueryType() == null) {
+            // This is unlikely to happen, as a validated GraphQLSchema must have a Query type by definition
```
dondonz, line 39: A valid GraphQLSchema should have already been checked for a query type, I am being defensive here

src/test/groovy/graphql/GraphQLTest.groovy
dondonz: Discussed offline - will keep in the old error in case, to be conservative

andimarek (approved):

PR #3395, opened 2023-12-12
Defer support on ENFs

felipe-gdr:
Add support for `@defer` on `ExecutableNormalizedField`. 

This implementation is based on the state of [Spec edits for `@defer` / `@stream` PR](https://github.com/graphql/graphql-spec/pull/742). More specifically at the state of this commit https://github.com/graphql/graphql-spec/pull/742/commits/c630301560d9819d33255d3ba00f548e8abbcdc6.

The execution behaviour should match what we get from running Apollo Server `4.9.5` with graphql-js `v17.0.0-alpha.2`


src/test/groovy/graphql/normalized/ExecutableNormalizedOperationToAstCompilerDeferTest.groovy
```diff
@@ -0,0 +1,448 @@
...
+            }
+          }
+        """
+        GraphQLSchema schema = mkSchema(sdl)
+        def fields = createNormalizedFields(schema, query)
+        when:
+        def result = compileToDocumentWithDeferSupport(schema, QUERY, null, fields, noVariables)
+        def printed = AstPrinter.printAst(new AstSorter().sort(result.document))
+        then:
+        printed == '''{
```
felipe-gdr, line 401: this test is currently failing. This is the result:
```
{
  dog {
    ... @defer {
      breed
      name
      owner {
        firstname
      }
    }
  }
}
```

felipe-gdr: I see 2 ways going forward:
- introduce a new class to represent a @defer block. This new class would share a type with ExecutableNormalizedField , so the children of an ENF could either be other ENFs, or @defer blocks.
- Build a map that associates fields with @defer outside of the ENF hierarchy, similarly to normalizedFieldToQueryDirectives.
I'm leaning towards option #2 because it seems less intrusive

src/main/java/graphql/normalized/ExecutableNormalizedOperationFactory.java
```diff
@@ -155,22 +187,47 @@ public int getMaxChildrenDepth() {
...
+     * in a runtime sense.
+     *
+     * @param graphQLSchema         the schema to be used
+     * @param document              the {@link Document} holding the operation text
+     * @param operationName         the operation name to use
+     * @param coercedVariableValues the coerced variables to use
+     * @param options               the {@link Options} to use for parsing
+     * @return a runtime representation of the graphql operation.
+     */
+    public static ExecutableNormalizedOperation createExecutableNormalizedOperation(
```
felipe-gdr-atlassian, line 220: I created a version of `createExecutableNormalizedOperation` that takes `Option`

src/main/java/graphql/normalized/ExecutableNormalizedOperationFactory.java
```diff
@@ -333,6 +386,13 @@ private ExecutableNormalizedOperation createNormalizedQueryImpl(GraphQLSchema gr
             normalizedFieldToMergedField.put(enf, mergedFld);
         };
 
+        ImmutableListMultimap.Builder<ExecutableNormalizedField, DeferExecution> normalizedFieldToDeferExecution = ImmutableListMultimap.builder();
+        normalizedFieldToDeferExecution.putAll(collectFromOperationResult.normalizedFieldToDeferExecution);
+
+        Consumer<CollectNFResult> captureCollectNFResult = (collectNFResult -> {
```
felipe-gdr-atlassian, line 395: I feel like working with an instance of some object created on a per-execution basis would make accessing these values much easier.  We could create instance fields instead of having to pass them down the stack.

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationToAstCompilerDeferTest.groovy
```diff
@@ -0,0 +1,552 @@
...
+          }
+        """
+        GraphQLSchema schema = mkSchema(sdl)
+        def tree = createNormalizedTree(schema, query)
+        when:
+        def result = compileToDocumentWithDeferSupport(schema, QUERY, null, tree.topLevelFields, noVariables, tree.normalizedFieldToDeferExecution)
+        def printed = AstPrinter.printAst(new AstSorter().sort(result.document))
+        then:
+        printed == '''{
+  dog {
```
felipe-gdr-atlassian, line 235: Not a blocker, but we should double-check what is the expected behaviour here with the drivers of the defer/stream spec change.
For reference, the Apollo server execution will block the response until `breed` is return in the initial payload, then it will sent a second, deferred, payload with `breed` again.

felipe-gdr-atlassian (commented):
- [ ] apply the code official code formatter for this repo
- [ ] add a reference to the version of the spec that this implementation is based on

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationToAstCompilerDeferTest.groovy
felipe-gdr-atlassian: Based on the direction the working group is taking (https://github.com/graphql/defer-stream-wg/discussions/69) regarding de-duplication, I propose we keep only the non-deferred version of the field. 
I can do it on a follow up PR.

src/main/java/graphql/normalized/ExecutableNormalizedOperationFactory.java
```diff
@@ -383,10 +453,9 @@ private void captureMergedField(ExecutableNormalizedField enf, MergedField merge
 
         private void buildFieldWithChildren(ExecutableNormalizedField executableNormalizedField,
                                             ImmutableList<FieldAndAstParent> fieldAndAstParents,
-                                            int curLevel,
-                                            int maxLevel) {
```
felipe-gdr-atlassian, line 387: Minor change: use `maxLevel` from `options` instead of passing it as argument

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationToAstCompilerDeferTest.groovy
```diff
@@ -0,0 +1,552 @@
...
+    name
+    ... @defer {
```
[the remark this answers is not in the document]
andimarek-atlassian: outdated ... we leave it like this

Issue #3409, opened 2024-01-05
ConditionalNodeDecision not invoked for every field


dfa1:
**Describe the bug**
I'm trying to use ConditionalNodeDecision but it not invoked for every field, just the top level one.

**To Reproduce**

```java
final ExecutionInput executionInput = ExecutionInput.newExecutionInput()
                .query(request.query())
                .operationName(request.operationName())
                .variables(request.variables())
                .graphQLContext(builder ->
.graphQLContext(builder ->
                        builder
.put(ConditionalNodeDecision.class, new Custom()) <-- 
```

but then in graphql.execution.conditional.ConditionalNodes#shouldInclude:
```java
        // if they have declared a decision callback, then we will use it but we expect this to be mostly
        // empty and hence the cost is a map lookup.
        if (graphQLContext != null) {
            ConditionalNodeDecision conditionalDecision = graphQLContext.get(ConditionalNodeDecision.class);
```

only the first call has the context, subsequent calls have empty context. 

The first call is originating in:
![image](https://github.com/graphql-java/graphql-java/assets/5798698/686ee057-c69b-4994-8071-15074f2a3e15)

and as you can see the context is defined.
The next call is originating in `getImmediateFields`:
![image](https://github.com/graphql-java/graphql-java/assets/5798698/7bc7905e-1cbd-4e20-aa26-ab8688320d20)
and at this point the context is empty.

Graphql-java version is 21.2 on JDK 17.

dondonz:
Thanks for reporting the bug! I was able to reproduce the problem and will take a closer look

dondonz:
Thanks again for the bug report! The fix is in #3411, the fix is to propagate the GraphQLContext when completing values as well.

dfa1:
@dondonz thanks for the quick fix. Would it be possible to merge to master? I can do some tests on my side as well :)

dondonz:
Yes it'll be merged soon, just waiting on a review

This branch is very close to the current HEAD of master

dondonz:
Fixed by #3411

PR #3411, opened 2024-01-09
Propagate GraphQLContext when completing fields - fix missing ConditionalNodesDecision issue

dondonz:
This PR fixes issue #3409. As @dfa1 wrote, when evaluating nodes at least 1 level lower than the top level field, the GraphQLContext is for some reason wiped out, which means a custom ConditionalNodesDecision is not applied.

The problem was the GraphQLContext was there when evaluating whether a top level field should be included, but any fields below top level is evaluated for inclusion when completing values. In this separate method, the GraphQLContext was not propagated in the parameters.

In case you're wondering, this bug did not affect `@include`/`@skip`, it only affected decisions stored in `GraphQLContext`.

### DIFF
--- a/src/test/groovy/graphql/execution/ConditionalNodesTest.groovy
+++ b/src/test/groovy/graphql/execution/ConditionalNodesTest.groovy
@@ -192,8 +192,7 @@ class ConditionalNodesTest extends Specification {
         then:
         er["data"] == ["in": "in", "out": "out"]
 
-        // TODO: this test demonstrates that the GraphQLContext is missing for Pet
-        // and therefore, fields are being incorrectly included
+        // A test for fields below the top level
         when:
         ei = ExecutionInput.newExecutionInput()
                 .graphQLContext(contextMap)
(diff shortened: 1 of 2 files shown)


src/test/groovy/graphql/execution/ConditionalNodesTest.groovy
```diff
@@ -183,6 +191,27 @@ class ConditionalNodesTest extends Specification {
...
+                  favouriteSnack @featureFlag(flagName : "OFF")
+                } 
+            }
+        """
+                ).build()
+        er = graphQL.execute(ei)
+
+        then:
+        // TODO: favouriteSnack should not appear in the data
+        er["data"] == ["in": "in", "pet": ["name": "name"]]
```
dondonz, line 214: ConditionalNodesTest > integration test showing conditional nodes can be custom included FAILED
Condition not satisfied:
er["data"] == ["in": "in", "pet": ["name": "name"]]
     | |        |
     | |        false
     | [in:in, pet:[name:name, favouriteSnack:favouriteSnack]]
     ExecutionResultImpl{errors=[], data={in=in, pet={name=name, favouriteSnack=favouriteSnack}}, dataPresent=true, extensions=null}
         at graphql.execution.ConditionalNodesTest.integration test showing conditional nodes can be custom included(ConditionalNodesTest.groovy:214)

dondonz:
Will chat with team later, I think this ought to be backported

PR #3430, opened 2024-02-01
Fix oneOf bug when variables is empty

felipe-gdr:
We were seeing occurrences of `NoSuchElementException`.

### DIFF
--- a/src/main/java/graphql/execution/ValuesResolverOneOfValidation.java
+++ b/src/main/java/graphql/execution/ValuesResolverOneOfValidation.java
@@ -75,23 +75,36 @@ final class ValuesResolverOneOfValidation {
     }
 
     private static void validateOneOfInputTypesInternal(GraphQLInputObjectType oneOfInputType, Value<?> argumentValue, Map<String, Object> objectMap, Locale locale) {
-        int mapSize;
-
+        final String fieldName;
         if (argumentValue instanceof ObjectValue) {
-            mapSize = ((ObjectValue) argumentValue).getObjectFields().size();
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/execution/ValuesResolverOneOfValidation.java
```diff
@@ -75,23 +75,36 @@ static void validateOneOfInputTypes(GraphQLType type, Object inputValue, Value<?
...
-            String msg = I18n.i18n(I18n.BundleType.Execution, locale)
-                    .msg("Execution.handleOneOfNotOneFieldError", oneOfInputType.getName());
-            throw new OneOfTooManyKeysException(msg);
+            if (objectMap.size() != 1) {
+                throwNotOneFieldError(oneOfInputType, locale);
+            }
+
+            fieldName = objectMap.keySet().iterator().next();
         }
-        String fieldName = objectMap.keySet().iterator().next();
```
felipe-gdr, line 90: The bug was in this line. We were checking the size of `argumentValue.getObjectFields()` but trying to get the first element of `objectMap`, which would be empty if no variables were passed.

russellyou: if (argumentValue instanceof ObjectValue)

we got mapSize from argumentValue. But objectMap can still be empty and we didn't check objectMap size. 

That's the problem , right?

felipe-gdr: Yep, that's right

bbakerman (approved):

PR #3439, opened 2024-02-07
Defer validation

Juliano-Prado:
I am adding Defer Validation


src/main/java/graphql/validation/rules/DeferDirectiveOnValidOperation.java
```diff
@@ -0,0 +1,56 @@
...
+import graphql.validation.AbstractRule;
+import graphql.validation.ValidationContext;
+import graphql.validation.ValidationErrorCollector;
+import java.util.List;
+import java.util.Optional;
+
+import static graphql.language.OperationDefinition.Operation.SUBSCRIPTION;
+import static graphql.validation.ValidationErrorType.MisplacedDirective;
+
+/**
```
felipe-gdr-atlassian, line 17: This type of documentation would be useful in the other rules as well.

src/main/resources/i18n/Validation.properties
```diff
@@ -98,3 +100,6 @@ ArgumentValidationUtil.handleNotObjectError=Validation error ({0}) : argument ''
 ArgumentValidationUtil.handleMissingFieldsError=Validation error ({0}) : argument ''{1}'' with value ''{2}'' is missing required fields ''{3}''
 # suppress inspection "UnusedProperty"
 ArgumentValidationUtil.handleExtraFieldError=Validation error ({0}) : argument ''{1}'' with value ''{2}'' contains a field not in ''{3}'': ''{4}''
+
+DirectiveMisplaced.notAllowedOperationRootLevel=Validation error ({0}) : Directive ''{1}'' is not allowed on root of operation ''{2}''
+DirectiveMisplaced.notAllowedOperation=Validation error ({0}) : Directive ''{1}'' is not on operation ''{2}''
```
felipe-gdr-atlassian, line 105: Should the text here say: "is not ALLOWED on operation"?

src/test/groovy/graphql/validation/rules/DeferDirectivesTest.groovy
```diff
@@ -0,0 +1,148 @@
...
+        def traversalContext = Mock(TraversalContext)
+        validationContext.getSchema() >> SpecValidationSchema.specValidationSchema
+        validationContext.getTraversalContext() >> traversalContext
+    }
+
+    def "Allow unique label directive"() {
+        given:
+        def query = """
+            query defer_query {
+              ... @defer(label: "name") {
```
felipe-gdr-atlassian, line 33: I'm missing tests that pass `label` as a variable.
Actually, validation should fail if `label` is passed as a variable.
The spec says:
```
If a label is passed, it must not be a variable
```

src/test/groovy/graphql/validation/rules/DeferDirectiveOnValidOperationTest.groovy
```diff
@@ -0,0 +1,252 @@
...
+        """
+        when:
+        traverse(query)
+
+        then:
+        errorCollector.errors.isEmpty()
+
+    }
+
+    def "Defer directive Label must be string"() {
```
felipe-gdr-atlassian, line 207: we have the exact same test in `DeferDirectivesTest.groovy`

src/test/groovy/graphql/validation/rules/DeferDirectiveOnValidOperationTest.groovy
```diff
@@ -0,0 +1,252 @@
...
+
+    }
+
+    def "Allow defer(if:false) on fragment when operation is subscription"() {
+        given:
+        def query = """
+            fragment doggo on PetMutationType {
+                ... { 
+                    dog {    
+                        ... @defer(if:false) {
```
felipe-gdr-atlassian, line 125: can we add tests where the value of `if` is passed as a variable? And also when its value is `true`?

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void checkDirective(Directive directive, List<Node> ancestors) {
+        if (!directive.getName().equals("defer") || directive.getArguments().size() == 0) {
+            return;
+        }
+        Argument labelArgument = directive.getArgument("label");
+        // argument type is validated in DeferDirectiveArgumentType
+        if (labelArgument != null && labelArgument.getValue() instanceof StringValue) {
+            if (labels.contains(((StringValue) labelArgument.getValue()).getValue())) {
```
felipe-gdr-atlassian, line 32: should we check the value of `if` before flagging a label as duplicate?
like, should this be considered valid:
```
query {
  ... @defer(if: false, label: "a") {
   a
  }
  ... @defer(if: true, label: "a") {
   anotherA
  }
}
```

src/main/resources/i18n/Validation.properties
```diff
@@ -98,3 +100,6 @@ ArgumentValidationUtil.handleNotObjectError=Validation error ({0}) : argument ''
 ArgumentValidationUtil.handleMissingFieldsError=Validation error ({0}) : argument ''{1}'' with value ''{2}'' is missing required fields ''{3}''
 # suppress inspection "UnusedProperty"
 ArgumentValidationUtil.handleExtraFieldError=Validation error ({0}) : argument ''{1}'' with value ''{2}'' contains a field not in ''{3}'': ''{4}''
+
+DirectiveMisplaced.notAllowedOperationRootLevel=Validation error ({0}) : Directive ''{1}'' is not allowed on root of operation ''{2}''
```
felipe-gdr-atlassian, line 104: Since these messages are specific to `@defer` (and in the future maybe shared with `@stream`), should we use a more explicit prefix for these i18n keys?
Maybe `IncrementalDirectiveMisplaced`?

felipe-gdr-atlassian (changes requested):

andimarek:
Can we split this up in several PRs?

src/main/java/graphql/validation/Validator.java
```diff
@@ -157,6 +160,15 @@ public List<AbstractRule> createRules(ValidationContext validationContext, Valid
...
 
+        DeferDirectiveLabel deferDirectiveLabel = new DeferDirectiveLabel(validationContext, validationErrorCollector);
+        rules.add(deferDirectiveLabel);
+
+        DeferDirectiveOnRootLevel deferDirectiveOnRootLevel = new DeferDirectiveOnRootLevel(validationContext, validationErrorCollector);
+        rules.add(deferDirectiveOnRootLevel);
+
+        DeferDirectiveOnValidOperation deferDirectiveOnValidOperation = new DeferDirectiveOnValidOperation(validationContext, validationErrorCollector);
+        rules.add(deferDirectiveOnValidOperation);
+
```
bbakerman, line 171: I think these rules should be controlled by the same graphql context control that controls execution

they should be added sure BUT inside them they should only act IF the flag is on

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
```diff
@@ -0,0 +1,41 @@
...
+@Internal
+public class DeferDirectiveLabel extends AbstractRule {
+    private Set<String> labels = new LinkedHashSet<>();
+    public DeferDirectiveLabel(ValidationContext validationContext, ValidationErrorCollector validationErrorCollector) {
+        super(validationContext, validationErrorCollector);
+    }
+
+    @Override
+    public void checkDirective(Directive directive, List<Node> ancestors) {
+        if (!directive.getName().equals("defer") || directive.getArguments().size() == 0) {
```
bbakerman, line 26: the string "defer" is repeat many times - can we use a constant from say `graphql.Directives#DEFER`

src/main/resources/i18n/Validation.properties
bbakerman: I agree these should have a specific prefix - I cant see these messages being re-used

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
bbakerman: if you pull out (StringValue) labelArgument.getValue() to a variable there will be less code later

src/main/resources/i18n/Validation.properties
```diff
@@ -98,3 +100,6 @@ ArgumentValidationUtil.handleNotObjectError=Validation error ({0}) : argument ''
...
+DirectiveMisplaced.notAllowedOperation=Validation error ({0}) : Directive ''{1}'' is not allowed on operation ''{2}''
```
bbakerman, line 105: I think this should be much more specific (eg to the rule)

`The ''{{1}}` directive is not allowed to be used on a subscription operation`

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    protected Optional<Node> getFragmentAncestor(List<Node> ancestors){
+        List<Node> ancestorsCopy = new ArrayList(ancestors);
+        Collections.reverse(ancestorsCopy);
+        return ancestorsCopy.stream().filter(node ->  !(
+                        node instanceof InlineFragment ||
+                                node instanceof SelectionSet ||
+                                node instanceof FragmentDefinition
+                )
+        ).findFirst();
```
bbakerman, line 55: this feels fragile some how

`List<Definition> definitions = getValidationContext().getDocument().getDefinitions();` will get all defs

I guess you want to know the one that this nodes specific operation

I think just go up the ancestor list and find the one that is a `OperationDefinition` - if you dont find one - it cant be a subscription

```
ancestorsCopy.stream().filter(node ->  
                        node instanceof OperationDefinition
        ).findFirst()
```

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
```diff
@@ -0,0 +1,41 @@
...
+import graphql.validation.AbstractRule;
+import graphql.validation.ValidationContext;
+import graphql.validation.ValidationErrorCollector;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+import static graphql.validation.ValidationErrorType.DuplicateArgumentNames;
+
+@Internal
```
bbakerman, line 17: can we put in links to the proposed spec PR.  Later we can update it to the real thing when it becomes real

bbakerman, line 17: can we put in links to the proposed spec PR.  Later we can update it to the real thing when it becomes real

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void checkDirective(Directive directive, List<Node> ancestors) {
+        if (directive.getName().equals("defer")){
+            Optional<Node> fragmentAncestor = getFragmentAncestor(ancestors);
+            if (fragmentAncestor.isPresent() && fragmentAncestor.get() instanceof OperationDefinition){
+                OperationDefinition operationDefinition = (OperationDefinition) fragmentAncestor.get();
+                if (OperationDefinition.Operation.MUTATION.equals(operationDefinition.getOperation()) || OperationDefinition.Operation.SUBSCRIPTION.equals(operationDefinition.getOperation())) {
+                    String message = i18n(MisplacedDirective, "DirectiveMisplaced.notAllowedOperationRootLevel", directive.getName(), operationDefinition.getOperation().name().toLowerCase());
+                    addError(MisplacedDirective, directive.getSourceLocation(), message);
```
bbakerman, line 35: This error enum seems fair enough

src/main/java/graphql/validation/rules/DeferDirectiveOnValidOperation.java
```diff
@@ -0,0 +1,56 @@
...
+        super(validationContext, validationErrorCollector);
+        this.setVisitFragmentSpreads(true);
+    }
+
+    @Override
+    public void checkDirective(Directive directive, List<Node> ancestors) {
+        if (!directive.getName().equals("defer") ||
+                (directive.getArgumentsByName().get("if") != null && !((BooleanValue) directive.getArgumentsByName().get("if").getValue()).isValue() )) {
+            return;
+        }
```
bbakerman, line 34: I am not sure we can use the `if` attr here - we dont know if its a variable value in validation and `$varName` is valid on the if.  This only handles AST booleans and this is not correct enough

bbakerman: This needs to be fixed

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
bbakerman: Hang on - this code looks nearly the same as the one rule below

What does `DeferDirectiveOnRootLevel` mean?

Can we link to the spec per rule which allows one know what is being enforced

src/main/java/graphql/validation/Validator.java
felipe-gdr: Good point 👍

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
```diff
@@ -0,0 +1,41 @@
...
+public class DeferDirectiveLabel extends AbstractRule {
```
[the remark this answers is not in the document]
Juliano-Prado: it doesn't only validate if label is unique. It also validades if label is not an ArgumentVariable

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
Juliano-Prado: I will improve the function name and result, but we have to check if the ancestor is `Operation` filtering out `inlineFragment`, `SelectionSet` and `FragmentDefinition`.Operation on bellow example is mutation, but it is valid as defer is not on root level.
`          mutation pets {
                createDog(input: {id: "1"}) {
                   ... @defer {
                        name
                    }
                }       
            }`

src/main/java/graphql/validation/rules/DeferDirectiveLabel.java
Juliano-Prado: > should we check the value of `if` before flagging a label as duplicate? like, should this be considered valid:
> 
> ```
> query {
>   ... @defer(if: false, label: "a") {
>    a
>   }
>   ... @defer(if: true, label: "a") {
>    anotherA
>   }
> }
> ```

As far as I know and checked Label must be unique regardless if value
Screenshot 2024-02-22 at 10 46 02 pm

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
Juliano-Prado: https://github.com/graphql/graphql-spec/pull/742">spec/Section 5 -- Validation.md ### Defer And Stream Directives Are Used On Valid Operations
(i've added on class javadoc

src/main/java/graphql/validation/Validator.java
Juliano-Prado: Added on each validation rule a check if ExperimentalApi.ENABLE_INCREMENTAL_SUPPORT is true

src/main/java/graphql/validation/rules/DeferDirectiveOnValidOperation.java
Juliano-Prado: Added ifArgumentMightBeFalse

src/test/groovy/graphql/validation/rules/DeferDirectiveOnValidOperationTest.groovy
Juliano-Prado: added https://github.com/graphql-java/graphql-java/pull/3439/files#diff-047a30a0c2f564205e0e9c9f70108a6e50f4e15bde07165d8a754fb8e3ea5c2cR428

src/main/java/graphql/validation/rules/DeferDirectiveOnRootLevel.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!isExperimentalApiKeyEnabled(ExperimentalApi.ENABLE_INCREMENTAL_SUPPORT)) {
+            return;
+        }
+
+        if (!Directives.DeferDirective.getName().equals(directive.getName())) {
+            return;
+        }
+        GraphQLObjectType mutationType = getValidationContext().getSchema().getMutationType();
+        GraphQLObjectType subscriptionType = getValidationContext().getSchema().getSubscriptionType();
+        GraphQLCompositeType parentType = getValidationContext().getParentType();
```
bbakerman, line 49: mutationType and subscriptionType can be null - there is no requirement to have them - so we need to null check them here please

bbakerman: well down lower I mean

bbakerman: Ahh you did - ignore me

bbakerman (approved):
nice work - its looks very clean - like I have seen it before ;)

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationFactoryDeferTest.groovy
```diff
@@ -647,34 +647,6 @@ class ExecutableNormalizedOperationFactoryDeferTest extends Specification {
         ]
     }
 
-    def "multiple fields and a multiple defers with same label are not allowed"() {
```
Juliano-Prado, line 650: validation is already catching as invalid query

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationFactoryDeferTest.groovy
```diff
@@ -861,35 +833,6 @@ class ExecutableNormalizedOperationFactoryDeferTest extends Specification {
         ]
     }
 
-    def "'if' argument with different values on same field and same label"() {
```
Juliano-Prado, line 864: validation is already catching as invalid query

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationFactoryDeferTest.groovy
```diff
@@ -647,34 +647,6 @@ class ExecutableNormalizedOperationFactoryDeferTest extends Specification {
...
-                }
-            }
-          }
-          
-        '''
-
-        Map<String, Object> variables = [:]
-
-        when:
-        executeQueryAndPrintTree(query, variables)
```
Juliano-Prado, line 671: query is set as invalid on validation, test was failing

src/test/groovy/graphql/normalized/ExecutableNormalizedOperationFactoryDeferTest.groovy
```diff
@@ -861,35 +833,6 @@ class ExecutableNormalizedOperationFactoryDeferTest extends Specification {
...
-                }
-            }
-          }
-          
-        '''
-
-        Map<String, Object> variables = [:]
-
-        when:
-        List<String> printedTree = executeQueryAndPrintTree(query, variables)
```
Juliano-Prado, line 885: query is set as invalid on validation, test was failing

felipe-gdr (approved):

PR #3440, opened 2024-02-07
Fix local maven publish & Upgrade to Gradle 8.6

dondonz:
We spotted a problem where the local Maven publish task fails on Gradle 8.5 and later. Something has changed with the `signing` plugin.

When running the publish locally task excluding signing on Gradle 8.5:
```
./gradlew publishToMavenLocal  -x signGraphqlJavaPublication
```
We got an invalid publication error
```
Invalid publication 'graphqlJava': artifact file does not exist: 'src/graphql-java/build/libs/graphql-java-0.0.0-defer-execution-2024-SNAPSHOT-sources.jar.asc'
```
It's a strange error because by skipping signing, we intentionally don't want this file there.

This PR fixes the problem by updating the signing task to only run when we're publishing (in other words, not a local `SNAPSHOT` version).

While I'm here I upgraded to Gradle 8.6. There are no accompanying gradlew etc file changes to check in.


dondonz:
I have a strange commit diff, I will reopen PR in a second

PR #3443, opened 2024-02-11
Fix invalid AST printer: add escape characters to single line descriptions

### LINKED ISSUE #3229: AstPrinter produces invalid GraphQL
**Describe the bug**
When a non-BlockString string contains escaped characters like newlines or quotes, AstPrinter does not escape them and the resulting document is not valid GraphQL.

**To Reproduce**
https://github.com/pcarrier/graphed/blob/main/src/jvmTest/kotlin/com/pcarrier/graphed/graphql/Repro1.kt

```
=== Original ===
"\"" scalar A
=== Parsed and printed ===
"""
scalar A

Exception in thread "main" graphql.parser.InvalidSyntaxException: Invalid syntax with ANTLR error 'token recognition error at: '"""\nscalar A\n'' at line 1 column 1
	at graphql.parser.Parser$1.syntaxError(Parser.java:368)
…
```

dondonz:
## This PR fixes #3229 

Add escape characters where appropriate to AST printer for single line descriptions.

#### What was the problem?
When parsing single line descriptions, escape characters are converted to their Java string equivalent. See `parseSingleQuotedString` inside `StringValueParsing`. https://github.com/graphql-java/graphql-java/blob/97ba988eabd74657f5455e9d32867c5170b0dffe/src/main/java/graphql/parser/StringValueParsing.java#L107

However, when we printed single line descriptions, we did not do the reverse and put the escape characters back into the printed output. Without the correct escape characters, it is possible to print invalid GraphQL (see #3229).

#### To do
I'll add more test cases. Currently only has a test for case described in the original issue

### DIFF
--- a/src/main/java/graphql/GraphQLError.java
+++ b/src/main/java/graphql/GraphQLError.java
@@ -39,8 +39,8 @@ public interface GraphQLError extends Serializable {
     ErrorClassification getErrorType();
 
     /**
-     * The graphql spec says that the (optional) path field of any error should be a list
-     * of path entries https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
+     * The graphql spec says that the (optional) path field of any error must be a list
+     * of path entries https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format
      *
      * @return the path in list format
      */
--- a/src/main/java/graphql/GraphQLError.java
+++ b/src/main/java/graphql/GraphQLError.java
@@ -39,8 +39,10 @@ public interface GraphQLError extends Serializable {
     ErrorClassification getErrorType();
 
     /**
-     * The graphql spec says that the (optional) path field of any error must be a list
-     * of path entries https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/GraphQLError.java
```diff
@@ -39,8 +39,10 @@ public interface GraphQLError extends Serializable {
     ErrorClassification getErrorType();
 
     /**
-     * The graphql spec says that the (optional) path field of any error should be a list
-     * of path entries https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
+     * The graphql spec says that the (optional) path field of any error must be
+     * a list of path entries starting at the root of the response
+     * and ending with the field associated with the error
+     * https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format
```
dondonz, line 45: Unrelated to AST printing, there is a minor spec change coming soon that specifies the path (if present) must be a list from root to the field. As this engine is already compliant this is only a comment change https://github.com/graphql/graphql-spec/pull/1073

bbakerman (approved):

PR #3448, opened 2024-02-13
ensure that instrumentations get an executionid

jbellenger:
👋  While migrating from v18 to v21.3, I noticed there was a regression where instrumentations would have `createState` called without an executionId.

I think this might have been introduced in ce83384b7ae682cc1eb099c10ef87e8a0daa938f. I added a fix and a test.


bbakerman (approved):
Thank you

PR #3449, opened 2024-02-14
Fix bug where deferred payloads were not using field aliases

felipe-gdr:


bbakerman (approved):
Easy to forget about aliases

andimarek (approved):

PR #3456, opened 2024-02-18
Allow empty DataLoader list in defer execution

felipe-gdr:
Support of deferred field execution with DataLoaders is still a WIP. We want to throw an exception when clients are trying to use DataLoaders with conjunction with deferred fields. The current code, however, is throwing an exception even when the list of DataLoaders is empty. This causes Spring GraphQL Java execution to blow up. My guess is that Spring GraphQL register a custom DataLoaderRegistry with an empty list of DataLoaders by default, and adds DataLoaders to the registry if the client code exposes certain beans.
This PRs handles this scenario, since it avoids throwing the error if the list of DataLoaders is empty


bbakerman (approved):

src/main/java/graphql/execution/instrumentation/dataloader/PerLevelDataLoaderDispatchStrategy.java
```diff
@@ -125,6 +119,10 @@ public void executionStrategyOnFieldValuesException(Throwable t, ExecutionStrate
 
     @Override
     public void executeObject(ExecutionContext executionContext, ExecutionStrategyParameters parameters) {
+        if (executionContext.getIncrementalCallState().getIncrementalCallsDetected()) {
```
andimarek-atlassian, line 122: why not in `executionStrategy` which happens before?

andimarek, line 122: why not in `executionStrategy` which happens before?

felipe-gdr-atlassian: Hm `executionStrategy` executes before any field is executed, and we need some fields to be executed in order to identify if `@defer` is being used.

rstoyanchev:
@felipe-gdr, I'm wondering where exactly does it fail? At the moment Spring for GraphQL checks if the registry is the `DataLoaderDispatcherInstrumentationState.EMPTY_DATALOADER_REGISTRY` instance and if so replaces it with a new instance, but otherwise performs registrations against the instance that's in place. This was a relatively recent change https://github.com/spring-projects/spring-graphql/issues/836 to allows use of custom instances. Maybe it needs some improvement still.

felipe-gdr:
@rstoyanchev the issue we're getting is that, if the `DataLoaderRegistry` in the `executionContext` is anything other than `EMPTY_DATALOADER_REGISTRY`, the execution will use a functional dispatch strategy instead of the `NO_OP` one:

https://github.com/graphql-java/graphql-java/pull/3447/files#diff-77b0941964d82259078eaf0ecfadd7e6625aa91dfe5c9195f416cab5b8f5970eR225-R232

even when there are no DataLoaders in the registry.

Normally this wouldn't be an issue. However, the `@defer` implementation is still a work-in-progress, and we haven't fully integrated the execution of deferred fields and DataLoaders. While we're in this "incomplete" state, we wanted to throw an `UnsupportedOperationException` whenever DataLoader code is called as part of an execution that includes any deferred fields. Since the functional dispatch strategy is being used (and not the NO_OP one), the exception is being thrown even when there are no DataLoaders in the registry.

Now, I'm not sure if the right thing to do. I see 3 possible options:

1. change the graphql-java code so it uses a NO_OP strategy if there are no registered DataLoaders
2. change the Spring GraphQL code, so it passes `EMPTY_DATALOADER_REGISTRY` instead of creating a new instance
3. change the defer code in graphql-java so another type of check is made when deciding whether the exception should be thrown or not

I think option 3 is the way to go, but I'm not too sure exactly how to achieve it.

@andimarek can you provide advice?

rstoyanchev:
Thanks for the extra details @felipe-gdr. We're going to make a change on our side https://github.com/spring-projects/spring-graphql/issues/915.

felipe-gdr:
That's great @rstoyanchev , I'm going to close this PR as it's no longer relevant.

PR #3478, opened 2024-02-28
Avoid repeated calls to getFieldDef and unnecessary immutable builders/collections

DanielThomas:
Avoid additional calls to `getFieldDef` in `ExecutionStrategy` and avoid immutable builder/collections where they're unnecessary  on hot paths.

This improves the throughput of `TwitterBenchmark` by about 15%:

```
Benchmark                  Mode  Cnt    Score   Error  Units
TwitterBenchmark.execute  thrpt   10  105.571 ± 0.367  ops/s
```

```
Benchmark                  Mode  Cnt    Score   Error  Units
TwitterBenchmark.execute  thrpt   10  123.840 ± 1.492  ops/s
```

### DIFF
--- a/src/main/java/graphql/execution/ExecutionStrategy.java
+++ b/src/main/java/graphql/execution/ExecutionStrategy.java
@@ -348,9 +348,9 @@ public abstract class ExecutionStrategy {
                 new InstrumentationFieldParameters(executionContext, executionStepInfo), executionContext.getInstrumentationState()
         ));
 
-        CompletableFuture<FetchedValue> fetchFieldFuture = fetchField(executionContext, parameters);
+        CompletableFuture<FetchedValue> fetchFieldFuture = fetchField(fieldDef, executionContext, parameters);
         CompletableFuture<FieldValueInfo> result = fetchFieldFuture.thenApply((fetchedValue) ->
-                completeField(executionContext, parameters, fetchedValue));
+                completeField(fieldDef, executionContext, parameters, fetchedValue));
 
         CompletableFuture<Object> fieldValueFuture = result.thenCompose(FieldValueInfo::getFieldValueFuture);
 
@@ -377,7 +377,12 @@ public abstract class ExecutionStrategy {
         MergedField field = parameters.getField();
         GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
         GraphQLFieldDefinition fieldDef = getFieldDef(executionContext.getGraphQLSchema(), parentType, field.getSingleField());
-        GraphQLCodeRegistry codeRegistry = executionContext.getGraphQLSchema().getCodeRegistry();
+        return fetchField(fieldDef, executionContext, parameters);
+    }
+
+    private CompletableFuture<FetchedValue> fetchField(GraphQLFieldDefinition fieldDef, ExecutionContext executionContext, ExecutionStrategyParameters parameters) {
+        MergedField field = parameters.getField();
+        GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
 
         // if the DF (like PropertyDataFetcher) does not use the arguments or execution step info then dont build any
 
@@ -412,6 +417,8 @@ public abstract class ExecutionStrategy {
                     .queryDirectives(queryDirectives)
                     .build();
         });
+
+        GraphQLCodeRegistry codeRegistry = executionContext.getGraphQLSchema().getCodeRegistry();
         DataFetcher<?> dataFetcher = codeRegistry.getDataFetcher(parentType, fieldDef);
 
         Instrumentation instrumentation = executionContext.getInstrumentation();
@@ -555,6 +562,11 @@ public abstract class ExecutionStrategy {
         Field field = parameters.getField().getSingleField();
         GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
         GraphQLFieldDefinition fieldDef = getFieldDef(executionContext.getGraphQLSchema(), parentType, field);
+        return completeField(fieldDef, executionContext, parameters, fetchedValue);
+    }
+
+    private FieldValueInfo completeField(GraphQLFieldDefinition fieldDef, ExecutionContext executionContext, ExecutionStrategyParameters parameters, FetchedValue fetchedValue) {
+        GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
         ExecutionStepInfo executionStepInfo = createExecutionStepInfo(executionContext, parameters, fieldDef, parentType);
 
         Instrumentation instrumentation = executionContext.getInstrumentation();
--- a/src/main/java/graphql/execution/FieldValueInfo.java
+++ b/src/main/java/graphql/execution/FieldValueInfo.java
@@ -1,5 +1,6 @@
 package graphql.execution;
 
+import com.google.common.collect.ImmutableList;
 import graphql.ExecutionResult;
 import graphql.ExecutionResultImpl;
 import graphql.PublicApi;
@@ -66,7 +67,7 @@ public class FieldValueInfo {
     public static class Builder {
         private CompleteValueType completeValueType;
         private CompletableFuture<Object> fieldValueFuture;
-        private List<FieldValueInfo> listInfos = new ArrayList<>();
+        private List<FieldValueInfo> listInfos = ImmutableList.of();
 
         public Builder(CompleteValueType completeValueType) {
             this.completeValueType = completeValueType;
--- a/src/main/java/graphql/execution/MergedSelectionSet.java
+++ b/src/main/java/graphql/execution/MergedSelectionSet.java
@@ -13,10 +13,12 @@ import java.util.Set;
 @PublicApi
 public class MergedSelectionSet {
 
-    private final ImmutableMap<String, MergedField> subFields;
+    private final Map<String, MergedField> subFields;
+    private final List<String> keys;
 
     protected MergedSelectionSet(Map<String, MergedField> subFields) {
-        this.subFields = ImmutableMap.copyOf(Assert.assertNotNull(subFields));
+        this.subFields = subFields == null ? ImmutableMap.of() : subFields;
+        this.keys =  ImmutableList.copyOf(this.subFields.keySet());
     }
 
     public Map<String, MergedField> getSubFields() {
@@ -40,7 +42,7 @@ public class MergedSelectionSet {
     }
 
     public List<String> getKeys() {
-        return ImmutableList.copyOf(keySet());
+        return keys;
     }
 
     public boolean isEmpty() {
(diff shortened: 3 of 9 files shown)


src/main/java/graphql/schema/GraphQLObjectType.java
```diff
@@ -82,9 +82,9 @@ void replaceInterfaces(List<GraphQLNamedOutputType> interfaces) {
         this.replacedInterfaces = ImmutableList.copyOf(sortTypes(interfaceComparator, interfaces));
     }
 
-    private ImmutableMap<String, GraphQLFieldDefinition> buildDefinitionMap(List<GraphQLFieldDefinition> fieldDefinitions) {
-        return ImmutableMap.copyOf(FpKit.getByName(fieldDefinitions, GraphQLFieldDefinition::getName,
-                (fld1, fld2) -> assertShouldNeverHappen("Duplicated definition for field '%s' in type '%s'", fld1.getName(), this.name)));
+    private Map<String, GraphQLFieldDefinition> buildDefinitionMap(List<GraphQLFieldDefinition> fieldDefinitions) {
+        return FpKit.getByName(fieldDefinitions, GraphQLFieldDefinition::getName,
+                (fld1, fld2) -> assertShouldNeverHappen("Duplicated definition for field '%s' in type '%s'", fld1.getName(), this.name));
```
bbakerman, line 87: yup - just checked - this does not change the immutability of what we give out since we never give out the map of fields directly

bbakerman: That said building object types is done once per schema - so this really wont affect runtime query performance at all I think - or is ImmutableMap lookup of fields somehow slower ???

src/test/java/benchmark/TwitterBenchmark.java
```diff
@@ -128,8 +132,14 @@ protected Optional<Object> getPersistedQueryId(ExecutionInput executionInput) {
...
-        System.out.println(result);
+    public static void main(String[] args) throws RunnerException {
+        Options opt = new OptionsBuilder()
+                .include("benchmark.TwitterBenchmark")
+                .forks(1)
+                .warmupIterations(5)
+                .measurementIterations(10)
+                .build();
+
+        new Runner(opt).run();
```
bbakerman, line 143: question - I noticed you put `main()` methods for running the JMH code.  Are you some how running this in a standard way (say on a controlled spec machine say?) ??

I ask because we have a desire to do this some how (we worked with Twitter on this until Elon Husk wrecked the joint)

src/main/java/graphql/execution/MergedSelectionSet.java
```diff
@@ -52,10 +54,10 @@ public static Builder newMergedSelectionSet() {
     }
 
     public static class Builder {
-        private Map<String, MergedField> subFields = ImmutableMap.of();
```
bbakerman, line 55: ok defaulting happening in constructor 👍

src/main/java/graphql/execution/FieldValueInfo.java
```diff
@@ -66,7 +67,7 @@ public String toString() {
     public static class Builder {
         private CompleteValueType completeValueType;
         private CompletableFuture<Object> fieldValueFuture;
-        private List<FieldValueInfo> listInfos = new ArrayList<>();
+        private List<FieldValueInfo> listInfos = ImmutableList.of();
```
bbakerman, line 70: I have another PR that intends to break this class away from a builder pattern

https://github.com/graphql-java/graphql-java/pull/3465

eg JUST allocate the the object - not a builder and then the object say

src/test/java/benchmark/TwitterBenchmark.java
DanielThomas: This makes it easier to iterate from the IDE and profile, setting the forks to `0`.

src/main/java/graphql/schema/GraphQLObjectType.java
DanielThomas: Yeah, `HashMap` and therefore `LinkedHashMap` compares hashes and identity equality first, so is particularly suited for `String` keys.

DanielThomas: Oh, I should add, this matters less than when I started working on this, because I found I could reduce the calls from 3 to 1, but it'll still help.

src/main/java/graphql/execution/ExecutionStrategy.java
```diff
@@ -377,7 +377,12 @@ protected CompletableFuture<FetchedValue> fetchField(ExecutionContext executionC
         MergedField field = parameters.getField();
         GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
         GraphQLFieldDefinition fieldDef = getFieldDef(executionContext.getGraphQLSchema(), parentType, field.getSingleField());
-        GraphQLCodeRegistry codeRegistry = executionContext.getGraphQLSchema().getCodeRegistry();
+        return fetchField(fieldDef, executionContext, parameters);
+    }
+
+    private CompletableFuture<FetchedValue> fetchField(GraphQLFieldDefinition fieldDef, ExecutionContext executionContext, ExecutionStrategyParameters parameters) {
+        MergedField field = parameters.getField();
+        GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
```
bbakerman, line 386: The old code is left in place

```
  protected CompletableFuture<FetchedValue> fetchField(ExecutionContext executionContext, ExecutionStrategyParameters parameters) {
        MergedField field = parameters.getField();
        GraphQLObjectType parentType = (GraphQLObjectType) parameters.getExecutionStepInfo().getUnwrappedNonNullType();
        GraphQLFieldDefinition fieldDef = getFieldDef(executionContext.getGraphQLSchema(), parentType, field.getSingleField());
        return fetchField(fieldDef, executionContext, parameters);
    }

    private CompletableFuture<FetchedValue> fetchField(GraphQLFieldDefinition fieldDef, ExecutionContext executionContext, ExecutionStrategyParameters parameters) {

```

and the engine only now calls the private method here.  This means we dont break backwards compatibility.  So nominally a 👍 

HOWEVER - we have long had the debate inside the team about JUST how much API contract we should have in the ExecutionStrategy land - eg can other compose their own engines based on using the base classes and I think the answer is... they really cant make whole new engines using this class via inheritance and hence perhaps these should not be API per say

So it might be prudent @andimarek to just change the method signature - since the old method is dangling and never called by the engine now

DanielThomas: Yeah, with the class marked `@PublicSpi` I didn't want to mess with the signatures, and three remaining references internally:

Screenshot 2024-02-28 at 5 23 59 pm

src/main/java/graphql/execution/FieldValueInfo.java
DanielThomas: I was also going to add a `SingletonMergedField` but `MergedField` is concrete and marked as a public API, so didn't feel comfortable changing that.

bbakerman:
I targetted this PR to a performance branch taken from master as of right now.

I want to double check the numbers here.  15% seems too good - so lets just double check before we merged to master

DanielThomas:
I retested here, I realised I had a memory leak on my Mac that was causing wonkiness in my first run. In fact it's closer to 25%:

```
Benchmark                  Mode  Cnt    Score   Error  Units
TwitterBenchmark.execute  thrpt   10  113.847 ± 2.123  ops/s
```

```
Benchmark                  Mode  Cnt    Score   Error  Units
TwitterBenchmark.execute  thrpt   10  149.603 ± 0.621  ops/s
```

bbakerman:
I just ran this on my machine.

Testing on personal laptops is personal and unstable - ideally we would have a controlled env to submit benchmarks but we don't.  But @dondonz was doing some innovation week work on such a thing... so we live in hope.

```
master

Benchmark                  Mode  Cnt   Score    Error  Units
TwitterBenchmark.execute  thrpt    9  75.879 ± 10.695  ops/s

performance-branch

Benchmark                  Mode  Cnt   Score   Error  Units
TwitterBenchmark.execute  thrpt    9  78.246 ± 4.922  ops/s
```

So I see a 3.1% increase  - still a win

DanielThomas:
Yeah, I can spin future work up on an `m7a.2xlarge` to benchmark. This is obviously particularly well suited for arm.

I can improve some more incidentally by interning the field keys, so you might want to try that too:
```
Benchmark                  Mode  Cnt    Score   Error  Units
TwitterBenchmark.execute  thrpt   10  158.237 ± 2.366  ops/s
```

https://github.com/DanielThomas/graphql-java/compare/dannyt/field-def...DanielThomas:graphql-java:dannyt/field-intern?expand=1

DanielThomas:
This should be better:

```
processor : 0
vendor_id : AuthenticAMD
cpu family  : 25
model   : 17
model name  : AMD EPYC 9R14
stepping  : 1
microcode : 0xa10113e
cpu MHz   : 2599.998
cache size  : 1024 KB
physical id : 0
siblings  : 8
core id   : 0
cpu cores : 8
apicid    : 0
initial apicid  : 0
fpu   : yes
fpu_exception : yes
cpuid level : 16
wp    : yes
flags   : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx mmxext fxsr_opt pdpe1gb rdtscp lm constant_tsc rep_good nopl nonstop_tsc cpuid extd_apicid aperfmperf tsc_known_freq pni pclmulqdq monitor ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm cmp_legacy cr8_legacy abm sse4a misalignsse 3dnowprefetch topoext perfctr_core invpcid_single ssbd perfmon_v2 ibrs ibpb stibp vmmcall fsgsbase bmi1 avx2 smep bmi2 invpcid avx512f avx512dq rdseed adx smap avx512ifma clflushopt clwb avx512cd sha_ni avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves avx512_bf16 clzero xsaveerptr rdpru wbnoinvd arat avx512vbmi pku ospke avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni avx512_bitalg avx512_vpopcntdq rdpid flush_l1d
bugs    : sysret_ss_attrs spectre_v1 spectre_v2 spec_store_bypass
bogomips  : 5199.99
TLB size  : 3584 4K pages
clflush size  : 64
cache_alignment : 64
address sizes : 48 bits physical, 48 bits virtual
power management:
```

```
# JMH version: 1.37
# VM version: JDK 11.0.21.0.101, OpenJDK 64-Bit Server VM, 11.0.21.0.101+2-LTS
# VM invoker: /usr/lib/jvm/zulu-11-amd64/bin/java
# VM options: <none>
# Blackhole mode: full + dont-inline hint (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 5 iterations, 5 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: benchmark.TwitterBenchmark.execute
```

```
Benchmark                  Mode  Cnt   Score   Error  Units
TwitterBenchmark.execute  thrpt    5  79.232 ± 0.500  ops/s # master
TwitterBenchmark.execute  thrpt    5  84.061 ± 1.326  ops/s # dannyt/field-def
TwitterBenchmark.execute  thrpt    5  91.914 ± 0.385  ops/s # dannyt/field-intern
```

PR #3537, opened 2024-03-19
19.x Backport PR 3525 max result nodes

dondonz:
Backports PR #3525, which enables setting a maximum number of result nodes

### DIFF
--- a/src/main/java/graphql/GraphQLContext.java
+++ b/src/main/java/graphql/GraphQLContext.java
@@ -225,6 +225,13 @@ public class GraphQLContext {
     }
 
     /**
+     * @return a new and empty graphql context object
+     */
+    public static GraphQLContext getDefault() {
+        return GraphQLContext.newContext().build();
+    }
+
+    /**
      * Creates a new GraphqlContext builder
      *
      * @return the new builder
--- a/src/main/java/graphql/execution/ExecutionStrategy.java
+++ b/src/main/java/graphql/execution/ExecutionStrategy.java
@@ -60,6 +60,7 @@ import static graphql.execution.FieldValueInfo.CompleteValueType.LIST;
 import static graphql.execution.FieldValueInfo.CompleteValueType.NULL;
 import static graphql.execution.FieldValueInfo.CompleteValueType.OBJECT;
 import static graphql.execution.FieldValueInfo.CompleteValueType.SCALAR;
+import static graphql.execution.ResultNodesInfo.MAX_RESULT_NODES;
 import static graphql.execution.instrumentation.SimpleInstrumentationContext.nonNullCtx;
 import static graphql.schema.DataFetchingEnvironmentImpl.newDataFetchingEnvironment;
 import static graphql.schema.GraphQLTypeUtil.isEnum;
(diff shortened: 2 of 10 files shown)


src/test/groovy/graphql/execution/AsyncExecutionStrategyTest.groovy
```diff
@@ -2,6 +2,7 @@ package graphql.execution
 
 import graphql.ErrorType
 import graphql.ExecutionResult
+import graphql.GraphQLContext
```
dondonz, line 5: Test changes: all tests instantiating a ExecutionContext ought to have had an empty or mocked GraphQLContext object added

In practice, during execution there is always an GraphQLContext inside the ExecutionContext, which is why we don't have a billion null checks for this in the engine. These tests have been brought up to date

In this file it's slightly different to the next file, because in master this file has an adaptation for the new defer feature. My intention was to keep this as close to master as possible

src/main/java/graphql/GraphQLContext.java
```diff
@@ -224,6 +224,13 @@ public static GraphQLContext of(Consumer<GraphQLContext.Builder> contextBuilderC
         return of(builder.map);
     }
 
+    /**
+     * @return a new and empty graphql context object
+     */
+    public static GraphQLContext getDefault() {
+        return GraphQLContext.newContext().build();
+    }
```
dondonz, line 232: This convenience method was added in a later version - backported to help with this PR

src/main/java/graphql/execution/FetchedValue.java
```diff
@@ -19,7 +19,7 @@ public class FetchedValue {
     private final Object localContext;
     private final ImmutableList<GraphQLError> errors;
 
-    private FetchedValue(Object fetchedValue, Object rawFetchedValue, ImmutableList<GraphQLError> errors, Object localContext) {
```
dondonz, line 22: Changed visibility to be same as master

src/main/java/graphql/execution/FieldValueInfo.java
```diff
@@ -25,7 +25,7 @@ public enum CompleteValueType {
     private final CompletableFuture<ExecutionResult> fieldValue;
     private final List<FieldValueInfo> fieldValueInfos;
 
-    private FieldValueInfo(CompleteValueType completeValueType, CompletableFuture<ExecutionResult> fieldValue, List<FieldValueInfo> fieldValueInfos) {
```
dondonz, line 28: Changed visibility to be same as master

src/main/java/graphql/execution/ExecutionStrategy.java
```diff
@@ -237,8 +238,23 @@ protected CompletableFuture<FetchedValue> fetchField(ExecutionContext executionC
...
-        GraphQLCodeRegistry codeRegistry = executionContext.getGraphQLSchema().getCodeRegistry();
+    private CompletableFuture<FetchedValue> fetchField(GraphQLFieldDefinition fieldDef, ExecutionContext executionContext, ExecutionStrategyParameters parameters) {
+
+        int resultNodesCount = executionContext.getResultNodesInfo().incrementAndGetResultNodesCount();
+
+        Integer maxNodes;
+        if ((maxNodes = executionContext.getGraphQLContext().get(MAX_RESULT_NODES)) != null) {
+            if (resultNodesCount > maxNodes) {
+                executionContext.getResultNodesInfo().maxResultNodesExceeded();
+                return CompletableFuture.completedFuture(new FetchedValue(null, null, ImmutableKit.emptyList(), null));
```
dondonz, line 252: FetchedValue constructor (and object) slightly different here vs master - there is one more field. But the meaning is the same, return early with a null

src/main/java/graphql/execution/ExecutionStrategy.java
```diff
@@ -266,6 +281,7 @@ protected CompletableFuture<FetchedValue> fetchField(ExecutionContext executionC
                 .queryDirectives(queryDirectives)
                 .build();
 
+        GraphQLCodeRegistry codeRegistry = executionContext.getGraphQLSchema().getCodeRegistry();
```
dondonz, line 284: Line was moved down

src/main/java/graphql/execution/ExecutionStrategy.java
```diff
@@ -528,6 +544,15 @@ protected FieldValueInfo completeValueForList(ExecutionContext executionContext,
         List<FieldValueInfo> fieldValueInfos = new ArrayList<>(size.orElse(1));
         int index = 0;
         for (Object item : iterableValues) {
+            int resultNodesCount = executionContext.getResultNodesInfo().incrementAndGetResultNodesCount();
+            Integer maxNodes;
+            if ((maxNodes = executionContext.getGraphQLContext().get(MAX_RESULT_NODES)) != null) {
+                if (resultNodesCount > maxNodes) {
+                    executionContext.getResultNodesInfo().maxResultNodesExceeded();
+                    return new FieldValueInfo(NULL, completedFuture(ExecutionResultImpl.newExecutionResult().build()), fieldValueInfos);
```
dondonz, line 552: Difference to master - must have a completed value of type ExecutionResult. Can't return a plain null here or the engine is going to hang because prior to forthcoming v22 changes, the engine previously always expected an ExecutionResult

v19 is different to the other backports. This builder moved files after v19

PR #3543, opened 2024-03-25
19.x Backport PR 3539

dondonz:
Backport #3539 and bring a number of related files up to date with v21.x, so that we minimise API changes between files connected to ExecutableNormalizedOperationFactory across v19/20/21

Note: introspection node count is 1 less in v19 because the new `@oneOf` directive is only included in v21 onwards


src/test/java/benchmark/BenchmarkUtils.java
```diff
@@ -1,19 +1,32 @@
...
             URL resource = BenchmarkUtils.class.getClassLoader().getResource(name);
-            return String.join("\n", Files.readLines(new File(resource.toURI()), Charset.defaultCharset()));
-
+            if (resource == null) {
+                throw new IllegalArgumentException("missing resource: " + name);
+            }
+            byte[] bytes;
+            try (InputStream inputStream = resource.openStream()) {
+                // In GraphQL Java version 21 and above, this Guava helper is replaced with Java 9's readAllBytes()
+                bytes = ByteStreams.toByteArray(inputStream);
```
dondonz, line 27: Fun one: readAllBytes is only in Java 9 onwards so here's the Guava equivalent. We selectively shade in Guava files (e.g. Immutable objects already in use) so we're alright to use Guava here.

bbakerman (approved):

PR #3553, opened 2024-04-01
Stricter parseValue coercion for String, Int, Float, Boolean

dondonz:
This PR (re)introduces stricter `parseValue` coercion for String, Int, Float, and Boolean to align with the [reference JS implementation](https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts). 

This is a breaking change. Here's what's changing:
* String `parseValue` now requires input of type `String`. For example, a Number input `123` or a Boolean input `true` will no longer be accepted.
* Boolean `parseValue` now requires input of type `Boolean`. For example, a String input `"true"` will no longer be accepted.
* Float `parseValue` now requires input of type `Number`. For example, a String input `"3.14"` will no longer be accepted.
* Int `parseValue` now requires input of type `Number`. For example, a String input `"42"` will no longer be accepted.

There's a little history to this PR. We previously introduced stricter `parseValue` coercion #3030 #3042 in version [20.0](https://github.com/graphql-java/graphql-java/releases/tag/v20.0) but received feedback that we needed a better migration pathway for this breaking change. We then reverted stricter `parseValue` coercion #3186 in [20.3](https://github.com/graphql-java/graphql-java/releases/tag/v20.3) and [21.0](https://github.com/graphql-java/graphql-java/releases/tag/v21.0).

In [version 21.0](https://github.com/graphql-java/graphql-java/releases/tag/v21.0), we introduced the `InputInterceptor` #3188 (and an implementation `LegacyCoercingInputInterceptor` #3218) to provide a migration pathway. You can use this interceptor to monitor and modify values.

Now we're about to release our next breaking release change, version 22.0. We are now re-introducing stricter `parseValue` coercion to realign with the JS reference implementation.

See the JS implementation of scalar coercion: https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts


src/main/java/graphql/scalar/GraphqlBooleanCoercing.java
```diff
@@ -129,7 +128,7 @@ public Boolean parseLiteral(@NotNull Object input) {
 
     @Override
     @Deprecated
-    public Value valueToLiteral(@NotNull Object input) {
+    public @NotNull Value<?> valueToLiteral(@NotNull Object input) {
```
dondonz, line 131: IDEA complained so I'm fixing these up while I'm here

src/main/java/graphql/scalar/GraphqlBooleanCoercing.java
```diff
@@ -67,13 +67,12 @@ private Boolean serializeImpl(@NotNull Object input, @NotNull Locale locale) {
```
dondonz, line 67: This PR is a reversion of PR #3186, plus little bits of tidying up

src/test/groovy/graphql/ScalarsBooleanTest.groovy
```diff
@@ -55,19 +55,19 @@ class ScalarsBooleanTest extends Specification {
...
         where:
-        value                        | result
-        true                         | true
-        "false"                      | false
-        "true"                       | true
-        "True"                       | true
-        0                            | false
-        1                            | true
-        -1                           | true
-        new Long(42345784398534785l) | true
```
dondonz, line 66: A bunch of constructors were deprecated in Java 9. Now we're on Java 11 let's update these

bbakerman (approved):

PR #3562, opened 2024-04-09
fix incremental partial result builder

sbarker2:
it looks like there was a copy-paste error for this one function (copied from the hasNext part of the builder).

This change is to fix the function to allow us to properly add "extensions" to the builder. Similar to this one [here](https://github.com/graphql-java/graphql-java/blob/e01013f65f4554643e8b7c3df5143438fc5d414a/src/main/java/graphql/incremental/IncrementalPayload.java#L135C1-L138C10)

have also added a test for this builder

### DIFF
--- a/src/test/groovy/graphql/incremental/IncrementalExecutionResultTest.groovy
+++ b/src/test/groovy/graphql/incremental/IncrementalExecutionResultTest.groovy
@@ -40,6 +40,7 @@ class IncrementalExecutionResultTest extends Specification {
                 ])
                 .hasNext(true)
                 .incremental([defer1, stream1, stream2])
+                .extensions([some: "map"])
                 .build()
 
         def toSpec = result.toSpecification()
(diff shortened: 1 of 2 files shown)


felipe-gdr (approved):

andimarek:
Can you please add a small test to show what the fix did?

sbarker2:
> Can you please add a small test to show what the fix did?

Sure @andimarek .
 
For context, this change is to fix the function that allows us to add "extensions" to the builder if it comes through the payload. Similar to this one [here](https://github.com/graphql-java/graphql-java/blob/e01013f65f4554643e8b7c3df5143438fc5d414a/src/main/java/graphql/incremental/IncrementalPayload.java#L135C1-L138C10)
  
turns out there was already a sanity test for the builder, but it just hadn't included the `extensions` part. I've added it in now

src/test/groovy/graphql/incremental/IncrementalExecutionResultTest.groovy
```diff
@@ -40,13 +40,15 @@ class IncrementalExecutionResultTest extends Specification {
                 ])
                 .hasNext(true)
                 .incremental([defer1, stream1, stream2])
+                .extensions([some: "map"])
```
sbarker2, line 43: this is a similar `extensions` function for another class that I've added as well

felipe-gdr (approved):

bbakerman (approved):

PR #3582, opened 2024-04-29
Add nullability annotations in DataFetchingEnvironment & sub-classes

Salzian:
While using Netflix' DGS with Kotlin, we ran into an issue where it was not immediately clear that `DataFetchingEnvironment.getLocalContext` can return `null`. After reading the documentation and the rest of the `DataFetchingEnvironment` interface, I noticed the lack of nullability annotations for Kotlin interoperability. However, I also saw that other classes in this project already use the Jetbrains annotations. 

In this PR I added nullability annotations to all methods of the `DataFetchingEnvironment` interface that had mentions about nullability in their documentation. For patching the whole class I currently lack the overview of the bigger picture of this interface.

I followed through on sub-classes that override these methods.

### DIFF
--- a/src/main/java/graphql/schema/DataFetchingEnvironment.java
+++ b/src/main/java/graphql/schema/DataFetchingEnvironment.java
@@ -13,6 +13,8 @@ import graphql.language.FragmentDefinition;
 import graphql.language.OperationDefinition;
 import org.dataloader.DataLoader;
 import org.dataloader.DataLoaderRegistry;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


bbakerman (approved):
Thanks for this

juriad:
IMHO this should not have been included in a minor version. Adding `@Nullable` annotation to a method is a breaking change for Kotlin code. We are getting errors like:
```
> Task :dmm-dq:compileKotlin FAILED
e: file:///builds/engineering/.../validation/onthefly/OnTheFlyValidationGqlOperation.kt:32:65 Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type '@Nullable() com...graphql.types.node.NodeWrapperBase?'.
```
caused by this line:
```
catalogItemId = env.getSource<NodeWrapperBase>().gid,
```

hpuac:
@juriad Lovely, the compiler now protects you against a possible NPE that you did not check in your code before, amazing!
I agree with the minor version argument though, including it in the next major release would have been fine as well 🙂

PR #3616, opened 2024-06-04
initial test case

jbellenger:
This test case fails with:
```
Condition not satisfied:

typeB.sourceLocation.sourceName == "b"
|     |              |          |
|     |              a          false
|     |                         1 difference (0% similarity)
|     |                         (a)
|     |                         (b)
|     SourceLocation{line=1, column=25, sourceName=a}
ObjectTypeDefinition{name='B', implements=[], directives=[], fieldDefinitions=[FieldDefinition{name='field', type=TypeName{name='String'}, inputValueDefinitions=[], directives=[]}]}

Condition not satisfied:

typeB.sourceLocation.sourceName == "b"
|     |              |          |
|     |              a          false
|     |                         1 difference (0% similarity)
|     |                         (a)
|     |                         (b)
|     SourceLocation{line=1, column=25, sourceName=a}
ObjectTypeDefinition{name='B', implements=[], directives=[], fieldDefinitions=[FieldDefinition{name='field', type=TypeName{name='String'}, inputValueDefinitions=[], directives=[]}]}

	at graphql.parser.MultiSourceReaderTest.can read sdl types when newlines are missing(MultiSourceReaderTest.groovy:128)


```

### DIFF
--- a/src/test/groovy/graphql/parser/MultiSourceReaderTest.groovy
+++ b/src/test/groovy/graphql/parser/MultiSourceReaderTest.groovy
@@ -1,5 +1,6 @@
 package graphql.parser
 
+import graphql.schema.idl.SchemaParser
 import spock.lang.Specification
 
 class MultiSourceReaderTest extends Specification {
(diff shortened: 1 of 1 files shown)


dondonz:
Hi @jbellenger !

Could you let me know a little more about the background to this case you've found? Is this a situation you encounter at work where schemas are appearing on one line (no new line character), or was it something you found via fuzz testing, or something else? I'd like to better understand how you came about this edge case.

dondonz:
Hello @jbellenger, could you explain more about the context of this test case? Do you often have schemas all on 1 line?

Asking so I can assess if it's worth a change to the MultiSourceReader

PR #3619, opened 2024-06-05
AstPrinter: Empty types should not include braces `{}`

tinnou:
## Description
In the latest version of the [spec](https://spec.graphql.org/draft/#ObjectTypeDefinition), empty types (input objects, object types) must be represented without curly braces, i.e the following is valid:
```graphql
type Query 
# below only here to make the sdl valid 
extend type Query {
  foo: String
}
```
while the following technically isn't (note the `{}`):
```graphql
type Query {}
# below only here to make the sdl valid 
extend type Query {
  foo: String
}
```
Because of historical reasons, `graphql-java` supports parsing of both syntaxes. However its `AstPrinter` prints the spec invalid form. While this is ok in systems where `graphql-java`'s GraphQL parser is the only one in use, it fails in poly-parser systems where stricter (albeit spec-compliant) parsers might fail to parse `graphql-java` printed SDL output.

## Approach
This PR updates the `AstPrinter` to be spec-compliant, i.e serialize empty types, empty selection sets (although should never happen) without the `{}`.

## Follow ups
If there is interest, I can follow up with another PR that (while breaking backwards-compatibilty) will make the parser spec compliant as well, i.e only allow the form without braces `{}`.

### DIFF
--- a/src/main/java/graphql/language/AstPrinter.java
+++ b/src/main/java/graphql/language/AstPrinter.java
@@ -619,7 +619,7 @@ public class AstPrinter {
 
     private <T extends Node> String block(List<T> nodes) {
         if (isEmpty(nodes)) {
-            return "{}";
+            return "";
         }
         if (compactMode) {
             String joinedNodes = joinTight(nodes, " ", "", "");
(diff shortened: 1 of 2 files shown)


bbakerman (approved):
Great to see you contributing again :)

tinnou:
> Great to see you contributing again :)

🙏  but really it's all you and the other core contributors that keep this library in tip-top shape!
@dondonz @andimarek 🤟

dondonz (approved):
Thanks for the PR!

PR #3621, opened 2024-06-06
@defer: Early defer calls

felipe-gdr:
Start the execution of deferred fields as soon as they are identified, instead of just after the initial response is complete. This will result in lower overall latency for executions that have deferred fields.

### DIFF
--- a/src/main/java/graphql/execution/ExecutionStrategy.java
+++ b/src/main/java/graphql/execution/ExecutionStrategy.java
@@ -550,7 +550,14 @@ public abstract class ExecutionStrategy {
 
         if (result instanceof DataFetcherResult) {
             DataFetcherResult<?> dataFetcherResult = (DataFetcherResult<?>) result;
-            executionContext.addErrors(dataFetcherResult.getErrors());
+
+            if (parameters.getField().isDeferred()) {
+                parameters.getDeferredCallContext().addErrors(dataFetcherResult.getErrors());
+            } else {
+                // Errors from a deferred field are handled in the deferred execution strategy
+                executionContext.addErrors(dataFetcherResult.getErrors());
+            }
+
             addExtensionsIfPresent(executionContext, dataFetcherResult);
 
             Object localContext = dataFetcherResult.getLocalContext();
@@ -585,12 +592,6 @@ public abstract class ExecutionStrategy {
                 .dataFetchingEnvironment(environment)
                 .exception(e)
                 .build();
-
-        parameters.getDeferredCallContext().onFetchingException(
-                parameters.getPath(),
-                parameters.getField().getSingleField().getSourceLocation(),
-                e
-        );
 
         try {
             return asyncHandleException(dataFetcherExceptionHandler, handlerParameters);
@@ -714,9 +715,12 @@ public abstract class ExecutionStrategy {
 
     private void handleUnresolvedTypeProblem(ExecutionContext context, ExecutionStrategyParameters parameters, UnresolvedTypeException e) {
         UnresolvedTypeError error = new UnresolvedTypeError(parameters.getPath(), parameters.getExecutionStepInfo(), e);
-        context.addError(error);
 
-        parameters.getDeferredCallContext().onError(error);
+        if (parameters.getField().isDeferred()) {
+            parameters.getDeferredCallContext().addError(error);
+        } else {
+            context.addError(error);
+        }
     }
 
     /**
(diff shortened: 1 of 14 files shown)


src/main/java/graphql/execution/ExecutionStrategy.java
```diff
@@ -997,9 +991,25 @@ protected Iterable<Object> toIterable(ExecutionContext context, ExecutionStrateg
...
     private void handleTypeMismatchProblem(ExecutionContext context, ExecutionStrategyParameters parameters) {
         TypeMismatchError error = new TypeMismatchError(parameters.getPath(), parameters.getExecutionStepInfo().getUnwrappedNonNullType());
-        context.addError(error);
 
-        parameters.getDeferredCallContext().onError(error);
+        addErrorToRightContext(error, parameters, context);
+    }
+
+    // Errors from a deferred field are handled in the deferred execution strategy
+    private static void addErrorToRightContext(GraphQLError error, ExecutionStrategyParameters parameters, ExecutionContext executionContext) {
```
felipe-gdr, line 999: I had to change the way we add errors to the `ExecutionContext`. 

Errors resulting from executing deferred fields were being added to the main `ExecutionContext`. This behaviour didn't result in a visible error before because we resolved the full initial response before starting the execution of any deferred field. After the execution order changed, errors from deferred fields were ending up in the initial response.

This PR changes this, so now, errors from deferred fields are only added to the `DeferredCallContext`.

src/main/java/graphql/execution/incremental/DeferredExecutionSupport.java
```diff
@@ -155,25 +153,24 @@ private Supplier<CompletableFuture<DeferredFragmentCall.FieldWithExecutionResult
...
-                                        );
-
-                                return executionResultCF
-                                        .thenApply(executionResult ->
-                                                new DeferredFragmentCall.FieldWithExecutionResult(currentField.getResultKey(), executionResult)
-                                        );
-                            }
-                    )
+                    // we cache the field resolution to avoid multiple calls to the same data fetcher
+                    key -> {
```
felipe-gdr, line 157: This is the main change - instead of creating suppliers that were being called just once the initial response was complete, we are executing the fields as they come and adding the CompletableFutures to the queue in `IncrementalCallState`

src/main/java/graphql/execution/incremental/IncrementalCallState.java
```diff
@@ -9,21 +9,32 @@
...
 public class IncrementalCallState {
     private final AtomicBoolean incrementalCallsDetected = new AtomicBoolean(false);
-    private final Deque<IncrementalCall<? extends IncrementalPayload>> incrementalCalls = new ConcurrentLinkedDeque<>();
+
+    // PriorityBlockingQueue doesn't have a constructor that accepts just the comparator and
+    // uses the default initial capacity. 11 is the default value used in the implementation I'm looking at,
+    // and it seems reasonable for our use case.
+    private static final int QUEUE_INITIAL_CAPACITY = 11;
+    private final Queue<IncrementalCall<? extends IncrementalPayload>> incrementalCalls =
+            new PriorityBlockingQueue<>(QUEUE_INITIAL_CAPACITY, Comparator.comparingInt(IncrementalCall::level));
```
felipe-gdr, line 36: Priority queue ordered by field level to ensure fields in higher levels don't resolve before fields in lower levels. This is a possibility when we have nested defers, and could be reproduced in a test.

src/test/groovy/graphql/execution/incremental/DeferredCallTest.groovy
```diff
@@ -82,38 +81,28 @@ class DeferredCallTest extends Specification {
         ]
     }
 
-    private static Supplier<CompletableFuture<DeferredFragmentCall.FieldWithExecutionResult>> createResolvedFieldCall(
```
felipe-gdr, line 85: Just getting rid of the `Supplier` that was wrapping the CF in these tests...

src/main/java/graphql/execution/ExecutionStrategy.java
felipe-gdr: We could also add the deferred errors to the `ExecutionContext`, marking them as such, so they would be ignored when building the initial response.
By doing this, one could always check `ExecutionContext` to get a comprehensive list of all the errors that happened during the execution - not sure how useful that would be though. ns.

src/main/java/graphql/execution/ExecutionStrategyParameters.java
```diff
@@ -3,6 +3,7 @@
 import graphql.PublicApi;
 import graphql.execution.incremental.DeferredCallContext;
 
+import javax.annotation.Nullable;
```
dondonz, line 6: Nit: can you please change it to the jetbrains Nullable annotation? We use the jetbrains version in other places

felipe-gdr:
closing this, as it makes more sense to start execution of deferred fields only when the initial response has been fully resolved.

PR #3639, opened 2024-06-20
Defer dataloader integration

felipe-gdr:
This is a proposal for a quick and dirty solution for integrating `@defer` and data loaders, which will unblock further work.
This implementation will not try to optimise dispatches at all for deferred fields. It will simply always dispatch, which will inevitably result in more dispatches than necessary.
Dispatches for non-deferred fields will behave the same as before - i.e. in an optimal way.


src/test/groovy/graphql/execution/instrumentation/dataloader/DataLoaderPerformanceWithChainedInstrumentationTest.groovy
```diff
@@ -126,73 +119,4 @@ class DataLoaderPerformanceWithChainedInstrumentationTest extends Specification
         incrementalSupport << [true, false]
     }
 
-    def "chainedInstrumentation: data loader will not work with deferred queries"() {
```
felipe-gdr, line 129: Moved defer+data loader tests to its own file

src/test/groovy/graphql/execution/instrumentation/dataloader/DataLoaderPerformanceData.groovy
```diff
@@ -192,159 +209,30 @@ class DataLoaderPerformanceData {
...
+                    ... @defer(if: $deferredEnabled) {
+                        expensiveDepartments { 
                             name 
-                        } 
-                    } 
-                } 
-            }
-            """
-
-    static def expectedInitialDeferredData = [
```
felipe-gdr, line 212: Moved the defer tests into their own test class

PR #3640, opened 2024-06-24
Fix bug with error handling in the defer execution code

felipe-gdr:
Improvements in the error handling for deferred code.
- fixes bug where errors added by data fetchers that returned `DataFetcherResult` where not being added to deferred payloads
- Errors generated during the execution of deferred fields are kept in the correspondent `DeferredCallContext` only, no longer being duplicated in `ExecutionContext`

### DIFF
--- a/src/main/java/graphql/execution/ExecutionStrategyParameters.java
+++ b/src/main/java/graphql/execution/ExecutionStrategyParameters.java
@@ -2,6 +2,7 @@ package graphql.execution;
 
 import graphql.PublicApi;
 import graphql.execution.incremental.DeferredCallContext;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.function.Consumer;
 
@@ -71,8 +72,38 @@ public class ExecutionStrategyParameters {
         return parent;
     }
 
+    /**
+     * Returns the deferred call context if we're in the scope of a deferred call.
+     * A new DeferredCallContext is created for each @defer block, and is passed down to all fields within the deferred call.
+     *
+     * <pre>
+     *     query {
+     *        ... @defer {
+     *            field1 {        # new DeferredCallContext created here
+     *                field1a     # DeferredCallContext passed down to this field
+     *            }
+     *        }
+     *
+     *        ... @defer {
+     *            field2          # new DeferredCallContext created here
+     *        }
+     *     }
+     * </pre>
+     *
+     * @return the deferred call context or null if we're not in the scope of a deferred call
+     */
+    @Nullable
     public DeferredCallContext getDeferredCallContext() {
         return deferredCallContext;
+    }
+
+    /**
+     * Returns true if we're in the scope of a deferred call.
+     *
+     * @return true if we're in the scope of a deferred call
+     */
+    public boolean isInDeferredContext() {
+        return deferredCallContext != null;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/graphql/execution/incremental/DeferredCallContext.java
```diff
@@ -23,12 +20,11 @@ public class DeferredCallContext {
 
     private final List<GraphQLError> errors = new CopyOnWriteArrayList<>();
 
-    public void onFetchingException(ResultPath path, SourceLocation sourceLocation, Throwable throwable) {
-        ExceptionWhileDataFetching error = new ExceptionWhileDataFetching(path, throwable, sourceLocation);
-        onError(error);
+    public void addErrors(List<GraphQLError> errors) {
```
felipe-gdr, line 23: keep naming consistent with `ExecutionContext`

Juliano-Prado (approved):

src/main/java/graphql/execution/ExecutionStrategyParameters.java
```diff
@@ -187,9 +218,6 @@ public Builder deferredCallContext(DeferredCallContext deferredCallContext) {
         }
 
         public ExecutionStrategyParameters build() {
-            if (deferredCallContext == null) {
-                deferredCallContext = new DeferredCallContext();
```
felipe-gdr, line 191: There was no point in creating a new instance of `DeferredCallContext` by default.
With the current change, the presence of an instance of `DeferredCallContext` signifies that the execution is part of a deferred call.

src/main/java/graphql/execution/NonNullableFieldValidator.java
```diff
@@ -46,8 +47,15 @@ public <T> T validate(ResultPath path, T result) throws NonNullableFieldWasNullE
...
+                final ResultPath path = parameters.getPath();
+
                 NonNullableFieldWasNullException nonNullException = new NonNullableFieldWasNullException(executionStepInfo, path);
-                executionContext.addError(new NonNullableFieldWasNullError(nonNullException), path);
+                final GraphQLError error = new NonNullableFieldWasNullError(nonNullException);
+                if(parameters.getDeferredCallContext() != null) {
+                    parameters.getDeferredCallContext().addError(error);
+                } else {
+                    executionContext.addError(error, path);
+                }
```
gnawf, line 58: I wonder if there's a better way to do this, because this is fairly error prone.

i.e. if you're adding an error you have to remember to check for the presence of a deferred context first.

Could we clone the `ExecutionContext` for a defer and have a separate error array?

ofc this is only a problem for framework code, and end users should be ok… unless they invoke `ExecutionContext.addError` themselves? Seems like it's public API?

gnawf (approved):
Left a comment, but this will work too

src/main/java/graphql/execution/NonNullableFieldValidator.java
felipe-gdr: I totally agree about it being high error-prone, but I couldn't figure out a simple solution. As long as `ExecutionContext#addError` exists, any code can always add errors directly to `ExecutionContext` without checking for defer...

> Could we clone the ExecutionContext for a defer and have a separate error array?
hm... `DeferredCallContext` is sort of that. Is a context but only with the state that is relevant for a defer block, which for now is just a list of errors.
Can you expand on your suggestion? :-)

sbarker2 (approved):

PR #3656, opened 2024-07-04
Add missing directive definitions

dondonz:
Adds DirectiveDefinitions that were missing on built-in directives `@skip`, `@include`, and`@defer`

We didn't previously have any tests for DirectiveDefinitions and I can see why, it's a definition rather than logic. But if you want tests I can add them.

### DIFF
--- a/src/main/java/graphql/Directives.java
+++ b/src/main/java/graphql/Directives.java
@@ -31,16 +31,25 @@ import static graphql.schema.GraphQLNonNull.nonNull;
 @PublicApi
 public class Directives {
 
-    private static final String SPECIFIED_BY = "specifiedBy";
     private static final String DEPRECATED = "deprecated";
+    private static final String INCLUDE = "include";
+    private static final String SKIP = "skip";
+    private static final String SPECIFIED_BY = "specifiedBy";
     private static final String ONE_OF = "oneOf";
     private static final String DEFER = "defer";
 
-    public static final String NO_LONGER_SUPPORTED = "No longer supported";
     public static final DirectiveDefinition DEPRECATED_DIRECTIVE_DEFINITION;
+    public static final DirectiveDefinition INCLUDE_DIRECTIVE_DEFINITION;
+    public static final DirectiveDefinition SKIP_DIRECTIVE_DEFINITION;
     public static final DirectiveDefinition SPECIFIED_BY_DIRECTIVE_DEFINITION;
     @ExperimentalApi
     public static final DirectiveDefinition ONE_OF_DIRECTIVE_DEFINITION;
+    @ExperimentalApi
+    public static final DirectiveDefinition DEFER_DIRECTIVE_DEFINITION;
+
+    public static final String BOOLEAN = "Boolean";
+    public static final String STRING = "String";
+    public static final String NO_LONGER_SUPPORTED = "No longer supported";
 
     static {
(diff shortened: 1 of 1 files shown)


src/main/java/graphql/Directives.java
```diff
@@ -31,16 +31,25 @@
...
+    public static final DirectiveDefinition SKIP_DIRECTIVE_DEFINITION;
     public static final DirectiveDefinition SPECIFIED_BY_DIRECTIVE_DEFINITION;
     @ExperimentalApi
     public static final DirectiveDefinition ONE_OF_DIRECTIVE_DEFINITION;
+    @ExperimentalApi
+    public static final DirectiveDefinition DEFER_DIRECTIVE_DEFINITION;
+
+    public static final String BOOLEAN = "Boolean";
+    public static final String STRING = "String";
+    public static final String NO_LONGER_SUPPORTED = "No longer supported";
```
dondonz, line 52: In more recent code, as a general rule we prefer to not have random strings hanging about. Some parts of this class haven't been changed in a long while and still have strings. If you prefer I can give the `public static final String` treatment to all the strings in this class

src/main/java/graphql/Directives.java
```diff
@@ -54,11 +63,39 @@ public class Directives {
...
+                .directiveLocation(newDirectiveLocation().name(FRAGMENT_SPREAD.name()).build())
+                .directiveLocation(newDirectiveLocation().name(INLINE_FRAGMENT.name()).build())
+                .directiveLocation(newDirectiveLocation().name(FIELD.name()).build())
+                .description(createDescription("Directs the executor to include this field or fragment only when the `if` argument is true"))
+                .inputValueDefinition(
+                        newInputValueDefinition()
+                                .name("if")
+                                .description(createDescription("Included when true."))
+                                .type(newNonNullType(newTypeName().name(BOOLEAN).build()).build())
+                                .build())
```
dondonz, line 82: A comment for myself: in AST land, DirectiveDefinitions ONLY have "inputValueDefinition", this is the AST type used to model "arguments". Confusingly there is another `graphql.language`/AST class called "Argument". It is meant to be `inputValueDefinition` here.

andimarek:
Do we need to declare the defer one as experimental?

src/main/java/graphql/Directives.java
```diff
@@ -31,16 +31,25 @@
...
 
-    public static final String NO_LONGER_SUPPORTED = "No longer supported";
     public static final DirectiveDefinition DEPRECATED_DIRECTIVE_DEFINITION;
+    public static final DirectiveDefinition INCLUDE_DIRECTIVE_DEFINITION;
+    public static final DirectiveDefinition SKIP_DIRECTIVE_DEFINITION;
     public static final DirectiveDefinition SPECIFIED_BY_DIRECTIVE_DEFINITION;
     @ExperimentalApi
     public static final DirectiveDefinition ONE_OF_DIRECTIVE_DEFINITION;
+    @ExperimentalApi
+    public static final DirectiveDefinition DEFER_DIRECTIVE_DEFINITION;
```
dondonz, line 48: > Do we need to declare the defer one as experimental?

@andimarek it is declared here, did you also want the annotation to appear elsewhere?

Elsewhere all the defer codepaths have been annotated with `@ExperimentalApi`

andimarek: perfect, didn't see that

andimarek (approved):

PR #3693, opened 2024-08-23
transformable IncrementalExecutionResult

jbellenger:
Calling `IncrementalExecutionResult.transform` will return an `ExecutionResultImpl` (rather than another `IncrementalExecutionResult`), which can cause incremental parts of the execution result to get lost.

This makes it difficult to add `@defer` support to a system with existing instrumentations, which might modify an ExecutionResult without knowing that it needs to be handled differently if it is incremental.

This PR overrides the implementation of `IncrementalExecutionResult.transform` to ensure that the transformed result retains its incremental nature.

### DIFF
--- a/src/main/java/graphql/incremental/IncrementalExecutionResultImpl.java
+++ b/src/main/java/graphql/incremental/IncrementalExecutionResultImpl.java
@@ -10,6 +10,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 @ExperimentalApi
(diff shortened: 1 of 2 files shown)


src/main/java/graphql/incremental/IncrementalExecutionResultImpl.java
```diff
@@ -52,6 +53,13 @@ public static Builder fromExecutionResult(ExecutionResult executionResult) {
         return new Builder().from(executionResult);
     }
 
+    @Override
+    public IncrementalExecutionResult transform(Consumer<ExecutionResult.Builder<?>> builderConsumer) {
```
jbellenger, line 57: It would be nice if this Consumer was actually on type `IncrementalExecutionResult.Builder`, but changing it would either break the override, or would require a lot of generics to be added to ExecutionResult.

As written, the builder given to the consumer can be cast to `IncrementalExecutionResult.Builder`. I'm open to feedback/ideas on if there's a better way to have nicer types.

felipe-gdr: Good point.
That's a tricky problem though, and I can't think of a nice solution right now.
I'm happy to merge the PR as is if it unblocks you.

felipe-gdr (approved):

dondonz:
Hi @jbellenger thanks for the fix! Later today I will release v22.3 which will include your PR, in case you need a point version release for work.

PR #3797, opened 2025-01-23
Avoid allocating in GraphQLInputObjectType.getFieldDefinitions

kilink:
Avoid copying / allocating a new List in GraphQLInputObjectType's getFieldDefinitions method. the field definitions are already stored as values in an ImmutableMap, and the default Guava implementation of the values method returns an ImmutableList; the issue arises due to the returned collection being a "partial view", which ImmutableList.copyOf uses as a signal that it should create a defensive copy of the passed in collection. In this particular case we just don't care if it's a partial view, since the owning ImmutableMap will be retained regardless.


bbakerman (approved):
Thanks for this

PR #3798, opened 2025-01-23
Optimize GraphQLUnionType.isPossibleType

kilink:
Remove the allocation overhead of the stream / lambda in isPossibleType; don't have the flame graph on hand any longer unfortunately but this was showing up as an allocation hotspot for one of our apps.


dondonz:
I reverted my PR that caused the problem, please update from master and your builds should be green again

bbakerman (approved):
Thanks for this

PR #3801, opened 2025-01-24
Return a singleton from CoercedVariables.emptyVariables and RawVariables.emptyVariables

kilink:
Make CoercedVariables.emptyVariables() and RawVariables.emptyVariables() return an immutable singleton instance.


dondonz (approved):
Nice! Thanks

bbakerman (approved):

PR #3803, opened 2025-01-28
Avoid allocation / stream overhead in ValuesResolverConversion

kilink:
Avoid the overhead of streams and extra copying incurred in externalValueToLiteralForList.


dondonz (approved):
Thank you!

bbakerman (approved):
Thanks

Once again good old `.stream()` hurts you at scale!

PR #3804, opened 2025-01-28
Avoid overhead of Stream / Iterator in RulesVisitor

kilink:
Avoid overhead of creating a Stream in filterRulesVisitingFragmentSpreads, and instead just do the filtering inline in a for loop.


dondonz (approved):
Thanks for all these performance improvements!

bbakerman (approved):
Did this turn up in a profiler or just getting rid of `.stream()`s

kilink:
> Did this turn up in a profiler or just getting rid of `.stream()`s

Hey @bbakerman . I am not trying to get rid of all stream calls, this came up when I was profiling an instrumentation implementation. It is true though that in hot code, stream can add quite a bit of overhead. For trivial scenarios like the filtering done in this method it seems straightforward to just unroll it into a loop and avoid the cost.

PR #3806, opened 2025-02-02
Bump Shadow to support Java 24 usages in byte-buddy

Goooler:
Follow up #3799.


dondonz:
Wow you are an absolute legend @Goooler , a new release and even a PR!!! Wow. Thanks for continuing on the Shadow project!

Issue #3760, opened 2024-12-02
A warning message when a type runtime wiring is added twice for the same type


dondonz:
This is only an early stage idea. Putting this here for me to investigate more later.

Today I had a very mysterious problem where a new type runtime wiring wasn't working as expected. It turns out, there was another type runtime wiring for the same type hidden away elsewhere in the codebase, but it was left unused for years. The older, incorrect type runtime wiring clobbered my new one.

This was in a Spring for GraphQL application. 

**RuntimeWiring in a forgotten file, written long ago:**
```kotlin
.type(
  TypeRuntimeWiring.newTypeWiring("MyInterface")
    .typeResolver { ... old out of date type resolver here ...},
)
```
**RuntimeWiring in my new file:**
```kotlin
.type(
  TypeRuntimeWiring.newTypeWiring("MyInterface")
    .typeResolver { ... my new type resolver here ...},
)
```
Spring won't always load type runtime wiring in the same order, so you don't know which one will clobber the other. In my case, the new one took precedence in 10+ integration tests, then the old one took precedence in deployment. You don't want two type runtime wirings for the same type.

There are schema validation checks in Spring for GraphQL https://docs.spring.io/spring-graphql/docs/1.2.2/reference/html/#execution.graphqlsource.schema-mapping-inspection, and perhaps that's a better place to add this sort of validation. Anyway putting this here as a reminder for later.

dondonz:
I had a look at RuntimeWiring, turns out someone already thought of this 😄 

It's called `strictMode` in `RuntimeWiring`, it'll raise an error if a type is defined twice.

```java
/**
* This puts the builder into strict mode, so if things get defined twice, for example, it will throw a {@link StrictModeWiringException}.
*
* @return this builder
*/
public Builder strictMode() {
  this.strictMode = true;
  return this;
}
```

PR #3824, opened 2025-02-15
Change runtime wiring redefinition checks to be strict by default

dondonz:
## The problem

As I reported in #3760 I was stumped by my code not working in prod where integration tests seemed fine. It turned out that the type wiring that I thought was brand new, was actually redefining another type in a long forgotten part of the codebase. 

This was a particularly tough bug because the order that Spring for GraphQL loaded the type wirings was non-deterministic, so you never knew which type wiring you got in prod. Of course Spring for GraphQL hasn't done anything wrong here, it is hard to distinguish which of two different wirings you want to use.

I think I haven't noticed this as a problem before, because when writing small tests in this repo, the order that datafetchers (and other elements) are redefined/clobbered are predictable.

## The solution

It turns out we already had a "strict mode" option to catch a number of type runtime wiring redefinitions, including type redefinition, as well as redefinitions on datafetchers, default datafetchers, scalars, and enums. However, it's not enabled by default so it's usually not switched on.

For most users, I think having strict checks for redefinition is the better default option. It is very strange to have two type runtime wirings for the same type, and one can clobber the other in an _unpredictable_ way. 

The only time we should allow more than 1 TypeRuntimeWiring definition for the same type is when adding DataFetchers for distinct child fields. By definition, we can only have one type resolver and one enum values provider in a TypeRuntimeWiring, so these must not be redefined.

This PR changes the default to strict checks set to true.

This will be a "breaking change", but it is possible to return to the previous behaviour.

### DIFF
--- a/src/main/java/graphql/schema/idl/RuntimeWiring.java
+++ b/src/main/java/graphql/schema/idl/RuntimeWiring.java
@@ -313,10 +313,17 @@ public class RuntimeWiring {
         public Builder type(TypeRuntimeWiring typeRuntimeWiring) {
             String typeName = typeRuntimeWiring.getTypeName();
             Map<String, DataFetcher> typeDataFetchers = dataFetchers.computeIfAbsent(typeName, k -> new LinkedHashMap<>());
+
+            Map<String, DataFetcher> additionalFieldDataFetchers = typeRuntimeWiring.getFieldDataFetchers();
             if (strictMode && !typeDataFetchers.isEmpty()) {
-                throw new StrictModeWiringException(format("The type %s has already been defined", typeName));
+                // Check if the existing type wiring contains overlapping DataFetcher definitions
+                for (String fieldName : additionalFieldDataFetchers.keySet()) {
+                    if (typeDataFetchers.containsKey(fieldName)) {
+                        throw new StrictModeWiringException(format("The field %s on type %s has already been defined", fieldName, typeName));
+                    }
+                }
             }
-            typeDataFetchers.putAll(typeRuntimeWiring.getFieldDataFetchers());
+            typeDataFetchers.putAll(additionalFieldDataFetchers);
 
             DataFetcher<?> defaultDataFetcher = typeRuntimeWiring.getDefaultDataFetcher();
             if (defaultDataFetcher != null) {
--- a/src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
+++ b/src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
@@ -206,7 +206,7 @@ class RuntimeWiringTest extends Specification {
 
         then:
         def e1 = thrown(StrictModeWiringException)
-        e1.message == "The type Foo has already been defined"
+        e1.message == "The field foo on type Foo has already been defined"
 
         when:
         RuntimeWiring.newRuntimeWiring()
@@ -253,6 +253,7 @@ class RuntimeWiringTest extends Specification {
         def runtimeWiring1 = RuntimeWiring.newRuntimeWiring()
                 .type(TypeRuntimeWiring.newTypeWiring("Foo").dataFetcher("foo", DF1))
                 .type(TypeRuntimeWiring.newTypeWiring("Foo").dataFetcher("bar", DF2))
+                .build()
 
         then:
         noExceptionThrown()
@@ -313,7 +314,7 @@ class RuntimeWiringTest extends Specification {
 
         then:
         noExceptionThrown()
-        runtimeWiring1.getDataFetchers().get("Foo").get("bar") == DF1
+        runtimeWiring1.getDataFetchers().get("Foo").get("foo") == DF2
 
         when:
         def runtimeWiring2 = RuntimeWiring.newRuntimeWiring()


src/test/groovy/graphql/execution/incremental/DeferExecutionSupportIntegrationTest.groovy
```diff
@@ -162,19 +162,25 @@ class DeferExecutionSupportIntegrationTest extends Specification {
...
-                .type(newTypeWiring("Post").dataFetcher("dataFetcherError", resolveWithException()))
-                .type(newTypeWiring("Post").dataFetcher("dataAndError", resolveWithDataAndError("data")))
-                .type(newTypeWiring("Post").dataFetcher("coercionError", resolve("Not a number", 10)))
-                .type(newTypeWiring("Post").dataFetcher("typeMismatchError", resolve([a: "A Map instead of a List"], 10)))
-                .type(newTypeWiring("Post").dataFetcher("nonNullableError", resolve(null)))
-                .type(newTypeWiring("Page").dataFetcher("summary", resolve("A page summary", 10)))
-                .type(newTypeWiring("Page").dataFetcher("text", resolve("The page full text", 100)))
-                .type(newTypeWiring("Comment").dataFetcher("content", resolve("Full content", 100)))
-                .type(newTypeWiring("Comment").dataFetcher("author", resolve([name: "Author name"], 10)))
+                .type(newTypeWiring("Post")
```
dondonz, line 165: Some test updates, extra datafetcher registrations should happen inside the same newTypeWiring builder

src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
```diff
@@ -193,15 +193,14 @@ class RuntimeWiringTest extends Specification {
         newWiring.fieldVisibility == fieldVisibility
     }
 
-    def "strict mode can stop certain redefinitions"() {
+    def "strict mode, on by default, can stop certain redefinitions"() {
```
dondonz, line 196: Updating tests and adding new ones to lock in the change in default behaviour

src/main/java/graphql/schema/idl/RuntimeWiring.java
```diff
@@ -192,11 +192,24 @@ private Builder() {
...
+        }
+
         /**
          * This puts the builder into strict mode, so if things get defined twice, for example, it will throw a {@link StrictModeWiringException}.
          *
          * @return this builder
+         *
+         * @deprecated strictMode default value changed to true, use {@link #strictMode(boolean)} instead
          */
+        @Deprecated(since = "2025-02-15", forRemoval = true)
```
dondonz, line 212: Setting `forRemoval = true` to give clear indication this is going to be removed soon

src/main/java/graphql/schema/idl/RuntimeWiring.java
```diff
@@ -307,6 +320,9 @@ public Builder type(TypeRuntimeWiring typeRuntimeWiring) {
 
             DataFetcher<?> defaultDataFetcher = typeRuntimeWiring.getDefaultDataFetcher();
             if (defaultDataFetcher != null) {
+                if (strictMode && defaultDataFetchers.containsKey(typeName)) {
+                    throw new StrictModeWiringException(format("The type %s already has a default data fetcher defined", typeName));
+                }
```
dondonz, line 325: An edge case that wasn't previously covered: if a type wiring's default datafetcher was redefined at the runtime wiring level (and not the type wiring level), previously didn't raise an error

See test below to lock in the new behaviour

src/test/groovy/graphql/execution/ExecutionStepInfoTest.groovy
```diff
@@ -160,8 +160,10 @@ class ExecutionStepInfoTest extends Specification {
 
         def runtimeWiring = newRuntimeWiring()
                 .type(newTypeWiring("Query").dataFetcher("hero", samwiseDF))
-                .type(newTypeWiring("User").dataFetcher("friends", friendsDF))
-                .type(newTypeWiring("User").dataFetcher("mates", friendsDF))
+                .type(newTypeWiring("User")
+                        .dataFetcher("friends", friendsDF)
+                        .dataFetcher("mates", friendsDF)
+                )
```
dondonz, line 166: Set datafetchers so it doesn't count as type redefinition

src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
```diff
@@ -193,15 +193,14 @@ class RuntimeWiringTest extends Specification {
...
         DataFetcher DF1 = env -> "x"
         DataFetcher DF2 = env -> "x"
         TypeResolver TR1 = env -> null
         EnumValuesProvider EVP1 = name -> null
 
         when:
         RuntimeWiring.newRuntimeWiring()
-                .strictMode()
```
dondonz, line 204: Remove these lines, as the default is strictness on

src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
```diff
@@ -248,34 +243,91 @@ class RuntimeWiringTest extends Specification {
         e5.message == "The type Foo has already has a default data fetcher defined"
     }
 
-    def "overwrite default data fetchers if they are null"() {
-
+    def "strict mode, if set to off, won't stop certain redefinitions"() {
```
dondonz, line 246: Lock in the strict mode off case

src/test/groovy/graphql/schema/idl/RuntimeWiringTest.groovy
```diff
@@ -248,34 +243,91 @@ class RuntimeWiringTest extends Specification {
...
+        // Do not permit redefinition of the default datafetcher
+        when:
+        RuntimeWiring.newRuntimeWiring()
+                .type(TypeRuntimeWiring.newTypeWiring("Foo").defaultDataFetcher(DEFAULT_DF))
+                .type(TypeRuntimeWiring.newTypeWiring("Foo").defaultDataFetcher(DEFAULT_DF2))
+                .build()
 
+        then:
+        def error = thrown(StrictModeWiringException)
+        error.message == "The type Foo already has a default data fetcher defined"
```
dondonz, line 331: Covering a new edge case, where the default datafetcher is redefined at the runtime wiring level

We already had a test for when it is redefined inside the type wiring

src/test/groovy/graphql/schema/idl/TypeRuntimeWiringTest.groovy
```diff
@@ -7,42 +7,41 @@ import spock.lang.Specification
 class TypeRuntimeWiringTest extends Specification {
 
     void setup() {
-        TypeRuntimeWiring.setStrictModeJvmWide(false)
+        TypeRuntimeWiring.setStrictModeJvmWide(true)
     }
 
     void cleanup() {
-        TypeRuntimeWiring.setStrictModeJvmWide(false)
+        TypeRuntimeWiring.setStrictModeJvmWide(true)
```
dondonz, line 14: Changes to tests to reflect the new default

src/test/groovy/graphql/schema/idl/TypeRuntimeWiringTest.groovy
```diff
@@ -51,38 +50,36 @@ class TypeRuntimeWiringTest extends Specification {
         e.message == "The field foo already has a data fetcher defined"
     }
 
-    def "strict mode can be turned on JVM wide"() {
-
-
+    def "strict mode can be turned off and on JVM wide"() {
```
dondonz, line 53: Invert the test

bbakerman (approved):

PR #3835, opened 2025-02-25
call Overlapping validation only once during validation

andimarek:
The goal here is simplification and performance improvements.


andimarek:
Our performance tests show a significant improvements for up to 150%.

Screenshot 2025-02-25 at 12 27 05 PM

PR #3837, opened 2025-02-27
Commit codeStyle settings

martinbonnin:
Follow up from https://github.com/graphql-java/graphql-java/pull/3772#discussion_r1880256539

This is so that IDE configuration is shared between contributors.


andimarek:
hi @martinbonnin  ... we have already a shared code style that is committed here in the repo: https://github.com/graphql-java/graphql-java/blob/master/graphql-java-code-style.xml

martinbonnin:
Ah! sorry I missed that.
Out of curiosity, is that automatically recognized by IJ? And if not, any reason not to add it to `.idea`? Reason I'm asking is that I inadvertently commited star imports which would probably have been caught by the IDE so I figured out it might as well be an automatic thing?

andimarek:
In our experience the committed code styles in the .idea folder never really worked consistently and often caused problems over time.  For example suddenly it changes it locally without any reason.

The import via exported code style worked well, if you import it :-) 

@bbakerman @dondonz thoughts?

martinbonnin:
> For example suddenly it changes it locally without any reason.

Agreed, I've seen that too during IJ updates I guess were they introduce new settings or change the defaults, I'm not sure...

We haven't seen it in Apollo Kotlin a long time so maybe it's sign things are stabilizing... But maybe not!

andimarek:
And that might seem trivial, but at the end it is very annoying ... this is why we just export/import the code style.

Would that work for you for now @martinbonnin ?

martinbonnin:
Yes of course, closing this.

PR #3869, opened 2025-03-29
Detect and prohibit new Guava classes

### LINKED ISSUE #3859: Explore how to prohibit using certain new Guava classes
The goal is to have as few dependencies as possible. To remove a dependency on Guava, we shade in selected classes such as ImmutableMap.

One gotcha is if new code uses a Guava class that's not on our shaded list, we don't have an automated way to detect that it would blow up later.

We could add an ArchUnit test to check for any Guava classes that aren't on our shaded list.

dondonz:
Fixes #3859 

## Problem
We want to keep dependencies to an absolute minimum as this is a low level library. However we make an exception for a few selected Guava classes which are shaded in.

There's a potential risk where someone uses a new Guava class in a pull request, and either: 
1. They are unaware that they should, where possible, use a non-Guava version of the method (which is different to their usual developer workflow)
2. Or they really must use a Guava class, we forget to shade in the new class in the Gradle build file, causing functionality to break

## Solution

This PR adds an Archunit test to detect new Guava class usage and produce a helpful error message to guide contributors to 1. use non-Guava versions where possible or 2. adjust the Gradle build file to shade in the new class

## Sample output

Example of deliberately failing pipeline (before I added the full set of exemptions): https://github.com/graphql-java/graphql-java/pull/3869/checks?check_run_id=39620708772

```
GuavaLimitCheck > should identify which classes use prohibited Guava dependencies STANDARD_OUT
    Prohibited Guava class usage found:

    Class: graphql.execution.ExecutionStrategy uses these prohibited Guava classes:
      - JavaClass{name='com.google.common.collect.Maps'}

    Class: graphql.schema.diffing.DiffImpl uses these prohibited Guava classes:
      - JavaClass{name='com.google.common.collect.Multiset'}
      - JavaClass{name='com.google.common.collect.HashMultiset'}
      - JavaClass{name='com.google.common.collect.Multisets'}

    Either:
    1. Preferred option: Replace them with alternatives that don't depend on Guava
    2. If absolutely necessary: Add these Guava classes to the shade configuration in the build.gradle file
```

## Note

We don't yet use anything from the `math` or `primitives` packages directly, but these are transitively used by `collect`, so they also need to be shaded in

```
 relocate('com.google.common', 'graphql.com.google.common') {
        include 'com.google.common.collect.*'
        include 'com.google.common.base.*'
        include 'com.google.common.math.*'
        include 'com.google.common.primitives.*'
    }
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,6 +125,8 @@ dependencies {
     testAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
     jmh 'org.openjdk.jmh:jmh-core:1.37'
     jmh 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
+
+    testImplementation "com.tngtech.archunit:archunit-junit5:1.2.0"
 }
 
 shadowJar {
--- a/src/test/groovy/graphql/GuavaLimitCheck.groovy
+++ b/src/test/groovy/graphql/GuavaLimitCheck.groovy
@@ -1,0 +1,78 @@
+package graphql
+
+import com.tngtech.archunit.core.domain.JavaClasses
+import com.tngtech.archunit.core.importer.ClassFileImporter
+import com.tngtech.archunit.core.importer.ImportOption
+import spock.lang.Specification
+
+/*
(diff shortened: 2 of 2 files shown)


bbakerman (approved):
Nice one

PR #3891, opened 2025-04-04
Add JSpecify nullability annotations to DataFetcherResult

dondonz:
I made a classic NPE mistake because I forgot to check if `data` was null while writing a DataFetcher

With JSpecify, we can detect NPEs like this at compile time rather than run time.

Have a look at the documentation for `@NullMarked` https://jspecify.dev/docs/user-guide/#nullmarked

This is the first PR to add JSpecify to an existing class so let's have a discussion.


andimarek (approved):
Exactly what we should do incrementally.

Issue #3852, opened 2025-03-13
Explore JSpecify tooling


dondonz:
We've recently added JSpecify annotations

Adding this issue to look into new JSpecify tooling that can run build-time null checks. We might need to do more work as well, as we don't yet have nullability annotations in all the places.

dondonz:
We will probably add Error Prone & Null Away for static checks as we progressively add more annotations

bclozel:
I've just left a comment in https://github.com/graphql-java/graphql-java/pull/3895#issuecomment-2875362794
From my experience, it is easier to configure tooling first before annotating the entire codebase, even if it means disabling it for the CI until things settle down. I can help out with Gradle configuration.

andimarek:
we have now adopted jspecifcy + tooling

PR #3899, opened 2025-04-06
Specify nullness for nodes in the `graphql.language` package

mk868:
Related: https://github.com/graphql-java/graphql-java/issues/3878

**I applied `@NullMarked` and `@Nullable` annotations to these classes:**

- `graphql.language.Node` - partially annotated with `@Nullable`
- `graphql.language.AbstractNode`
- `graphql.language.ArrayValue`
- `graphql.language.BooleanValue`
- `graphql.language.EnumValue`
- `graphql.language.FloatValue`
- `graphql.language.IntValue`
- `graphql.language.NullValue`
- `graphql.language.ObjectValue`
- `graphql.language.ScalarValue`
- `graphql.language.StringValue` - I was not sure about the nullability of `value`, for now I marked it `@Nullable`.
- `graphql.language.Value`
- `graphql.language.VariableReference`

**Notes**

-  `SourceLocation sourceLocation` - can contain a null value in all places

**Potential problems (will be pointed by https://github.com/graphql-java/graphql-java/issues/3852 )**

* `EnumValue.Builder#build()` can build an `EnumValue` object with a null `name`
* `FloatValue.Builder#build()` can build an `FloatValue` object with a null `value`
* `IntValue.Builder#build()` can build an `IntValue` object with a null `value`
* `VariableReference.Builder#build()` can build an `VariableReference` object with a null `name`


dondonz:
Giving Copilot a try for fun, I'm curious what it says

src/main/java/graphql/language/EnumValue.java
```diff
@@ -112,8 +115,8 @@ public EnumValue transform(Consumer<Builder> builderConsumer) {
     }
 
     public static final class Builder implements NodeBuilder {
-        private SourceLocation sourceLocation;
-        private String name;
+        private @Nullable SourceLocation sourceLocation;
+        private @Nullable String name;
```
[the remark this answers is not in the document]
dondonz: Hello, thanks for your patience while away for Easter.

In this builder, name is set to nullable, but the constructor above has name as non-nullable, and the builder method below has it as non-nullable. Did you mean for the Builder's name to be instead non-nullable?

src/main/java/graphql/language/FloatValue.java
```diff
@@ -111,8 +114,8 @@ public static Builder newFloatValue(BigDecimal value) {
     }
 
     public static final class Builder implements NodeBuilder {
-        private SourceLocation sourceLocation;
-        private BigDecimal value;
+        private @Nullable SourceLocation sourceLocation;
+        private @Nullable BigDecimal value;
```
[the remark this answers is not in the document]
dondonz: This builder should be consistent with the constructor for `value`

@andimarek @bbakerman: is it possible in practice for `FloatValue` to ever be constructed with a null? Or would such an object always be constructed as a `NullValue`?

src/main/java/graphql/language/IntValue.java
```diff
@@ -110,8 +113,8 @@ public IntValue transform(Consumer<Builder> builderConsumer) {
     }
 
     public static final class Builder implements NodeBuilder {
-        private SourceLocation sourceLocation;
-        private BigInteger value;
+        private @Nullable SourceLocation sourceLocation;
+        private @Nullable BigInteger value;
```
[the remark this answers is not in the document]
dondonz: Same comment as for FloatValue, I will double check if a null `value` is in practice possible or not

src/main/java/graphql/language/StringValue.java
```diff
@@ -110,8 +113,8 @@ public StringValue transform(Consumer<Builder> builderConsumer) {
     }
 
     public static final class Builder implements NodeBuilder {
-        private SourceLocation sourceLocation;
-        private String value;
+        private @Nullable SourceLocation sourceLocation;
+        private @Nullable String value;
```
dondonz, line 117: Similar to above Value classes, I will double check if value can be nullable or if in practice it is represented by a `NullValue`

src/main/java/graphql/language/VariableReference.java
```diff
@@ -106,9 +109,9 @@ public VariableReference transform(Consumer<Builder> builderConsumer) {
     }
 
     public static final class Builder implements NodeBuilder {
-        private SourceLocation sourceLocation;
+        private @Nullable SourceLocation sourceLocation;
         private ImmutableList<Comment> comments = emptyList();
-        private String name;
+        private @Nullable String name;
```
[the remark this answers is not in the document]
dondonz: `name` in the above constructor and builder method are not nullable, did you meant to make this field nullable?

I think in practice, a VariableReference should always have a non-nullable name

src/main/java/graphql/language/EnumValue.java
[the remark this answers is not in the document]
mk868: No, nullable fields in the builder are expected.  
By default, when a builder instance is created, all values are set to their defaults (null for reference fields) - thus, it's necessary to mark these fields `@Nullable`.  
This initial `Builder` state represents an incomplete configuration, and the `EnumValue` should not be constructible in this state - calling the `build()` method should result in an `IllegalStateException`.  

I planned to add something like this later:
```java
public EnumValue build() {
    if (name == null) {
        throw new IllegalStateException("name must be set");
    }
    return new EnumValue(name, sourceLocation, comments, ignoredChars, additionalData);
}
```

dondonz:
Coming back in here to say thanks for waiting and thanks for the PR. We have a new PR up to add ErrorProne & NullAway checks which I'd like to merge in before new JSpecify PRs

src/main/java/graphql/language/ArrayValue.java
```diff
@@ -87,7 +91,8 @@ public String toString() {
 
     @Override
     public ArrayValue deepCopy() {
-        return new ArrayValue(deepCopy(values), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
+        List<Value> copiedValues = deepCopy(values);
+        return new ArrayValue(copiedValues != null ? copiedValues : emptyList(), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
```
dondonz: Want to call out this is a small change to not allow a null value. A list will be used instead. Without this change, there is a NullAway error

src/main/java/graphql/language/ArrayValue.java
```diff
@@ -101,6 +106,7 @@ public ArrayValue transform(Consumer<Builder> builderConsumer) {
         return builder.build();
     }
 
+    @NullUnmarked
```
dondonz, line 109: After giving a few classes the JSpecify treatment: we now generally annotate Builders with `@NullUnmarked`

src/main/java/graphql/language/ObjectValue.java
```diff
@@ -77,7 +81,8 @@ public boolean isEqualTo(Node o) {
 
     @Override
     public ObjectValue deepCopy() {
-        return new ObjectValue(deepCopy(objectFields), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
+        List<ObjectField> copiedFields = deepCopy(objectFields);
+        return new ObjectValue(copiedFields != null ? copiedFields : emptyList(), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
```
dondonz: Similar to ArrayValue, want to call out this is a small change to not allow a null value. A list will be used instead. Without this change, there is a NullAway error

dondonz:
Hello after a few other PRs with JSpecify annotations, we settled on using `@NullUnmarked` for Builders as they can be instantiated with null variables, which are later set to non-null values. I updated your PR.

We've also since added NullAway compile time checks which identified a very small extra change, to make the values of both ArrayValue and ObjectValue empty lists rather than null.

Thanks for the PR!

src/main/java/graphql/language/AbstractNode.java
```diff
@@ -56,15 +59,15 @@ public Map<String, String> getAdditionalData() {
     }
 
     @SuppressWarnings("unchecked")
-    protected <V extends Node> V deepCopy(V nullableObj) {
+    protected <V extends @Nullable Node> @Nullable V deepCopy(@Nullable V nullableObj) {
```
andimarek: I don't understand this change @dondonz ... why should we allow null values here?

dondonz: The first line we check whether the input is nullable `if (nullableObj == null) {` so I have allowed the input to be nullable

I agree it's strange in the context of `Value` like `FloatValue`. 

Maybe there was another reason to allow this to be nullable in the first place?

andimarek: So we have two options here: either declare `V` as a type that also can be null or declare input and output as Nullable. We have done both here, which doesn't make sense I think.

andimarek (approved):
Can be merged, but also could first fix the small comment I left.

src/main/java/graphql/language/AbstractNode.java
dondonz: Repeating conversation here: I will change the method signature to become
```
protected <V extends Node> @Nullable V deepCopy(@Nullable V nullableObj)
```

This in English is meant to express that V extending Node is not nullable, but because the method will handle null inputs, it may still return a null

src/main/java/graphql/language/ArrayValue.java
```diff
@@ -87,7 +91,8 @@ public String toString() {
 
     @Override
     public ArrayValue deepCopy() {
-        return new ArrayValue(deepCopy(values), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
+        List<Value> copiedValues = assertNotNull(deepCopy(values));
```
dondonz, line 94: Add an assert here to keep NullAway happy

In practice it is not possible to have null values. Have to put this here otherwise you'll get a NullAway error

src/main/java/graphql/language/ObjectValue.java
```diff
@@ -77,7 +81,8 @@ public boolean isEqualTo(Node o) {
 
     @Override
     public ObjectValue deepCopy() {
-        return new ObjectValue(deepCopy(objectFields), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
+        List<ObjectField> copiedFields = assertNotNull(deepCopy(objectFields));
```
dondonz, line 84: Add this assert to keep NullAway happy, even though objectFields is not null

glasser:
@dondonz We are working to upgrade to GraphQL-Java v25. We use Kotlin so nullability annotations directly affect our ability to compile. I'm a bit confused about why StringValue.getValue() is now null. I looked through the codebase and could not find any instances where a StringValue is constructed with a null value, which makes sense — there's a NullValue in the type hierarchy to cover that case.

As far as I can tell, the impact of declaring StringValue.getValue() as nullable rather than non-null is just to force anyone writing software in a way that depends on these annotations (like Kotlin) to insert pointless null checks that will never fail.

I suppose that declaring the value as nullable is a compatibility issue since users could be creating StringValues with null value ourselves, but is there a real use case for this, or could we change StringValue in v26 to require (and check!) non-null value?

PR #3902, opened 2025-04-06
Cleanup  ReactiveSupport

hantsy:
ReactiveStreams java provides a utility class to convert between ReactiveStreams API and Jdk 9+ Flow APIs. Utilize with this class, the `ReactiveSupport` can be simplified.


hantsy:
Under windows powershell, the checkout action failed due to invalid path of the performance tests files. I used the `git restore HEAD :/`(this command is appeared in the error message) and `git status` there are no new files.

OK, I will try to use cygwin later.

PR #3926, opened 2025-04-23
support variable forward-references in directives

jbellenger:
This PR is a repro of a potential issue, filed at #3927

The potential issue is related to variable reference validation when variables are used in directives that are applied to other variable definitions 🤪

See line comments in diff.


src/test/groovy/graphql/validation/rules/NoUndefinedVariablesTest.groovy
```diff
@@ -181,4 +181,37 @@ class NoUndefinedVariablesTest extends Specification {
...
+        when:
+        traverse(query)
+
+        then:
+        errorCollector.errors.isEmpty()
+    }
+
+    def "variable-directive forward-reference"() {
+        given:
+        def query = "query (\$v1: Int @dir(arg: \$v2), \$v2: Int) { __typename }"
```
jbellenger, line 209: the other test cases in this file pass, though this one fails:
```
Undefined variable 'v2'
```

I think (but am not certain) that this pattern is allowed by the graphql spec.

PR #3953, opened 2025-05-09
fix case when a fragment is deferred and non deferred at the same time

andimarek:


src/test/groovy/graphql/execution/incremental/DeferExecutionSupportIntegrationTest.groovy
```diff
@@ -1604,6 +1604,37 @@ class DeferExecutionSupportIntegrationTest extends Specification {
...
+          }
+        '''
+        when:
+        def initialResult = executeQuery(query)
+
+        then:
+        initialResult.toSpecification() == [
+                data: [post: [summary: "A summary", text: "The full text"]]
+        ]
+
```
felipe-gdr, line 1633: can you assert on the incremental payloads as well?

andimarek: What do you want to assert exactly? there is no incremental result in this case: is that correct?

Should we add another test case with some incremental result?

bbakerman (approved):

src/test/groovy/graphql/execution/incremental/DeferExecutionSupportIntegrationTest.groovy
felipe-gdr: You're right. I misread your test scenario.
In that case, do we need to assert that there aren't incremental payloads? Like, `initialResult` is not `IncrementalExecutionResult`? 
And perhaps change the name of the variable, `initialResult` gives the impression that this is an incremental result.

felipe-gdr (approved):

Issue #3963, opened 2025-05-14
Add toString, hashcode, equals methods to DataFetcherResult


dondonz:
Thanks @AlexandreCarlton for reporting

dondonz:
Closed by #3964 . Thanks @AlexandreCarlton ! This was also cherry picked into the 24.0 release.

PR #3964, opened 2025-05-14
Implement toString/hashCode/equals for DataFetcherResult

AlexandreCarlton:
This facilitates usage in tests for quick checking of equality (and useful error messages in assertion errors).

References #3963.


dondonz (approved):
Thanks!

PR #3976, opened 2025-05-19
adding Profiler

andimarek:
While we recently added a Java agent to profile a GraphQL Java application the following things became clear:

Java agents are not the future of the JVM: soon a dynamic loading will be forbidden. See https://openjdk.org/jeps/451
We added the ability to track the engine running state
We added the ability to chain DataLoaders, which as a side effect allow for tracking which dataloaders are used by which field / DataFetcher
Together this means Java agents are not really future proof + they are not needed anymore, as we have all the abilities now inside GraphQL Java directly itself.

The goal is to remove the Java agent eventually and offer a better alternative in form of this PRs: a built-in Profiler in GraphQL Java.

A Profiler is enabled per execution (request) and will collect with very minimal overhead all relevant informations to understand the performance and execution. The result will be made available via GraphQLContext.


src/test/groovy/graphql/ProfilerTest.groovy
```diff
@@ -0,0 +1,286 @@
...
+        profilerResult.getDispatchEvents()[0].dataLoaderName == "name"
+        profilerResult.getDispatchEvents()[0].level == 1
+        profilerResult.getDispatchEvents()[0].count == 2
+        profilerResult.getDispatchEvents()[1].dataLoaderName == "name"
+        profilerResult.getDispatchEvents()[1].level == 1
+        profilerResult.getDispatchEvents()[1].count == 2
+
+    }
+
+
```
ronodhirSoumik, line 589: extra line - intentional?

src/main/java/graphql/ExecutionInput.java
```diff
@@ -414,6 +422,11 @@ public Builder dataLoaderRegistry(DataLoaderRegistry dataLoaderRegistry) {
             return this;
         }
 
+        public Builder profileExecution(boolean profileExecution) {
+            this.profileExecution = profileExecution;
+            return this;
+        }
```
dondonz, line 428: For another PR, not this one

Can you add the toggle to `GraphQLUnusualConfiguration` instead?

src/main/java/graphql/ProfilerImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public void dataLoaderUsed(String dataLoaderName) {
+        profilerResult.addDataLoaderUsed(dataLoaderName);
+    }
+
+    @Override
+    public void chainedStrategyDispatching(int level) {
+        profilerResult.chainedStrategyDispatching(level);
```
dondonz: To make it more obvious to a user, can comment on this method (or interface method) to indicate that this refers to the option `DataLoaderDispatchingContextKeys.ENABLE_DATA_LOADER_CHAINING`

I think to us, we know what Chained DataLoaders are but this is not a well known phrase

I'll also make a note to mention in the documentation, this is a way to measure the impact of Chained DataLoaders

src/main/java/graphql/Profiler.java
```diff
@@ -0,0 +1,57 @@
...
+import graphql.execution.EngineRunningObserver;
+import graphql.execution.ResultPath;
+import graphql.language.OperationDefinition;
+import graphql.schema.DataFetcher;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@Internal
+@NullMarked
+public interface Profiler {
```
dondonz, line 15: Add a quick JavaDoc explaining what this does

> A Profiler monitors important execution timings and counters. Enable this via ExecutionInput/new Unusual Configuration method

Of course 1 sentence isn't enough, let's add a documentation link here when it's written

andimarek: This is an internal class btw

src/main/java/graphql/ProfilerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+import graphql.schema.PropertyDataFetcher;
+import graphql.schema.SingletonPropertyDataFetcher;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+
```
dondonz, line 24: Not for this PR, but in the future, with more Profiler features you might want to introduce options to toggle sets of tracking on and off

e.g. toggle DataLoader related tracking on and off

src/main/java/graphql/Profiler.java
```diff
@@ -51,7 +51,11 @@ default void batchLoadedOldStrategy(String name, int level, int count) {
 
     }
 
-    default void batchLoadedNewStrategy(String name, @Nullable Integer level, int count) {
+    default void batchLoadedNewStrategy(String dataLoaderName, @Nullable Integer level, int count) {
+
+    }
+
+    default <V> void manualDispatch(String dataLoaderName, int level, int count) {
```
dondonz, line 56: Nice

src/main/java/graphql/ExecutionInput.java
dondonz: To be discussed!

dondonz: Can be considered in another PR

dondonz (approved):

PR #3980, opened 2025-05-22
make dataloader work inside defer blocks

andimarek:

PR #3981, opened 2025-05-22
makes the level notion more clear

andimarek:
Fixes the level tracking. This will probably not be merged, but combined with #3980


andimarek:
Will be included in #3980

PR #3984, opened 2025-05-23
Add interface addition event in SchemaDiff

mgadda:
Update `SchemaDiff.checkImplements` to emit ADDITION (info) events for new interfaces. 

* Add a loop to iterate over new interfaces and check if they are present in old interfaces.
* Emit an ADDITION event if a new interface is not present in the old interfaces.
* Retain the existing behavior with respect to interface removals.

Add a new test case in `SchemaDiffTest.groovy` to verify that `checkImplements` emits ADDITION events for new interfaces.


bbakerman (approved):
Thanks for this

PR #3985, opened 2025-05-23
Add ArchUnit rule to ban javax & JetBrains nullability annotations

dondonz:
Going forward we will be using JSpecify Nullability annotations.

It's possible to accidentally pull in the old javax Nullability annotations because this doesn't need any changes to our build.gradle dependencies.

While it's unlikely someone would accidentally add JetBrains Nullability annotations, because these require a new dependency to be added, I have also banned them.

Sample test output when a banned usage is found:
```
We are using JSpecify nullability annotations only. Please change the following to use JSpecify instead:
- Parameter <java.util.concurrent.CompletableFuture<T>> of method <graphql.execution.Async.orNullCompletedFuture(java.util.concurrent.CompletableFuture)> is annotated with <javax.annotation.Nullable> in (Async.java:0)
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -176,7 +176,7 @@ shadowJar {
     bnd('''
 -exportcontents: graphql.*
 -removeheaders: Private-Package
-Import-Package: !android.os.*,!com.google.*,!org.checkerframework.*,!javax.annotation.*,!graphql.com.google.*,!org.antlr.*,!graphql.org.antlr.*,!sun.misc.*,org.jspecify.annotations;resolution:=optional,*
+Import-Package: !android.os.*,!com.google.*,!org.checkerframework.*,!graphql.com.google.*,!org.antlr.*,!graphql.org.antlr.*,!sun.misc.*,org.jspecify.annotations;resolution:=optional,*
 ''')
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -176,7 +176,7 @@ shadowJar {
     bnd('''
 -exportcontents: graphql.*
 -removeheaders: Private-Package
-Import-Package: !android.os.*,!com.google.*,!org.checkerframework.*,!javax.annotation.*,!graphql.com.google.*,!org.antlr.*,!graphql.org.antlr.*,!sun.misc.*,org.jspecify.annotations;resolution:=optional,*
+Import-Package: !android.os.*,!com.google.*,!org.checkerframework.*,!graphql.com.google.*,!org.antlr.*,!graphql.org.antlr.*,!sun.misc.*,org.jspecify.annotations;resolution:=optional,*
```
dondonz, line 179: While I'm here, we can clean up our config to remove javax.annotation

andimarek:
We also need to add this to gradle itself, right? Or am I missing something here?

dondonz:
> We also need to add this to gradle itself, right? Or am I missing something here?

This runs as a standard test. If there's a violation, it'll turn up in our usual build as a failing test

Or did you mean about adding ArchUnit as a dependency? It was already added in another PR that added a rule

andimarek (approved):

PR #3987, opened 2025-05-24
deterministic SourceLocation serialization

jbellenger:
In upgrading from 22.3 to 24.0, some of our tests that make assertions on serialized responses started failing.

I think this was probably caused by [#3753](https://github.com/graphql-java/graphql-java/pull/3753/files#r1855350447), which changed the serialized representation from a LinkedHashMap to a `Map.of`, which has [randomized iteration order](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/ImmutableCollections.java#L62-L68). This PR brings back the LinkedHashMap serialization format.

I'd appreciate some feedback on this! While the spec doesn't describe how the fields of an error should be ordered, non-deterministic ordering can make it difficult to build reliable tests that interact with GraphQL errors.

### DIFF
--- a/src/main/java/graphql/GraphqlErrorHelper.java
+++ b/src/main/java/graphql/GraphqlErrorHelper.java
@@ -73,7 +73,10 @@ public class GraphqlErrorHelper {
         if (line < 1 || column < 1) {
             return null;
         }
-        return Map.of("line", line, "column", column);
+        LinkedHashMap<String, Object> map = new LinkedHashMap<>(2);
+        map.put("line", line);
+        map.put("column", column);
+        return map;
     }
 
     static List<GraphQLError> fromSpecification(List<Map<String, Object>> specificationMaps) {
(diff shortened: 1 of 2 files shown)


andimarek:
Agree ... this was an oversight and we will change it back

src/main/java/graphql/GraphqlErrorHelper.java
```diff
@@ -73,7 +73,10 @@ public static Object location(SourceLocation location) {
         if (line < 1 || column < 1) {
             return null;
         }
-        return Map.of("line", line, "column", column);
+        LinkedHashMap<String, Object> map = new LinkedHashMap<>(2);
```
andimarek, line 76: Don't we have a Util method for that? Maybe not, but we should have that emulates the Map.of ...

andimarek (approved):

src/test/groovy/graphql/GraphqlErrorHelperTest.groovy
```diff
@@ -154,4 +155,15 @@ class GraphqlErrorHelperTest extends Specification {
             assert gErr.getExtensions() == null
         }
     }
+
+    @RepeatUntilFailure(maxAttempts = 1_000)
```
andimarek, line 159: Oh ... never knew that annotation ... learned that something new here

andimarek:
Hey @jbellenger ... We backported this to 24.1 which is available now.

Issue #3956, opened 2025-05-11
Sonatype migration before 30 June 2025


dondonz:
Sonatype is changing their publishing APIs and we have to migrate before 30 June 2025.

Instructions are here https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/#self-service-migration

Additional discussion on https://github.com/gradle-nexus/publish-plugin/issues/330

To be done after our current release wave

dondonz:
At a minimum, before 30 June we must:
1. Register our namespace ✅ done
2. Redirect our current plugin to use a new URL as per this documentation: https://github.com/gradle-nexus/publish-plugin/pull/377/files

We can think about keeping the same plugin or changing it at a later date

dondonz:
We have updated all repos with the new credentials

PR #4009, opened 2025-06-10
Add the new Sonatype URL

dondonz:
Part 2 of #3956 

Sonatype is changing their publishing API. We have to do 2 things before 30 June 2025, the first was to register our namespace (already done), the second is to switch over to the new URL (this PR).

Note: new auth tokens should be added before merging in this PR.

Sonatype documentation: https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration. Cross-referenced with Gradle publish plugin documentation https://github.com/gradle-nexus/publish-plugin/blob/master/README.md#publishing-to-maven-central-via-sonatype-central

From the documentation:
> If you previously published via an OSSRH server, you will need to replace your OSSRH token with a Portal token. Publishing with an OSSRH token will result in 401 responses.


build.gradle
```diff
@@ -352,6 +352,10 @@ nexusPublishing {
         sonatype {
             username = System.env.MAVEN_CENTRAL_USER
             password = System.env.MAVEN_CENTRAL_PASSWORD
+            // https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration
+            nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
+            // GraphQL Java does not publish snapshots, but adding this URL for completeness
+            snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
```
dondonz, line 358: We didn't need to specify these URLs before, because we were using the default (and soon to be sunsetted) URLs

Issue #3989, opened 2025-05-26
Ban Map.of


dondonz:
Write a rule to ban Map.of to ensure iteration order, to avoid an issue such as #3987 in future

dondonz:
While we're at it, can also make a helper method to make the right map instead

PR #4068, opened 2025-07-30
WIP Profiler tweaks

dondonz:
We're putting the finishing touches on the Profiler

First tweak: `ExecutionId` is better cast to a string for the output map, otherwise a logger can complain it doesn't know how to serialise this type to JSON

WIP, testing now and likely to find more improvements

### DIFF
--- a/src/main/java/graphql/ProfilerResult.java
+++ b/src/main/java/graphql/ProfilerResult.java
@@ -273,7 +273,7 @@ public class ProfilerResult {
 
     public Map<String, Object> shortSummaryMap() {
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("executionId", Assert.assertNotNull(executionId));
+        result.put("executionId", Assert.assertNotNull(executionId).toString());
         result.put("operation", operationType + ":" + operationName);
         result.put("startTime", startTime);
         result.put("endTime", endTime);


andimarek-atlassian (approved):

src/main/java/graphql/ProfilerResult.java
```diff
@@ -273,12 +273,12 @@ public List<String> getInstrumentationClasses() {
 
     public Map<String, Object> shortSummaryMap() {
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("executionId", Assert.assertNotNull(executionId));
+        result.put("executionId", Assert.assertNotNull(executionId).toString());
         result.put("operation", operationType + ":" + operationName);
         result.put("startTime", startTime);
         result.put("endTime", endTime);
-        result.put("totalRunTime", (endTime - startTime) + "(" + (endTime - startTime) / 1_000_000 + "ms)");
-        result.put("engineTotalRunningTime", engineTotalRunningTime + "(" + engineTotalRunningTime / 1_000_000 + "ms)");
```
dondonz, line 281: Change to nanoseconds only, if need be someone can convert to ms later

Remove string concatenation to make it easier to read

src/main/java/graphql/ProfilerResult.java
```diff
@@ -310,11 +310,21 @@ public Map<String, Object> shortSummaryMap() {
...
+        completableFutureNotCompleted.put("count", notCompletedCount);
+        completableFutureNotCompleted.put("invocations", notCompletedInvokeCount);
+        LinkedHashMap<String, Integer> materialized = new LinkedHashMap<>(2);
+        materialized.put("count", materializedCount);
+        materialized.put("invocations", materializedInvokeCount);
+
+        dFRTinfo.put(DataFetcherResultType.COMPLETABLE_FUTURE_COMPLETED.name(), completableFutureCompleted);
+        dFRTinfo.put(DataFetcherResultType.COMPLETABLE_FUTURE_NOT_COMPLETED.name(), completableFutureNotCompleted);
+        dFRTinfo.put(DataFetcherResultType.MATERIALIZED.name(), materialized);
+        result.put("dataFetcherResultTypes", dFRTinfo);
```
dondonz, line 318: I know, it's much more verbose this way compared to Map.of

I plan to ban Map.of in the near future https://github.com/graphql-java/graphql-java/issues/3989 - in another place, it was important to have consistent ordering when serialising source location. It doesn't matter as much here but might as well make it consistent and ban Map.of

andimarek: it is the right thing todo ... we don't want Map.of. 

We should invent a version of it that is based on LinkedHashMap

andimarek (approved):

src/main/java/graphql/ProfilerResult.java
```diff
@@ -273,12 +273,12 @@ public List<String> getInstrumentationClasses() {
 
     public Map<String, Object> shortSummaryMap() {
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("executionId", Assert.assertNotNull(executionId));
+        result.put("executionId", Assert.assertNotNull(executionId).toString());
         result.put("operation", operationType + ":" + operationName);
         result.put("startTime", startTime);
```
andimarek: actually we should suffix startTime and endTime also with `Ns` to be clear here

src/main/java/graphql/ProfilerResult.java
```diff
@@ -273,12 +273,12 @@ public List<String> getInstrumentationClasses() {
 
     public Map<String, Object> shortSummaryMap() {
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("executionId", Assert.assertNotNull(executionId));
+        result.put("executionId", Assert.assertNotNull(executionId).toString());
         result.put("operation", operationType + ":" + operationName);
```
andimarek: should that be two fields? operation Type and operation name? or one Map inside "operation"?

dondonz: Two fields are better, and better to be clearer on operation name vs type.

dondonz: I see what you meant, should they be nested or not

We've done both: nesting some information, but keeping all the timing information at the top level. I think operation name is one of the most important keys and I'd like to have that at the top level.

src/main/java/graphql/ProfilerResult.java
dondonz: I made that a todo on the ban Map.of ticket

src/test/groovy/graphql/ProfilerTest.groovy
```diff
@@ -39,9 +40,9 @@ class ProfilerTest extends Specification {
         def schema = TestUtil.schema(sdl, [Query: [
                 hello: { DataFetchingEnvironment dfe -> return "world" } as DataFetcher
         ]])
-        def graphql = GraphQL.newGraphQL(schema).build();
+        def graphql = GraphQL.newGraphQL(schema).build()
 
-        ExecutionInput ei = ExecutionInput.newExecutionInput()
+        ExecutionInput ei = newExecutionInput()
```
dondonz, line 45: This doesn't change anything, only IDE complaints

src/test/groovy/graphql/ProfilerTest.groovy
```diff
@@ -477,11 +480,10 @@ class ProfilerTest extends Specification {
...
                                                       "/foo"      : COMPLETABLE_FUTURE_NOT_COMPLETED]
-        profilerResult.shortSummaryMap().get("dataFetcherResultTypes") == ["COMPLETABLE_FUTURE_COMPLETED"    : "(count:2, invocations:4)",
-                                                                           "COMPLETABLE_FUTURE_NOT_COMPLETED": "(count:2, invocations:2)",
-                                                                           "MATERIALIZED"                    : "(count:2, invocations:4)"]
-
-
+        profilerResult.shortSummaryMap().get("dataFetcherResultTypes") == ["COMPLETABLE_FUTURE_COMPLETED"    : ["count":2, "invocations":4],
+                                                                           "COMPLETABLE_FUTURE_NOT_COMPLETED": ["count":2, "invocations":2],
+                                                                           "MATERIALIZED"                    : ["count":2, "invocations":4]]
+        profilerResult.shortSummaryMap().get("executionId") == "myExecutionId"
```
dondonz, line 486: This is the only real test change

src/main/java/graphql/ProfilerResult.java
```diff
@@ -273,12 +273,13 @@ public List<String> getInstrumentationClasses() {
 
     public Map<String, Object> shortSummaryMap() {
         Map<String, Object> result = new LinkedHashMap<>();
-        result.put("executionId", Assert.assertNotNull(executionId));
-        result.put("operation", operationType + ":" + operationName);
-        result.put("startTime", startTime);
-        result.put("endTime", endTime);
-        result.put("totalRunTime", (endTime - startTime) + "(" + (endTime - startTime) / 1_000_000 + "ms)");
-        result.put("engineTotalRunningTime", engineTotalRunningTime + "(" + engineTotalRunningTime / 1_000_000 + "ms)");
+        result.put("executionId", Assert.assertNotNull(executionId).toString());
```
dondonz, line 276: Cast to string to serialise in JSON logs

PR #4071, opened 2025-07-31
comment out the processor by default

andimarek:
Hey @dfa1 @bbakerman ,

unfortunately this really breaks idea rebuild/build functionality. We need to keep this commented out by default and put it back in if we want to run JMH tests from idea.

If you find another solution that would be great ... I spend some time finding one, but could not . 😢

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -150,7 +150,9 @@ dependencies {
     // this is needed for the idea jmh plugin to work correctly
     jmh 'org.openjdk.jmh:jmh-core:1.37'
     jmh 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
-    jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
+
+    // comment this in if you want to run JMH benchmarks from idea
+//    jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
 
     errorprone 'com.uber.nullaway:nullaway:0.12.7'
     errorprone 'com.google.errorprone:error_prone_core:2.41.0'


dfa1:
@andimarek I'm trying to reproduce the issue mentioned here. 

If I uncomment the line above, I'm able to:
- run build/rebuild operation in IntelliJ without issues
- run tests without issues
- run "gradle jmh" without issues
- run again any tests within IntelliJ

Do you have a specific workflow that is broken?
Do you have these settings in IntelliJ? (don't remember if this default or not)
image

andimarek:
@dfa1 one thing we always use is the IntelliJ compiler instead of gradle because it is much faster:

Do you have the same?

Screenshot 2025-08-06 at 10 47 01 am

dfa1:
> @dfa1 one thing we always use is the IntelliJ compiler instead of gradle because it is much faster:
ok this is nice to know 
> 
> Do you have the same?
now yes... I got this error:
image

it is same error?

bbakerman:
Capto_Capture 2025-08-07_06-17-53_pm


I find that when I set "Builds via IDEA" in the gradle runner I have problems where the ANTLR generator task does not run 

```
Task :generateGrammarSource
```

and then get compilation errors.  but if I clean and build and something else generates the grammar source - it works

I alo noticed that `idea.version` is set when the gradle is rub by idea versus the command line.  I wonder if we can some how use that in some way

```
    if (System.getProperty("idea.version") == null) {
        jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
    }
```

bbakerman:
I also get

```
Groovyc: Internal groovyc error: code 1
```

when I compile with 

Capto_Capture 2025-08-07_06-32-34_pm

bbakerman:
but if I changed it to 

Capto_Capture 2025-08-07_06-35-32_pm

and also put in

```
jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
```

then I get the JMH to run inside IDEA without error


I think the days of being able to compile via "IDEA" are over.  We have reached a complication point where in order to do ALL thing things (compile + grammar generation + groovy tests + jmh running) we need to build using gradle.

My view is we should put in the annotation processor BUT lose the ability to compile via IDEA

andimarek:
@bbakerman @dfa1 you need to increate the compiler memory setting ("Shared heap size") to something higher than the default:

Screenshot 2025-08-07 at 7 15 41 pm

This will solve your compile errors and will make it quite fast ... much faster than build with gradle

PR #4086, opened 2025-08-16
Prohibit Map.of and provide alternative methods

### LINKED ISSUE #3989: Ban Map.of

dondonz:
Fixes #3989

Turns out Map.of doesn't guarantee consistent order when serialized, which is easy to forget. This PR adds an ArchUnit rule to ban it.

Additionally I created LinkedHashMap equivalents of Map.of to make things easier. It is verbose but intended to be a copy of the original Map methods.

Sample error message for ArchUnit test
```
Map.of() usage detected. Please use LinkedHashMap instead for consistent serialization order:
- Method <graphql.normalized.ExecutableNormalizedOperationToAstCompiler.compileToDocument(graphql.schema.GraphQLSchema, graphql.language.OperationDefinition$Operation, java.lang.String, java.util.List, graphql.normalized.VariablePredicate)> calls method <java.util.Map.of()> in (ExecutableNormalizedOperationToAstCompiler.java:103)
- Method <graphql.normalized.ExecutableNormalizedOperationToAstCompiler.compileToDocumentWithDeferSupport(graphql.schema.GraphQLSchema, graphql.language.OperationDefinition$Operation, java.lang.String, java.util.List, graphql.normalized.VariablePredicate)> calls method <java.util.Map.of()> in (ExecutableNormalizedOperationToAstCompiler.java:154)

```

### DIFF
--- a/src/main/java/graphql/util/LinkedHashMapFactory.java
+++ b/src/main/java/graphql/util/LinkedHashMapFactory.java
@@ -1,0 +1,344 @@
+package graphql.util;
+
+import graphql.Internal;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Factory class for creating LinkedHashMap instances with insertion order preservation.
+ * Use this instead of Map.of() to ensure consistent serialization order.
+ * <p>
+ * This class provides static factory methods similar to Map.of() but returns mutable LinkedHashMap
+ * instances that preserve insertion order, which is important for consistent serialization.
+ */
+@Internal
+public final class LinkedHashMapFactory {
+
+    private LinkedHashMapFactory() {
+        // utility class
+    }
+
+    /**
+     * Returns an empty LinkedHashMap.
+     *
+     * @param <K> the key type
+     * @param <V> the value type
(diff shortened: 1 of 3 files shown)


src/main/java/graphql/util/LinkedHashMapFactory.java
```diff
@@ -0,0 +1,344 @@
...
+
+/**
+ * Factory class for creating LinkedHashMap instances with insertion order preservation.
+ * Use this instead of Map.of() to ensure consistent serialization order.
+ * <p>
+ * This class provides static factory methods similar to Map.of() but returns mutable LinkedHashMap
+ * instances that preserve insertion order, which is important for consistent serialization.
+ */
+@Internal
+public final class LinkedHashMapFactory {
```
dondonz, line 16: This is verbose. Originally I was only going to have one method (the last one) with variable length parameters, but saw the original Map.java had all of these extra methods for performance.

I don't think the difference is massive for our purposes, but I thought I'd follow the original

bbakerman (approved):

PR #4097, opened 2025-08-20
Draft - Update to Shadow 9.x and configure special task

dondonz:
With the major Shadow 9.x update, the fat JAR didn't seem to get created for jmh. Found that out when our jmh pipeline failed, being unable to find the main class to start the tests

This might be the config to set it right. I'll come back to this later


dondonz:
I will check after this commit is merged, that our performance pipeline completes successfully

dondonz:
Something's still wrong: the result is an empty list. Revert & will investigate

PR #4135, opened 2025-10-11
Memoize ResultPath level

AlexandreCarlton:
In our new `PerLevelDataLoaderDispatchStrategy`, we incorporate the level of the current field (through the `ResultPath`) to detemrine when we should dispatch `DataLoader`s. For:

 - a deep enough level
 - a large number of fields

this can become quite taxing.

To optimise this, we memoize the level by calculating it in the constructor, which should have a negligible cost.


andimarek:
I have the same change on another branch 👍😃

PR #4149, opened 2025-10-23
Avoid List and String allocation in inRightLocation

kilink:
The inRightLocation helper was recently optimized to remove Stream usage, however creation of the intermediary List is not necessary at all, and String allocations can be avoided by using equalsIgnoreCase.

Additionally, make some other methods static where possible.


bbakerman (approved):
Thanks for that

PR #4150, opened 2025-10-24
Fix nullability annotations for mapAndDropNulls

kilink:
ImmutableKit is annotated with NullMarked, but the mapper passed to mapAndDropNulls may return null.


bbakerman (approved):
Thanks

PR #4151, opened 2025-10-24
Optimize GraphqlTypeComparators

kilink:
- Return singleton Comparator from byNameAsc
- Use ImmutableList.sortedCopyOf to avoid extra List allocation in sortTypes


bbakerman (approved):
Thanks

PR #4152, opened 2025-10-24
Improve FpKit.concat, presize collections / maps

kilink:
FpKit.concat now sizes the resulting ArrayLists precisely. Additionally, update some more locations where we can easily presize Maps.


bbakerman (approved):
Thanks

PR #4154, opened 2025-10-26
Fix FpKit.concat parameters

kilink:
Make FpKit.concat parameters contravariant, so that it can be used in more call-sites; update a few such places that didn't work before in various Definition classes.


kilink:
Note: FpKit has this issue with several other methods, can be fixed in a follow up if deemed necessary.

Issue #4179, opened 2025-11-25
DataFetchingEnvironment getDataLoader doesn't match DataLoader signature


ctbarbour:
**Describe the bug**

[`DataFetchingEnvironment.getDataLoader<K, V>()`](https://github.com/graphql-java/graphql-java/blob/ed0791c75c681aa4cc750ca7a359bdb32225c1bd/src/main/java/graphql/schema/DataFetchingEnvironment.java#L241) has type parameters that are implicitly non-null due to [`@NullMarked`](https://github.com/graphql-java/graphql-java/blob/ed0791c75c681aa4cc750ca7a359bdb32225c1bd/src/main/java/graphql/schema/DataFetchingEnvironment.java#L31) but [`DataLoader<K, V extends @Nullable Object>`](https://github.com/graphql-java/java-dataloader/blob/918203c03acb041426171ae8dc672d88174747dd/src/main/java/org/dataloader/DataLoader.java#L72) explicitly allows nullable V. This creates a type mismatch when calling from Kotlin. As of Kotlin 2.1.0, Kotlin compiler treats JSpecify types as null-safe types. 

**To Reproduce**

From Kotlin, attempt to get a DataLoader with a nullable value type:

```kotlin
// Kotlin sees getDataLoader as requiring V : Any
val loader: DataLoader<String, Player?>? = env.getDataLoader("PlayerLoader")
// Error: Type argument is not within its bounds: must be subtype of 'Any'
```

dondonz:
Adding keep open - this should be fixed

bbakerman:
I think this is fixed

https://github.com/graphql-java/graphql-java/commit/28e1e7a96583416c5fd4c594160ac9446701fa9c

https://github.com/graphql-java/graphql-java/pull/4180

I am going to close this

PR #4180, opened 2025-11-25
Getdataloader type bounds

ctbarbour:
The @NullMarked annotation on DataFetchingEnvironment makes type parameters implicitly non-null, but java-dataloader's DataLoader declares V extends @Nullable Object to allow nullable values.

This mismatch prevents Kotlin consumers from expressing nullable value types without unsafe casts when returning CompletableFuture from data fetchers.

Add explicit V extends @Nullable Object bound to match DataLoader's type signature, enabling proper nullability propagation to Kotlin.

https://github.com/graphql-java/graphql-java/issues/4179


bbakerman (approved):
Thanks for this.

Moving to JSpecify is  an interesting exercise because while it "allows for more precise nullability specifications" its only the Kotlin compiler that truly enforces it.  

So this mismatch for example survives a Java compile but not a Kotlin one.

I wonder if we are missing some Java setting that would catch this earlier

PR #4184, opened 2025-12-04
JSpecify big wave 2

dondonz:
We have many classes left to go for JSpecify, so rather than one class at a time I will do the remainder in large bundles. This adds annotations on ~50 classes.

The general recipe here is
* Use IntelliJ's infer nullity tool, which is a start and covers the basics
* AI prompt to read JavaDoc, inspect nearby methods, with self-verification using NullAway
* And finally a manual check from me, as some items are hard to infer nullity without understanding the GraphQL specification or the codebase

What will be in a future PR: better Kotlin tooling as reported recently in https://github.com/graphql-java/graphql-java/issues/4179

### DIFF
--- a/src/main/java/graphql/schema/GraphQLSchema.java
+++ b/src/main/java/graphql/schema/GraphQLSchema.java
@@ -184,7 +184,7 @@ public class GraphQLSchema {
         ImmutableMap.Builder<String, ImmutableList<GraphQLObjectType>> interfaceMapBuilder = ImmutableMap.builder();
         for (Map.Entry<String, ImmutableList<String>> entry : finalInterfaceNameMap.entrySet()) {
             ImmutableList<GraphQLObjectType> objectTypes = map(entry.getValue(),
-                    name -> (GraphQLObjectType) finalTypeMap.get(name));
+                    name -> (GraphQLObjectType) assertNotNull(finalTypeMap.get(name)));
             interfaceMapBuilder.put(entry.getKey(), objectTypes);
         }
         ImmutableMap<String, ImmutableList<GraphQLObjectType>> finalInterfaceMap = interfaceMapBuilder.build();


.claude/commands/jspecify-annotate.md
```diff
@@ -0,0 +1,25 @@
+I have already asked IntelliJ to infer nullity on this class. Can you help me make this more accurate.
```
dondonz: I originally tinkered with Claude Code, happy to move this to another directory

src/main/java/graphql/ExceptionWhileDataFetching.java
```diff
@@ -16,13 +18,14 @@
  * This graphql error will be used if a runtime exception is encountered while a data fetcher is invoked
  */
 @PublicApi
+@NullMarked
 public class ExceptionWhileDataFetching implements GraphQLError {
```
dondonz, line 22: The nullability of these fields depends on a combination of the code and the GraphQL specification

* Message is not nullable from https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format
> Every error must contain an entry with the key "message" with a string description of the error intended for the developer as a guide to understand and correct the error.
* Path is not nullable, there is an assert not null in the constructor below
* Exception is not nullable, assert in constructor
* Locations is at least a list, but not null
* Extensions can be nullable from https://spec.graphql.org/draft/#sec-Extensions
> The "extensions" entry in an [execution result](https://spec.graphql.org/draft/#execution-result) or [request error result](https://spec.graphql.org/draft/#request-error-result), if set, must have a map as its value.

src/main/java/graphql/ExecutionResult.java
```diff
@@ -22,16 +27,16 @@ public interface ExecutionResult {
      *
      * @return the data in the result or null if there is none
      */
-    <T> T getData();
+    <T> @Nullable T getData();
```
dondonz, line 30: Annotation inferred from the javadoc (and spec too)

src/main/java/graphql/ExecutionResult.java
```diff
@@ -42,14 +47,14 @@ public interface ExecutionResult {
     /**
      * @return a map of extensions or null if there are none
      */
-    Map<Object, Object> getExtensions();
+    @Nullable Map<Object, Object> getExtensions();
```
dondonz, line 50: Inferred from javadoc (and also spec)

src/main/java/graphql/ExecutionResult.java
```diff
@@ -88,6 +93,7 @@ static Builder<?> newExecutionResult() {
         return ExecutionResultImpl.newExecutionResult();
     }
 
+    @NullUnmarked
```
dondonz, line 96: All builders are NullUnmarked

src/main/java/graphql/GraphQL.java
```diff
@@ -559,14 +559,14 @@ private PreparsedDocumentEntry parseAndValidate(AtomicReference<ExecutionInput>
 
         ParseAndValidateResult parseResult = parse(executionInput, graphQLSchema, instrumentationState);
         if (parseResult.isFailure()) {
-            return new PreparsedDocumentEntry(parseResult.getSyntaxException().toInvalidSyntaxError());
+            return new PreparsedDocumentEntry(assertNotNull(parseResult.getSyntaxException(), "Parse result syntax exception cannot be null when failed").toInvalidSyntaxError());
```
dondonz, line 562: A similar change now has to happen in a bunch of places: now we must be asserting not null on selected lines or else you'll get a NullAway compile warning because the nullity is not clear at this point in the execution

src/main/java/graphql/GraphQL.java
```diff
@@ -599,7 +599,7 @@ private List<ValidationError> validate(ExecutionInput executionInput, Document d
         validationCtx.onDispatched();
 
         Predicate<Class<?>> validationRulePredicate = executionInput.getGraphQLContext().getOrDefault(ParseAndValidate.INTERNAL_VALIDATION_PREDICATE_HINT, r -> true);
-        Locale locale = executionInput.getLocale() != null ? executionInput.getLocale() : Locale.getDefault();
+        Locale locale = executionInput.getLocale();
```
dondonz, line 602: And now there are IDE alerts of null checks we can remove. When we are done with all JSpecify annotations on public API classes, we will find many more null checks to remove

src/main/java/graphql/GraphQLError.java
```diff
@@ -20,18 +22,19 @@
...
  */
 @PublicApi
+@NullMarked
 public interface GraphQLError extends Serializable {
 
     /**
      * @return a description of the error intended for the developer as a guide to understand and correct the error
      */
-    String getMessage();
+    @Nullable String getMessage();
```
dondonz: TODO for me - message in another ExceptionWhileDataFetching is not nullable, does this need to be consistent?

src/main/java/graphql/GraphqlErrorException.java
```diff
@@ -1,24 +1,28 @@
...
- * A base class for graphql runtime exceptions that also implement {@link graphql.GraphQLError} and can be used
+ * A base class for graphql runtime exceptions that also implement {@link GraphQLError} and can be used
  * in a general sense direct or have specialisations made of it.
  * <p>
- * This is aimed amongst other reasons at Kotlin consumers due to https://github.com/graphql-java/graphql-java/issues/1690
+ * This is aimed amongst other reasons at Kotlin consumers due to <a href="https://github.com/graphql-java/graphql-java/issues/1690">...</a>
  * as well as being a way to share common code.
  */
 @PublicApi
+@NullMarked
```
dondonz, line 20: TODO for me - message in another ExceptionWhileDataFetching is not nullable, does this need to be consistent?

src/main/java/graphql/ParseAndValidateResult.java
```diff
@@ -40,7 +44,7 @@ public boolean isFailure() {
     /**
      * @return the parsed document or null if it's syntactically invalid.
      */
-    public Document getDocument() {
+    public @Nullable Document getDocument() {
```
dondonz, line 47: Inferred from the Javadoc

src/main/java/graphql/SerializationError.java
```diff
@@ -38,7 +41,7 @@ public String getMessage() {
     }
 
     @Override
-    public List<SourceLocation> getLocations() {
+    public @Nullable List<SourceLocation> getLocations() {
```
dondonz, line 44: TODO for me: double check if locations is consistent with ExceptionWhileDataFetching

GraphQLError nullity also impacts all other implementing classes like this

src/main/java/graphql/analysis/FieldComplexityEnvironment.java
```diff
@@ -4,19 +4,22 @@
...
 public class FieldComplexityEnvironment {
     private final Field field;
     private final GraphQLFieldDefinition fieldDefinition;
     private final GraphQLCompositeType parentType;
-    private final FieldComplexityEnvironment parentEnvironment;
+    private final @Nullable FieldComplexityEnvironment parentEnvironment;
     private final Map<String, Object> arguments;
 
-    public FieldComplexityEnvironment(Field field, GraphQLFieldDefinition fieldDefinition, GraphQLCompositeType parentType, Map<String, Object> arguments, FieldComplexityEnvironment parentEnvironment) {
+    public FieldComplexityEnvironment(Field field, GraphQLFieldDefinition fieldDefinition, GraphQLCompositeType parentType, Map<String, Object> arguments, @Nullable FieldComplexityEnvironment parentEnvironment) {
```
dondonz, line 22: It can be nullable because a method that calls this constructor may pass in a null value for parentEnvironment

src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
```diff
@@ -79,20 +80,20 @@ public MaxQueryComplexityInstrumentation(int maxComplexity, FieldComplexityCalcu
     }
 
     @Override
-    public @Nullable CompletableFuture<InstrumentationState> createStateAsync(InstrumentationCreateStateParameters parameters) {
+    public CompletableFuture<InstrumentationState> createStateAsync(InstrumentationCreateStateParameters parameters) {
```
dondonz, line 83: I think these `@Nullable` annotations were not correct initially. This method always will return a completed future

src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
```diff
@@ -79,20 +80,20 @@ public MaxQueryComplexityInstrumentation(int maxComplexity, FieldComplexityCalcu
...
         return CompletableFuture.completedFuture(new State());
     }
 
     @Override
-    public @Nullable InstrumentationContext<List<ValidationError>> beginValidation(InstrumentationValidationParameters parameters, InstrumentationState rawState) {
+    public InstrumentationContext<List<ValidationError>> beginValidation(InstrumentationValidationParameters parameters, InstrumentationState rawState) {
```
dondonz, line 88: Same as above, I think this was incorrectly marked as nullable initially

src/main/java/graphql/analysis/MaxQueryComplexityInstrumentation.java
```diff
@@ -79,20 +80,20 @@ public MaxQueryComplexityInstrumentation(int maxComplexity, FieldComplexityCalcu
...
         State state = ofState(rawState);
         // for API backwards compatibility reasons we capture the validation parameters, so we can put them into QueryComplexityInfo
         state.instrumentationValidationParameters.set(parameters);
         return noOp();
     }
 
     @Override
-    public @Nullable InstrumentationContext<ExecutionResult> beginExecuteOperation(InstrumentationExecuteOperationParameters instrumentationExecuteOperationParameters, InstrumentationState rawState) {
+    public InstrumentationContext<ExecutionResult> beginExecuteOperation(InstrumentationExecuteOperationParameters instrumentationExecuteOperationParameters, InstrumentationState rawState) {
```
dondonz, line 96: Same as above, I think this was incorrectly marked as nullable initially

src/main/java/graphql/analysis/MaxQueryDepthInstrumentation.java
```diff
@@ -48,7 +49,7 @@ public MaxQueryDepthInstrumentation(int maxDepth, Function<QueryDepthInfo, Boole
     }
 
     @Override
-    public @Nullable InstrumentationContext<ExecutionResult> beginExecuteOperation(InstrumentationExecuteOperationParameters parameters, InstrumentationState state) {
+    public InstrumentationContext<ExecutionResult> beginExecuteOperation(InstrumentationExecuteOperationParameters parameters, InstrumentationState state) {
```
dondonz, line 51: Same as above, I think this was incorrectly marked as nullable initially

src/main/java/graphql/relay/Connection.java
```diff
@@ -1,25 +1,28 @@
...
+ * <p>
+ * See <a href="https://relay.dev/graphql/connections.htm">https://relay.dev/graphql/connections.htm</a>
  */
 @PublicApi
+@NullMarked
 public interface Connection<T> {
 
     /**
-     * @return a list of {@link graphql.relay.Edge}s that are really a node of data and its cursor
+     * @return a list of {@link graphql.relay.Edge}s that contain a node of data and its cursor. Can be null as defined in the spec.
```
dondonz, line 20: From the specification

src/main/java/graphql/relay/DefaultConnectionCursor.java
```diff
@@ -2,16 +2,19 @@
...
 
 @PublicApi
+@NullMarked
 public class DefaultConnectionCursor implements ConnectionCursor {
 
     private final String value;
 
     public DefaultConnectionCursor(String value) {
-        Assert.assertTrue(value != null && !value.isEmpty(), "connection value cannot be null or empty");
+        Assert.assertTrue(!value.isEmpty(), "connection value cannot be null or empty");
```
dondonz, line 17: Remove null check

src/main/java/graphql/relay/DefaultPageInfo.java
```diff
@@ -2,32 +2,34 @@
 
 
 import graphql.PublicApi;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
 
 import java.util.Objects;
 
 @PublicApi
+@NullMarked
```
dondonz, line 11: A wider question: these Relay classes are quite old and never got much love. Do we want to retain these classes at all?

src/main/java/graphql/schema/GraphQLScalarType.java
```diff
@@ -29,34 +32,34 @@
...
     private final ScalarTypeDefinition definition;
     private final ImmutableList<ScalarTypeExtensionDefinition> extensionDefinitions;
     private final DirectivesUtil.DirectivesHolder directivesHolder;
-    private final String specifiedByUrl;
+    private final @Nullable String specifiedByUrl;
 
     @Internal
     private GraphQLScalarType(String name,
-                              String description,
+                              @Nullable String description,
```
dondonz, line 56: By definition in the specification, the description, and specified by URL can be null.

src/main/java/graphql/schema/GraphQLSchema.java
```diff
@@ -63,7 +65,7 @@ public class GraphQLSchema {
     private final SchemaDefinition definition;
     private final ImmutableList<SchemaExtensionDefinition> extensionDefinitions;
     private final String description;
-    private final GraphQLCodeRegistry codeRegistry;
+    private final @Nullable GraphQLCodeRegistry codeRegistry;
```
dondonz, line 72: See previous discussion in https://github.com/graphql-java/graphql-java/pull/4098#discussion_r2297295623 (this PR supercedes the older one)

dondonz: If we want to remove the nullable annotation here, we'd also need to change another callsite that assumes this parameter to be null

src/main/java/graphql/schema/PropertyDataFetcher.java
```diff
@@ -35,10 +37,11 @@
  * @see graphql.schema.DataFetcher
  */
 @PublicApi
+@NullMarked
 public class PropertyDataFetcher<T> implements LightDataFetcher<T> {
 
-    private final String propertyName;
-    private final Function<Object, Object> function;
+    private final @Nullable String propertyName;
+    private final @Nullable Function<Object, Object> function;
```
dondonz, line 44: See the constructors below, it is possible for either field to be nullable

src/main/java/graphql/util/Anonymizer.java
```diff
@@ -926,8 +927,7 @@ private static Type replaceTypeName(Type type, String newName) {
         } else if (type instanceof NonNullType) {
             return NonNullType.newNonNullType(replaceTypeName(((NonNullType) type).getType(), newName)).build();
         } else {
-            graphql.Assert.assertShouldNeverHappen();
-            return null;
```
dondonz, line 929: This throws, so the null is never reached

src/main/java/graphql/validation/ValidationError.java
```diff
@@ -15,11 +18,12 @@
...
 
 @PublicApi
+@NullMarked
 public class ValidationError implements GraphQLError {
 
     private final List<SourceLocation> locations = new ArrayList<>();
-    private final String description;
-    private final ValidationErrorClassification validationErrorType;
+    private final @Nullable String description;
+    private final @Nullable ValidationErrorClassification validationErrorType;
```
dondonz, line 27: TODO for me: check alongside other GraphQLError classes if we want to keep the description nullable

Double check ValidationErrorClassification - ErrorClassification is not nullable in GraphQLError

src/test/groovy/graphql/archunit/JSpecifyAnnotationsCheck.groovy
```diff
@@ -323,9 +276,7 @@ class JSpecifyAnnotationsCheck extends Specification {
             "graphql.util.TraverserContext",
             "graphql.util.TreeTransformer",
             "graphql.util.TreeTransformerUtil",
-            "graphql.validation.ValidationError",
-            "graphql.validation.ValidationErrorClassification",
-            "graphql.validation.ValidationErrorType",
+            // These classes will not be public API later, exempt here while marked as experimental
```
dondonz, line 278: There is an edge case where some classes have been annotated with "Experimental API" for the purposes of the defer work. They won't become public API later so I made a note in the JSpecify exemption rule

src/main/java/graphql/GraphQLError.java
dondonz: This ought to be non-nullable given the specification uses the word "MUST"

> Every error must contain an entry with the key "message" with a string description of the error intended for the developer as a guide to understand and correct the error.

https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format

src/main/java/graphql/GraphqlErrorException.java
dondonz: Message changed to non-nullable in GraphQLError

src/main/java/graphql/SerializationError.java
dondonz: Locations are nullable, note the word "if"

> If an error can be associated to a particular point in the requested GraphQL document, it should contain an entry with the key "locations"

https://spec.graphql.org/draft/#sec-Errors.Error-Result-Format

src/main/java/graphql/GraphqlErrorException.java
```diff
@@ -1,24 +1,28 @@
...
 public class GraphqlErrorException extends GraphQLException implements GraphQLError {
 
-    private final List<SourceLocation> locations;
-    private final Map<String, Object> extensions;
-    private final List<Object> path;
+    private final @Nullable List<SourceLocation> locations;
```
dondonz, line 23: The builder creates a non-nullable list, but that's only if the builder is called

src/main/java/graphql/validation/ValidationError.java
dondonz: Change description to be non-nullable, as it's the legacy name for "message" in GraphQLError

src/main/java/graphql/GraphQLContext.java
```diff
@@ -66,7 +71,7 @@ public GraphQLContext delete(Object key) {
      *
      * @return a value or null
      */
-    public <T> T get(Object key) {
+    public <T> @Nullable T get(Object key) {
```
dondonz, line 74: TODO: another case of the nullable generic issue

src/main/java/graphql/TypeResolutionEnvironment.java
```diff
@@ -52,7 +55,7 @@ public TypeResolutionEnvironment(TypeResolutionParameters parameters) {
      * @return the object that needs to be resolved into a specific graphql object type
      */
     @SuppressWarnings("unchecked")
-    public <T> T getObject() {
+    public @Nullable <T> T getObject() {
```
dondonz, line 58: TODO: another generic nullable check

src/main/java/graphql/GraphQLContext.java
dondonz: Update - all values stored in the GraphQLContext are non-null. It's only going to return null if the key is not found. Let's keep it like this.

src/main/java/graphql/TypeResolutionEnvironment.java
dondonz: Edit: object should also be nullable

dondonz: I updated the prompt with more information on generics

dondonz:
I had a fix a bunch of tests when I changed Validation Error messages to be non-nullable

This is as per the spec - messages in GraphQL errors must be present

dondonz:
Special note for release notes: ValidationError technically has a breaking change, now a description must be not-nullable, which is consistent with the specification

Previously there was a mismatch that ValidationError didn't enforce description. However that doesn't make much sense, if a user receives a validation error they should also receive a description to help explain and diagnose the error

.claude/commands/jspecify-annotate.md
```diff
@@ -0,0 +1,59 @@
+The task is to annotate public API classes (marked with `@PublicAPI`) with JSpecify nullability annotations.
```
dondonz, line 1: This prompt is checked in so subagents can all share this context

This prompt will be deleted after this series of JSpecify annotations is completed

Issue #4196, opened 2025-12-20
GraphQLSchema.FastBuilder for High-Performance Schema Construction


rstata:
# Proposal: GraphQLSchema.FastBuilder for High-Performance Schema Construction

## Human-written overview

Airbnb's [Viaduct](https://github.com/airbnb/viaduct) platform is written on top of GraphQL Java.  Airbnb's schema has grown to be over 25K types with over 120K "members" (fields and enumeration values).  At this scale the construction of GraphQL Java's `GraphQLSchema` object has become a bottleneck.  Futher, at both build time and runtime Viaduct creates the same schema (and variants of it) many times, multiplying the performance issues we've seen in this construction.

In the Viaduct project itself we [addressed this problem](https://github.com/airbnb/viaduct/issues/254) in part by developing a file format that effectively eliminates all parsing-related overhead.  By our measurements on `large-schema-4`:

| Benchmark           | Avg Time | Memory    |
|---------------------|----------|-----------|
| From SDL Text       | 481 ms   | 626 MB/op |
| From optimized file | 59 ms    | 166 MB/op |

These numbers measure the time and space required to effectively create a `TypeDefinitionRegistry` - an 8x speedup and close to 4x reduction in space.

However, it turns out that creating a `TypeDefinitionRegistry` consumes only half the time it takes overall to construct a `GraphQLSchema` object, so the improvements just reported resulted in just a 50% improvement.  We were looking for something more significant, so we took a look at optimizing `GraphQLSchema.Builder`.  This Builder is very general, in that it allows you to "clear" the builder mid-construction, to update the code-registry factory during construction, and to add just the "roots" of the schema versus adding every single type defined by the schema.  Also, the Builder _always_ performs schema validation, which for systems like Viaduct that read the same schema over and over again is redundant and somewhat expensive.

We determined that the current semantics of this Builder were not something we should change, but also not semantics we could effectively optimize.  So instead we've developed an alternative "FastBuilder", which is more restrictive in its semantics than is the current Builder, but in turn is significantly less resource intensive.  On that same large schema:

| Metric | Standard Builder | FastBuilder | Improvement |
|--------|------------------|-------------|-------------|
| **Build Time** | 1,933 ms | 278 ms | **6.96x faster** |
| **Memory Allocated** | 2.71 GB/op | 411 MB/op | **85% reduction** |
| **GC Time (total)** | 258 ms | 183 ms | 29% reduction |

[...]

PR #4197, opened 2025-12-20
feat: GraphQLSchema.FastBuilder for High-Performance Schema Construction

rstata:
Defines a new `GraphQLSchema` builder class that is more restrictive than the existing one but reduces time and space requirements by more than a factor of five.  See Issue #4196 for details.

### DIFF
--- a/src/test/resources/large-schema-federated-2.graphqls
+++ b/src/test/resources/large-schema-federated-2.graphqls
@@ -1,0 +1,33419 @@
+"This directive allows results to be deferred during execution"
+directive @defer(
+    "Deferred behaviour is controlled by this argument"
+    if: Boolean! = true,
+    "A unique label that represents the fragment being deferred"
+    label: String
+  ) on FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+"Marks the field, argument, input field or enum value as deprecated"
+directive @deprecated(
+    "The reason for the deprecation"
+    reason: String! = "No longer supported"
+  ) on FIELD_DEFINITION | ARGUMENT_DEFINITION | ENUM_VALUE | INPUT_FIELD_DEFINITION
+
+"Subject to change without notice."
+directive @experimental on FIELD_DEFINITION | ARGUMENT_DEFINITION | ENUM_VALUE | INPUT_FIELD_DEFINITION
+
+"This directive disables error propagation when a non nullable field returns null for the given operation."
+directive @experimental_disableErrorPropagation on QUERY | MUTATION | SUBSCRIPTION
+
+directive @extends on OBJECT | INTERFACE
+
+directive @external on OBJECT | FIELD_DEFINITION
+
+"For federated schema composition. When a service wants to add a new field to an existing type +, it may use this directive to temporarily hide the field from the composed graph.This avoid breaking composition until others users of the value type have added the field."
+directive @inaccessible on FIELD_DEFINITION | ENUM | ENUM_VALUE
+
+"Directs the executor to include this field or fragment only when the `if` argument is true"
+directive @include(
+    "Included when true."
+    if: Boolean!
+  ) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+
+directive @key(fields: String!) repeatable on OBJECT | INTERFACE
+
+"The @nonNull directive gives clients the ability to treat a field as if it were marked as Non-Null (!) in the schema. When this directive is placed after a field in a query, that field isconsidered not nullable. If a null value would be returned for that field, the entire parents will insteadbe returned as null. The null value will propagate in the same way as a Non-Null type."
+directive @nonNull on FIELD
+
+"Indicates an Input Object is a OneOf Input Object."
+directive @oneOf on INPUT_OBJECT
+
+"For federated schema composition. When a service wants to expose a field that another service already exposes, it may use this directive to claim the field."
+directive @override(
+    "Service to override the field from."
+    from: String,
+    "If a label is defined, the directive becomes conditional"
+    label: String
+  ) on FIELD_DEFINITION
+
+directive @provides(fields: String!) on FIELD_DEFINITION
+
+directive @requestContext(fields: [String!]!) on FIELD_DEFINITION
+
+directive @requires(fields: String!) on FIELD_DEFINITION
+
+directive @scope(id: String) repeatable on FIELD_DEFINITION
+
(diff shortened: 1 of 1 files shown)


andimarek:
hey @rstata ... thanks a lot of this PR.

I think the fundamental idea and your use case makes a lot of sense and the implementation also looks acceptable broadly.

Is this PR shape what would solve your problem? Do you want to work through this PR to get it merged or you would prefer opening a new one?

rstata:
This PR works for us. We would use the `GraphQLSchema.FastBuilder` directly because we'd want to use it in conjunction with a (fast to read) binary-schema representation we've developed for Viaduct.

You should take a look at [the following](https://github.com/graphql-java/graphql-java/pull/4197/files#diff-ed83b75efc855050e2aa3dbd88ca91ab7e0e725bb44ac418d9de703ac981481d):
```
/**
 * A schema generator that uses GraphQLSchema.FastBuilder for improved performance.
 * This is intended for benchmarking and performance testing purposes.
 */
@Internal
public class FastSchemaGenerator {
```
This class converts a `TypeDefinitionRegistry` into a `GraphQLSchema` using the `FastBuilder`.  On the one hand, this could be of interest to the broader GraphQL Java community.  On the other hand, the Viaduct use case isn't for executable schemas (our initial use cases are for build-time use cases) and i'm reluctant to vouch for the correctness of the wiring part.  Maybe we merge with the above limitation, and then remove that limitation if there's demand and also more confidence?

andimarek:
@rstata one thought: we don't need the FindDetachedTypes traversal I think. Additionally types don't have to be restricted, but can contain all types (minus the root types of query/mutation/subscription)

andimarek:
question @rstata: is this builder to be used programmatically directly or only every via FastSchemaGenerator? How are you using it?

rstata:
> question @rstata: is this builder to be used programmatically directly or only every via FastSchemaGenerator? How are you using it?

Our intent is to use it directly.  We have a binary file representation for schemas that is fast and memory-efficient to read.  Here is the performance for reading the large4 schema:

| Benchmark           | Avg Time |  Memory    |
|------------------|-----------|------------|
| TypeDefRegistry  | 481 ms    | 626 MB/op |
| Binary file             | 59 ms      | 166 MB/op |

The result of reading the binary file is something that is close in spirit to a type-definition registry, but is not a type-definition registry.  The fast builder was written expressly to support going from our TDR-like schema representation to a `GraphQLSchema` -- optimizing end-to-end startup time for us.  In fact, `FastSchemaGenerator` was written initially just so I could provide some apples-to-apples performance numbers, although having written it we see use cases for it in Viaduct as well.

rstata:
> @rstata one thought: we don't need the FindDetachedTypes traversal I think. Additionally types don't have to be restricted, but can contain all types (minus the root types of query/mutation/subscription)

just pushed a commit that removes the detached-type detection.  it does remove a lot of code (esp. a lot of testing code, because that was tested pretty extensively).  so this change is a win in terms of simplicity.  here is the impact on the benchmarks:

  | Benchmark                              |    Baseline        |   After Change  | Difference              |
  |------------------------------|----------------|----------------|--------------------|
  | CreateSchemaBenchmark      | 465.891 ms/op | 446.331 ms/op | -4.2% improved    |

rstata:
FastBuilder was not handling introspective types nor the scalars needed by them correctly.  The most recent commit ("Fixes") adds tests for these bugs, and fixes the bugs.  We also removed some tests for FastBuilder.addType (et al) to adjust for recent changes in its behavior.

rstata:
I've added another commit here because an airbnb reviewer noted that the classic builder and the new "fast" builder handled built-in directives using two separate lists of built-ins, which creates a maintenance burden (because changes need to be updated in two places).  This commit consolidates the lists into one place.

This PR maintains and also documents the behavior that says that the `@include` and `@skip` directives can be eliminated from a schema by calling `clear`, but that other built-ins (e.g., `@oneOf`) cannot be.  Is there good motivation for these different treatments of spec-defined directives?

andimarek:
> I've added another commit here because an airbnb reviewer noted that the classic builder and the new "fast" builder handled built-in directives using two separate lists of built-ins, which creates a maintenance burden (because changes need to be updated in two places). This commit consolidates the lists into one place.
> 
> This PR maintains and also documents the behavior that says that the `@include` and `@skip` directives can be eliminated from a schema by calling `clear`, but that other built-ins (e.g., `@oneOf`) cannot be. Is there good motivation for these different treatments of spec-defined directives?

@rstata the different treatment of skip/include vs others seems weird and not intentionally ... I will double check, thanks

andimarek:
@rstata I streamlined and simpllifed the directives handling: https://github.com/graphql-java/graphql-java/pull/4229

Please merge with master ... it should make things more clear.

rstata:
@andimarek - i've merged

rstata:
oops - i left a small fixed unstaged in the previous push, the latest commit fixes that.

andimarek:
@rstata I am happy to merge it. Your thoughts?

One question: Is FastSchemaGenerator useable in general? We could make it even public or experimental or are there edges cases with SDL based fast schema generation that you know of?

rstata:
@andimarek  - i do think `FastSchemaGenerator` should be public a experimental: while the original use-case for the fast builder at airbnb was to work in conjunction with our binary schema format, now we're seeing places where we want to go from text SDL to `GraphQLSchema` using the fast builder, and `FastSchemaGenerator` is good for that.

I made changes to make it more appropriate for public use.  the most important of these changes is that i added a switch to toggle validation which is set to do validation by default, which seems safer for a public api.  i also updated some documentation, added nullability annotations, and added some tests.  (regarding the validation switch, i originally drafted this as an additional argument to `makeExecutableSchema`, but decided to consolidate it into `SchemaGenerator.Options` using `@ExperimentalApi` -- i can revert to a spearate argument if desired.)

while making these changes i noticed that the new fast-builder does not do the type-resolver validation that the existing builder does.  when the `withValidation` switch is enabled, the fast-builder should do all validations that the existing one does, so this is a miss, so i added it.  the check is so inexpensive that i added it as non-optional to `FastBuilder` (ie, it runs regardless of whether full schema validation is requested).

andimarek:
@rstata looking good ... I agree with making the FastSchemaGenerator experimental, it will be very useful as most schema generation is done this way.

I am happy to merge it.

rstata:
@andimarek  - i'm ready to have it merged, thanks for your help!!

PR #4204, opened 2026-01-10
Fix ExecutableNormalizedField to respect GraphqlFieldVisibility

Samjin:
## Problem

`ExecutableNormalizedField.forEachFieldDefinition()` and `getOneFieldDefinition()` call `type.getField(fieldName)` directly, bypassing the schema's configured `GraphqlFieldVisibility`.

This is inconsistent with other parts of graphql-java (validation, execution) that respect field visibility, and causes issues when:

- A custom GraphqlFieldVisibility provides field definitions that differ from what's in the schema type (e.g., placeholder fields, virtual fields, or dynamically-computed fields)
- Application code accesses DataFetchingEnvironment.getSelectionSet().getFields() in a DataFetcher
- Normalization fails with "No field X found for type Y" because it bypasses the visibility

### Reproduction

1. Configure a custom `GraphqlFieldVisibility` that provides virtual/placeholder fields
2. Execute a query that includes a field only available through the visibility
3. Access `DataFetchingEnvironment.getSelectionSet().getFields()` in a DataFetcher
4. Normalization fails because `ExecutableNormalizedField.forEachFieldDefinition()` calls `type.getField()` directly

## Solution

Change the field lookup to use the schema's field visibility:

```java
// Before
type.getField(fieldName)

// After  
schema.getCodeRegistry().getFieldVisibility().getFieldDefinition(type, fieldName)

### DIFF
--- a/src/main/java/graphql/normalized/ExecutableNormalizedField.java
+++ b/src/main/java/graphql/normalized/ExecutableNormalizedField.java
@@ -198,9 +198,13 @@ public class ExecutableNormalizedField {
             return;
         }
 
+        var fieldVisibility = schema.getCodeRegistry().getFieldVisibility();
         for (String objectTypeName : objectTypeNames) {
             GraphQLObjectType type = (GraphQLObjectType) assertNotNull(schema.getType(objectTypeName));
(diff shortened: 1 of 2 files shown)


Samjin:
> Thanks for this


Hi bbakerman, can we please include this v26 beta? Let me know if you need a hand. Thanks

PR #4207, opened 2026-01-14
Call onExceptionHandled in ChainedInstrumentation

tinnou:
This is a miss from PR #4206, we needed to ensure that the new hook is being properly delegated to each instrumentation within a `ChainedInstrumentation`


tinnou:
@bbakerman sorry for the @, would you guys be able to review this one, it's a quick fast follow on #4206. I saw that @dondonz just cut 26 first beta 🎉 and would love for this to get in with #4206.

PR #4209, opened 2026-01-19
Fix schema transformation and Field Visibility  for complex deletion cases

andimarek:
For certain complex schema deletion cases the current schema transformer fails.

This PR introduces a new transform method `transformSchemaWithDeletes` which should be used specifically for these cases.

The FieldVisibilityTransformer is also changed internally to use this new method.

The existing schema transformation was unchanged to keep existing behavior as the new `transformSchemaWithDeletes` changes slightly the traversal behavior.

The original bug was first reported in https://github.com/graphql-java/graphql-java/issues/4133, but we had additional reports after.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+# AI Agent Context for graphql-java
+
+This file provides context for AI assistants working with this codebase.
+
+## Test Execution
+
+When running tests, exclude the Java version-specific test tasks to avoid failures:
+
+```bash
+./gradlew test -x testWithJava17 -x testWithJava11 -x testng
+```


src/main/java/graphql/schema/GraphQLSchema.java
```diff
@@ -275,10 +275,10 @@ public GraphQLObjectType getIntrospectionSchemaType() {
      *
      * @return an immutable set of types that were explicitly added as additional types
      *
-     * @see Builder#additionalType(GraphQLType)
+     * @see Builder#additionalType(GraphQLNamedType)
      * @see Builder#additionalTypes(Set)
      */
-    public Set<GraphQLType> getAdditionalTypes() {
+    public Set<GraphQLNamedType> getAdditionalTypes() {
```
bbakerman, line 281: This is a breaking change.  Just in case you didnt realise.

src/main/java/graphql/schema/GraphQLSchema.java
```diff
@@ -808,7 +808,7 @@ public Builder codeRegistry(GraphQLCodeRegistry codeRegistry) {
          *
          * @see GraphQLSchema#getAdditionalTypes()
          */
-        public Builder additionalTypes(Set<GraphQLType> additionalTypes) {
+        public Builder additionalTypes(Set<? extends GraphQLNamedType> additionalTypes) {
             this.additionalTypes.addAll(additionalTypes);
             return this;
         }
```
bbakerman, line 814: also a breaking change

src/main/java/graphql/schema/GraphQLSchema.java
```diff
@@ -831,7 +831,7 @@ public Builder additionalTypes(Set<GraphQLType> additionalTypes) {
          * @see GraphQLSchema#getAdditionalTypes()
          * @see #additionalTypes(Set)
          */
-        public Builder additionalType(GraphQLType additionalType) {
+        public Builder additionalType(GraphQLNamedType additionalType) {
             this.additionalTypes.add(additionalType);
             return this;
         }
```
bbakerman, line 837: and again - thats ok -  just pointing it out

bbakerman (approved):

bbakerman:
I have labelled it as a breaking change because it it.  Its in the minor category of possible impact but none the less Java breaking

src/main/java/graphql/schema/GraphQLSchema.java
andimarek: Well .. they never contained not named types so far. Technically breaking, but I think nobody ever used it to add `[Foo]` instead of `Foo`

PR #4211, opened 2026-01-21
fix: treat empty operation name the same as null

RafeArnold:
if an empty operation name is given and multiple
operations are provided in the query, then the first operation is
 returned. not entirely sure why this would be the desired behaviour,
 when an exception is thrown for any other value that doesn't
 match an existing operation in the query.

### DIFF
--- a/src/main/java/graphql/language/NodeUtil.java
+++ b/src/main/java/graphql/language/NodeUtil.java
@@ -73,12 +73,13 @@ public class NodeUtil {
                 fragmentsByName.put(fragmentDefinition.getName(), fragmentDefinition);
             }
         }
-        if (operationName == null && operationsByName.size() > 1) {
+        boolean operationNameProvided = operationName != null && !operationName.isEmpty();
+        if (!operationNameProvided && operationsByName.size() > 1) {
             throw new UnknownOperationException("Must provide operation name if query contains multiple operations.");
         }
         OperationDefinition operation;
 
-        if (operationName == null || operationName.isEmpty()) {
+        if (!operationNameProvided) {
             operation = operationsByName.values().iterator().next();
         } else {
             operation = operationsByName.get(operationName);
--- a/src/test/groovy/graphql/language/NodeUtilTest.groovy
+++ b/src/test/groovy/graphql/language/NodeUtilTest.groovy
@@ -15,11 +15,14 @@ class NodeUtilTest extends Specification {
         ''')
 
         when:
-        NodeUtil.getOperation(doc, null)
+        NodeUtil.getOperation(doc, operationName)
 
         then:
         def ex = thrown(UnknownOperationException)
         ex.message == "Must provide operation name if query contains multiple operations."
+
+        where:
+        operationName << [null, '']
     }
 
     def "getOperation: when multiple operations are defined in the query and operation name doesn't match any of the query operations then it should throw UnknownOperationException"() {


andimarek:
@RafeArnold can you please give more context? What do you mean with previously? What is exactly the problem or error you are experiencing? I believe this code has like this forever and we need to have a good reason to change it.

RafeArnold:
when executing a graphql query that contains multiple operations, the operation to execute must be specified via the operationName parameter. currently, if the provided operation name does not match any of the operations in the query, an error is thrown. however there's an exception to this behaviour when the provided operation name is empty. in that case, the first found operation is returned instead. i dont understand what the motivation for this exception to the rule is. why would an empty operation name imply "just give me the first operation you find in the query"? id expect it to still throw an error explaining that there are no operations with an empty name.

andimarek:
> Why would an empty operation name imply "just give me the first operation you find in the query". 

We kinda agree with this in general, but we are wondering what is your use case? Can you provide more context? This code has been like this for 8+ years.

bbakerman:
I kinda agree that its weird that in the case of `""` it gives back the first.

Its trying to handle this case

```
{
   this
   valid
   operation
   has
   no 
   name
}
```

BUT its also a low level utitlity class.  I would love to see examples of how this gets invoked with `""` as an input.

We could fix it but I want to see the knock on effects.  What parts of other higher level code are calling this and can we get tests for that

RafeArnold:
maybe i was too hasty with my solution here (teaches me not to submit prs as im rushing to leave the office).

an alternative solution to an empty operation name retrieving the first operation arbitrarily when multiple operations are supplied is to treat the empty operation name in the exact same way as the null operation name. i.e. throw an exception explaining that an operation name must be supplied if multiple operations are present in the query. that way, the existing functionality remains that allows the library users to supply an empty operation name when there is only a single (named or unnamed) operation in the query, but the unfortunate side effect of the implementation that allows an empty operation name when there are multiple operations is removed. it seems very unlikely that any library users would be relying on this side effect, as it seems more likely to be a bug than a design decision.

the fix for this would be to simply add the same `operationName.isEmpty()` check to [the condition of line 76](https://github.com/graphql-java/graphql-java/blob/e860aff5292e1227e57e35f251b345cbc9a9ac9c/src/main/java/graphql/language/NodeUtil.java#L76).

would this proposed change be more acceptable?

RafeArnold:
i've pushed a new commit that implements my proposed change

RafeArnold:
reading the [GraphQL over HTTP spec](https://graphql.github.io/graphql-over-http/draft/#sec-GET), i see that "Setting the value of the operationName parameter to the empty string is equivalent to omitting the operationName parameter.", so my second pass at the fix brings it in line with the spec.

in terms of knock on effects, the method this PR alters is called with an operationName argument supplied directly via the `ExecutionInput` object that the user supplies to the high level `GraphQL.execute` and equivalent methods. so, if the user creates an ExecutionInput with an operationName field of an empty string, it would pass straight down into `NodeUtil.getOperation` unaltered. i'm happy to add a higher level test that shows this behaviour if you think it's worth it, @bbakerman.

RafeArnold:
is there something i can do to get this pr moved forwards?

andimarek:
@RafeArnold we are not against this fix in general, but this is an internal class of GJ which is like this for a lot of years.

Please give us more background info in which context you encountered this problem to help us decide if this is really the right fix.

Thanks

RafeArnold:
Thanks for the reply @andimarek. We encountered this issue while building a request validator for our GraphQL mocking tool. Our tool allows users to create mocks of their own GraphQL APIs and the validator allows them to check that the requests being sent to their mock APIs are valid, both according to their schema and the GraphQL specification. One of the checks in the validator is that the request provides a valid operation name when multiple operations are present in the request's query document. We use `NodeUtil.getOperation` to do the majority of this check's validation, but currently have to prefix the call to this method with a check for an empty operation name.

This issue can also be encountered outside of the validator by any request made to a mock GraphQL API, unless we prefix all entry points to the GraphQL engine with a similar check.

I'm assuming this spec non-compliance was never encountered naturally in the wild due to it being so niche. Seems unlikely that someone would attempt to send an empty operation name if their request contained multiple operations.

I hope that is enough info to assess whether this fix is appropriate. Let me know otherwise.

PR #4248, opened 2026-02-17
OneOf Inhabitability

jbellenger:
This PR implements the validation described at https://github.com/graphql/graphql-spec/pull/1211

More discussion is at the link above, but the core issue is around uninhabited OneOf types -- the property that a type can be defined in a way that a value cannot be generated for it.

The simplest example of an uninhabited type is this recursive OneOf:
```
input A @oneOf { a:A }
```

While this type definition is currently valid by the rules of OneOf types and circular input types, it violates the spirit of the [Circular References](https://spec.graphql.org/draft/#sec-Input-Objects.Circular-References) because it does not allow construction of a finite value. This uninhabitation property applies to any closed subgraph of OneOf types.

This PR adds a new validator for OneOf inhabitation. I've tried to follow the existing conventions, but any feedback on style or substance is welcome.

### DIFF
--- a/src/main/java/graphql/schema/validation/OneOfInputObjectRules.java
+++ b/src/main/java/graphql/schema/validation/OneOfInputObjectRules.java
@@ -4,10 +4,15 @@ import graphql.ExperimentalApi;
 import graphql.schema.GraphQLInputObjectField;
 import graphql.schema.GraphQLInputObjectType;
 import graphql.schema.GraphQLSchemaElement;
+import graphql.schema.GraphQLType;
 import graphql.schema.GraphQLTypeUtil;
 import graphql.schema.GraphQLTypeVisitorStub;
+import graphql.schema.GraphQLUnmodifiedType;
 import graphql.util.TraversalControl;
 import graphql.util.TraverserContext;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 import static java.lang.String.format;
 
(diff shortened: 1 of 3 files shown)


bbakerman (approved):
Thanks for this

src/main/java/graphql/schema/validation/OneOfInputObjectRules.java
```diff
@@ -19,6 +24,45 @@
...
+            if (GraphQLTypeUtil.isList(fieldType)) {
+                return true;
+            }
+            GraphQLUnmodifiedType namedFieldType = GraphQLTypeUtil.unwrapAll(fieldType);
+            if (!(namedFieldType instanceof GraphQLInputObjectType)) {
+                return true;
+            }
+            GraphQLInputObjectType inputFieldType = (GraphQLInputObjectType) namedFieldType;
+            if (!inputFieldType.isOneOf()) {
+                return true;
```
martinbonnin, line 57: @jbellenger was looking at this when implementing the validation rule in Apollo Kotlin and I believe this is not correct? 

The validation has to consider all `@oneOf` and `non-null` in the same pass? 

Example:
```graphql
    input A @oneOf { b: B }
    input B { a: A! } # here it's not oneOf but it still can't provide a finite value
```

jbellenger: 👋 righteous call out. This was from before I'd better understood the problem. I've revisited this in #4458, to true this up to graphql-js and the revised spec algorithm.

PR #4254, opened 2026-02-20
Add generic DataFetcherResult.newBuilder(T data) method

jord1e:
This is something that has always been on my mind, I usually put it in a utility method myself. Would be great if it makes 26.0

It turns:

```java
@MutationMapping
public DataFetcherResult<MyPayload> myField() {
  return DataFetcherResult.<MyPayload>newResult()
      .data(MyPayload.newBuilder().build())
      .localContext(...)
      .build();
}
```

into

```java
@MutationMapping
public DataFetcherResult<MyPayload> myField() {
  return DataFetcherResult.newResult(MyPayload.newBuilder().build())
      .localContext(...)
      .build();
}
```

This prevents you from having to write your own generic before newBuilder <>

Thank you for the work as always

### DIFF
--- a/.github/workflows/commit_performance_result.yml
+++ b/.github/workflows/commit_performance_result.yml
@@ -17,7 +17,7 @@ jobs:
   commitPerformanceResults:
     runs-on: ubuntu-latest
     steps:
-      - uses: aws-actions/configure-aws-credentials@v5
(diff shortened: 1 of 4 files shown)


bbakerman (approved):
Fair enough - thanks

PR #4257, opened 2026-02-22
Add JSpecify annotations to language package nodes - redo (12 more)

dondonz:
## Summary

Adds JSpecify nullability annotations (`@NullMarked`, `@NullUnmarked`, `@Nullable`) to the following `graphql.language` classes and interfaces, removing each from the `JSpecifyAnnotationsCheck` exemption list:

- `NodeTraverser`
- `NonNullType`
- `ObjectField`
- `ObjectTypeDefinition` / `ObjectTypeExtensionDefinition`
- `OperationDefinition`
- `OperationTypeDefinition`
- `PrettyAstPrinter`
- `SDLDefinition` / `SDLExtensionDefinition`
- `SDLNamedDefinition` / `TypeDefinition`

### Key decisions

Key interesting point - **`NamedNode.getName()` is `@Nullable`**
`OperationDefinition` can be anonymous (e.g. `{ user { name } }`) — the GraphQL spec makes the operation name optional. Pushing `@Nullable` onto `NamedNode.getName()` is the truthful representation of the interface contract.

**`TypeDefinition.getName()` overrides as non-null**
`TypeDefinition` extends both `SDLNamedDefinition` (non-null `getName()`) and `NamedNode` (`@Nullable getName()`). SDL type definitions always have a name, so `TypeDefinition` explicitly overrides `getName()` to return non-null. This prevents `@Nullable` from cascading into callers like `TypeDefinitionRegistry` that work with concrete schema types.

**`OperationDefinition.selectionSet` is non-null**
Legacy convenience constructors that passed `null` for `selectionSet` were deleted (they were only used in one test). The field, constructor, and `getSelectionSet()` are all non-null, matching the GraphQL spec requirement that every operation has a selection set.

**`OperationDefinition.operation` is `@Nullable`**
The shorthand query form (e.g. `{ user { name } }`) omits the `query` keyword, so `getOperation()` can legitimately return `null`.

**Builder inner classes are `@NullUnmarked`**
All builder static inner classes are annotated `@NullUnmarked` rather than annotating each field individually.

**`PrettyAstPrinter` private helpers**
`node(@Nullable Class startClass)`, `isEmpty(@Nullable ...)`, `nvl(@Nullable ...)`, `block(... @Nullable String separatorSingleLine, @Nullable String whenEmpty)`, and `spaced(@Nullable String... args)` / `join(String, @Nullable String...)` are nullable because null is actually passed at the call sites or because the method bodies already handle null gracefully.

### Cascading fixes

- **`OperationValidator`**: `getName()` is now `@Nullable` on `OperationDefinition`, so error message construction uses a pre-checked local variable or `Objects.toString(x, "")` fallback for the anonymous operation case.
- **`TypeDefinitionRegistry`**: two `Assert.assertNotNull(t.getName(), ...)` workarounds are removed now that `TypeDefinition.getName()` is non-null.

### DIFF
--- a/src/main/java/graphql/analysis/QueryTraverser.java
+++ b/src/main/java/graphql/analysis/QueryTraverser.java
@@ -166,10 +166,7 @@ public class QueryTraverser {
     }
 
     private GraphQLObjectType getRootTypeFromOperation(OperationDefinition operationDefinition) {
-        OperationDefinition.Operation op = operationDefinition.getOperation() != null
-                ? operationDefinition.getOperation()
-                : OperationDefinition.Operation.QUERY;
-        switch (op) {
+        switch (operationDefinition.getOperation()) {
             case MUTATION:
                 return assertNotNull(schema.getMutationType());
             case QUERY:
--- a/src/main/java/graphql/language/OperationDefinition.java
+++ b/src/main/java/graphql/language/OperationDefinition.java
@@ -34,7 +34,7 @@ public class OperationDefinition extends AbstractNode<OperationDefinition> imple
 
     private final @Nullable String name;
 
-    private final @Nullable Operation operation;
+    private final Operation operation;
     private final ImmutableList<VariableDefinition> variableDefinitions;
     private final DirectivesHolder directives;
     private final SelectionSet selectionSet;
@@ -45,7 +45,7 @@ public class OperationDefinition extends AbstractNode<OperationDefinition> imple
 
     @Internal
     protected OperationDefinition(@Nullable String name,
-                                  @Nullable Operation operation,
+                                  Operation operation,
                                   List<VariableDefinition> variableDefinitions,
                                   List<Directive> directives,
                                   SelectionSet selectionSet,
@@ -93,7 +93,7 @@ public class OperationDefinition extends AbstractNode<OperationDefinition> imple
         return name;
     }
 
-    public @Nullable Operation getOperation() {
+    public Operation getOperation() {
         return operation;
     }
 
@@ -184,7 +184,7 @@ public class OperationDefinition extends AbstractNode<OperationDefinition> imple
         private SourceLocation sourceLocation;
         private ImmutableList<Comment> comments = emptyList();
         private String name;
-        private Operation operation;
+        private Operation operation = Operation.QUERY;
         private ImmutableList<VariableDefinition> variableDefinitions = emptyList();
         private ImmutableList<Directive> directives = emptyList();
         private SelectionSet selectionSet;
--- a/src/main/java/graphql/language/AstPrinter.java
+++ b/src/main/java/graphql/language/AstPrinter.java
@@ -365,11 +365,11 @@ public class AstPrinter {
             // Anonymous queries with no directives or variable definitions can use
             // the query short form.
             if (isEmpty(name) && isEmpty(node.getDirectives()) && isEmpty(node.getVariableDefinitions())
-                    && (node.getOperation() == null || node.getOperation() == OperationDefinition.Operation.QUERY)) {
+                    && node.getOperation() == OperationDefinition.Operation.QUERY) {
                 node(out, node.getSelectionSet());
             } else {
                 OperationDefinition.Operation op = node.getOperation();
-                out.append(op != null ? op.toString().toLowerCase() : "query");
+                out.append(op.toString().toLowerCase());
                 if (!isEmpty(name)) {
                     out.append(' ');
                     out.append(name);
--- a/src/main/java/graphql/language/AstSorter.java
+++ b/src/main/java/graphql/language/AstSorter.java
@@ -281,7 +281,7 @@ public class AstSorter {
         Function<Definition, Integer> byType = d -> {
             if (d instanceof OperationDefinition) {
                 OperationDefinition.Operation operation = ((OperationDefinition) d).getOperation();
-                if (OperationDefinition.Operation.QUERY == operation || operation == null) {
+                if (OperationDefinition.Operation.QUERY == operation) {
                     return 101;
                 }
                 if (OperationDefinition.Operation.MUTATION == operation) {


dondonz:
This is a redo of a previous PR that went off the rails

.claude/commands/jspecify-annotate.md
```diff
@@ -12,6 +12,13 @@ Analyze this Java class and add JSpecify annotations based on:
 5. Method implementations that return null or check for null
 6. GraphQL specification details (see details below)
 
+## API Compatibility and Breaking Changes
+
+When adding JSpecify annotations, **DO NOT break existing public APIs**.
+- **Do not remove interfaces** from public classes (e.g., if a class implements `NamedNode`, it must continue to do so).
+- **Be extremely careful when changing methods to return `@Nullable`**. If an interface contract (or widespread ecosystem usage) expects a non-null return value, changing it to `@Nullable` is a breaking change that will cause compilation errors or `NullPointerException`s for callers. For example, if a method returned `null` implicitly but its interface requires non-null, you must honor the non-null contract (e.g., returning an empty string or default value instead of `null`).
+- **Do not change the binary signature** of methods or constructors in a way that breaks backwards compatibility.
+
```
dondonz, line 21: Previous attempt randomly deleted a constructor parameter which caused that PR to go off the rails. Let's fix that here

src/main/java/graphql/language/NamedNode.java
```diff
@@ -12,7 +13,7 @@
 public interface NamedNode<T extends NamedNode> extends Node<T> {
 
     /**
-     * @return the name of this node
+     * @return the name of this node, or null if this node is anonymous (e.g. an anonymous operation definition)
      */
-    String getName();
+    @Nullable String getName();
```
dondonz, line 18: This has to be nullable because an operation definition might not have a name

src/main/java/graphql/language/OperationDefinition.java
```diff
@@ -57,15 +61,6 @@ protected OperationDefinition(String name,
         this.selectionSet = selectionSet;
     }
 
-    public OperationDefinition(String name,
```
dondonz, line 60: These constructors date back to 2018, no longer needed with the builders we have today.

I am removing these lines because these parameters should be not-nullable. 

I would usually first deprecate the methods, then remove, but if I keep them here with a deprecated annotation, the nullability annotations will not be correct.

dondonz: This is technically a breaking change

src/main/java/graphql/analysis/QueryTraverser.java
```diff
@@ -166,7 +166,10 @@ public void visitField(QueryVisitorFieldEnvironment env) {
     }
 
     private GraphQLObjectType getRootTypeFromOperation(OperationDefinition operationDefinition) {
-        switch (operationDefinition.getOperation()) {
+        OperationDefinition.Operation op = operationDefinition.getOperation() != null
+                ? operationDefinition.getOperation()
+                : OperationDefinition.Operation.QUERY;
+        switch (op) {
```
dondonz: Some follow on changes after adding nullability annotations

src/test/groovy/graphql/schema/DataFetchingEnvironmentImplTest.groovy
```diff
@@ -29,7 +30,10 @@ class DataFetchingEnvironmentImplTest extends Specification {
     def frag = FragmentDefinition.newFragmentDefinition().name("frag").typeCondition(new TypeName("t")).build()
 
     def dataLoader = DataLoaderFactory.newDataLoader({ keys -> CompletableFuture.completedFuture(keys) } as BatchLoader)
-    def operationDefinition = new OperationDefinition("q")
+    def operationDefinition = OperationDefinition.newOperationDefinition()
+            .name("q")
+            .selectionSet(SelectionSet.newSelectionSet().selection(new Field("f")).build())
+            .build()
```
dondonz, line 36: Minor test change after selection set not nullable

src/main/java/graphql/language/ObjectField.java
```diff
@@ -88,7 +92,7 @@ public boolean isEqualTo(Node o) {
 
     @Override
     public ObjectField deepCopy() {
-        return new ObjectField(name, deepCopy(this.value), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
+        return new ObjectField(name, assertNotNull(deepCopy(this.value), "value deepCopy should not return null"), getSourceLocation(), getComments(), getIgnoredChars(), getAdditionalData());
```
bbakerman, line 95: I used to think assertion messages matter but they dont.  A stacktrace would lead to this code location - so less code can be better code

src/main/java/graphql/language/OperationDefinition.java
```diff
@@ -57,15 +61,6 @@ protected OperationDefinition(String name,
...
-                               Operation operation) {
-        this(name, operation, emptyList(), emptyList(), null, null, emptyList(), IgnoredChars.EMPTY, emptyMap());
-    }
-
-    public OperationDefinition(String name) {
-        this(name, null, emptyList(), emptyList(), null, null, emptyList(), IgnoredChars.EMPTY, emptyMap());
-    }
-
```
bbakerman, line 68: Fair enough - cruft must be removed at some point

src/main/java/graphql/language/OperationDefinition.java
```diff
@@ -93,11 +88,12 @@ public OperationDefinition withNewChildren(NodeChildrenContainer newChildren) {
...
     }
 
-    public String getName() {
+    @Override
+    public @Nullable String getName() {
         return name;
     }
 
-    public Operation getOperation() {
+    public @Nullable Operation getOperation() {
```
bbakerman: Ummmmmm can this be null???  if its not specified surely is defaults to QUERY??

src/main/java/graphql/language/ObjectField.java
dondonz: This assert is a special one to satisfy the NullAway check

The problem is that deepCopy is a generic utility and I can't guarantee that always returns not-nullable result

src/main/java/graphql/language/OperationDefinition.java
dondonz: Yep I agree, should have defaulted to query instead

dondonz: It does in the parser as well

PR #4258, opened 2026-02-22
Add agentic CI doctor workflow

dondonz:
Adds the [ci-doctor](https://github.com/githubnext/agentics/blob/main/docs/ci-doctor.md) agentic workflow, which automatically investigates CI failures. This is one of GitHub's own agentic pipelines from their documentation.

When a monitored workflow fails, ci-doctor analyses failing builds and either comments on the PR or opens an issue (in the unlikely case of a master build failure)

It's not easy to test this as a dry-run. The agentic pipeline will only show up in the list of pipelines after merging into master. My plan is to give it a test on a deliberately failing pull request after this is merged into master.


dondonz:
Keeping this in draft for now, I want to give this a test, even though the downside risk is minimal

.gitignore
```diff
@@ -1,3 +1,4 @@
+.claude/worktrees/
```
dondonz, line 1: And adding a slightly off topic change: I've found Claude & git worktrees very handy for multiple PRs at the same time

dondonz:
An idea: we don't read our performance pipeline output often, and until now I was manually reading a JSON diff whenever we had a "big" change worth measuring!

What if we had an agentic pipeline constantly monitoring for performance regressions?

We can also have a pipeline to explain errors, although everyone can click through the build output, it's handy to have the right paragraph already extracted.

.github/workflows/ci-doctor.md
```diff
@@ -0,0 +1,196 @@
...
+  patterns, and provide actionable remediation steps. Analyzes logs, error messages,
+  and workflow configuration to help diagnose and resolve CI issues efficiently.
+
+on:
+  workflow_run:
+    workflows: ["Daily Perf Improver", "Daily Test Coverage Improver"]  # Monitor the CI workflow specifically
+    types:
+      - completed
+    branches:
+      - main
```
dondonz: Should be master

dondonz:
Awkwardly I have to test this one after merging. If the error messages are low quality I'll bin it

PR #4271, opened 2026-02-25
Validate constant directive values on variable definitions (#3927)

andimarek:
Per the GraphQL spec, directives applied to variable definitions accept
only constant values. Variable references like $v are not allowed in
directive arguments on variable definitions. For example, the query
`query ($v:Int @dir(arg:$v)) { ... }` should be rejected.

Add a new validation rule VARIABLES_NOT_ALLOWED_IN_DIRECTIVES_ON_VARIABLE_DEFINITIONS
that detects variable references inside directive arguments on variable
definitions and reports a VariableNotAllowed error.

https://claude.ai/code/session_01QH4Ce5Fs9zfEjktZV4j1i2


bbakerman (approved):
Thanks Claude

PR #4272, opened 2026-02-27
Add JSpecify annotations to execution instrumentation classes (10 more)

dondonz:
## Summary

Annotates execution instrumentation public API classes with `@NullMarked` and appropriate `@Nullable` annotations, removing them from the JSpecify archunit exemption list.

**Classes annotated:**

- **`DeferredExecution`** — `label` is `@Nullable` per the GraphQL spec (`@defer` label is optional)
- **`ResponseMapFactory`** — map values are `@Nullable Object` since GraphQL fields can resolve to null
- **`ChainedInstrumentation`** — removes redundant `@NonNull` on `instrument*` overrides (covered by `@NullMarked`); propagates `@Nullable` from the `Instrumentation` interface onto all `begin*` return types and `onCompleted` parameters
- **`DocumentAndVariables`** — `@NullUnmarked` on `Builder` (uninitialized fields)
- **`NoContextChainedInstrumentation`** — `runAll()` return type annotated `@Nullable T` (always returns null)
- **`SimpleInstrumentation`** — `@NullMarked` only; all methods delegate to the interface defaults
- **`SimpleInstrumentationContext`** — `codeToRunOnComplete` and `codeToRunOnDispatch` fields are `@Nullable` (set to null by the no-arg constructor); `nonNullCtx` parameter and `onCompleted` params annotated `@Nullable`
- **`SimplePerformantInstrumentation`** — removes redundant `@NonNull` on `instrument*` overrides
- **`FieldAndArguments`** — `getParentFieldAndArguments()` is `@Nullable` (root field has no parent); `getArgumentValue()` returns `@Nullable T` (argument values can be absent/null)
- **`FieldValidationEnvironment`** — `@NullMarked` only; all return types are non-null

**Also fixes** a pre-existing gap in `Instrumentation`: `beginDeferredField` was the only `begin*` method missing `@Nullable` on its return type, despite the Javadoc stating it returns "a nullable `InstrumentationContext`". This would have caused a nullability conflict once `Instrumentation` itself becomes `@NullMarked`.

## Test plan

- [x] Archunit `JSpecifyAnnotationsCheck` passes — the 10 exemptions for these classes are removed
- [x] All existing tests pass

### DIFF
--- a/src/main/java/graphql/execution/instrumentation/ChainedInstrumentation.java
+++ b/src/main/java/graphql/execution/instrumentation/ChainedInstrumentation.java
@@ -67,7 +67,7 @@ public class ChainedInstrumentation implements Instrumentation {
         return instrumentations;
     }
 
-    private <T> @Nullable InstrumentationContext<T> chainedCtx(InstrumentationState state, BiFunction<Instrumentation, InstrumentationState, @Nullable InstrumentationContext<T>> mapper) {
+    private <T> @Nullable InstrumentationContext<T> chainedCtx(InstrumentationState state, BiFunction<Instrumentation, InstrumentationState, InstrumentationContext<T>> mapper) {
         // if we have zero or 1 instrumentations (and 1 is the most common), then we can avoid an object allocation
         // of the ChainedInstrumentationContext since it won't be needed
         if (instrumentations.isEmpty()) {


src/main/java/graphql/execution/instrumentation/ChainedInstrumentation.java
```diff
@@ -66,7 +67,7 @@ public List<Instrumentation> getInstrumentations() {
         return instrumentations;
     }
 
-    private <T> InstrumentationContext<T> chainedCtx(InstrumentationState state, BiFunction<Instrumentation, InstrumentationState, InstrumentationContext<T>> mapper) {
+    private <T> @Nullable InstrumentationContext<T> chainedCtx(InstrumentationState state, BiFunction<Instrumentation, InstrumentationState, InstrumentationContext<T>> mapper) {
```
dondonz, line 70: This project is a good exercise in catching a bunch of extra nullable annotations

Issue #4185, opened 2025-12-04
FieldVisibilitySchemaTransformation doesn't take union members into account


dondonz:
Found an interesting bug today

The FieldVisibilitySchemaTransformation eagerly deletes types that can no longer be accessed after a field is removed. This doesn't check for references from unions though, so it's possible to end up with an illegal schema where a union type has a union member that doesn't exist anymore.

I'll put up a proper test case to explain this better

andimarek:
@dondonz I believe that is fixed now ... but lets double check.

andimarek:
I am very sure this is fixed ... therefore closing

dondonz:
My bad I missed your comment earlier

PR #4275, opened 2026-03-01
Check in field visibility transformer test - union member error now fixed

dondonz:
With the large refactor of the field visibility transformer, this issue is now also fixed https://github.com/graphql-java/graphql-java/issues/4185

For completeness I'm checking in a test with the original bug reported at work, with anonymised schema element names


dondonz:
Checked out against a copy of the field visibility transformer prior to the fixes (in April 2025) and confirmed the test was failing on the old version.

PR #4276, opened 2026-03-02
Re-add GraphQLSchema.Builder.clearDirectives()

### LINKED ISSUE #4259: Did PR #4229 go too far?
PR #4229 unifies the behavior of all built-in directives, where in the past skip and include was treated differently from other built ins.  In the process, the `GraphQLSchema.Builder.clearDirectives` method was removed, because this was one of the places where different built-ins were being treated differently.

However, we have a use-case where we use `GraphQLSchema.transform` to transform all of the non-built-in directives in a schema.  This works by taking all the existing non-graphql-builtin directives from the source schema, then calling `clearDirectives` on the builder for the new schema, then adding transformations of the original directives into the new schema.  So a semantics of `clearDirectives` which is "clear all non-graphql-builtin directives" would be helpful.  So perhaps the change in #4229 should've been: "`clearDirectives` clears non-graphql built-in directives, but treats all graphql built-ins uniformly (contra the old behavior)" -- rather than getting rid of it altogether.

rstata:
## Summary

PR #4229 removed `clearDirectives()` from `GraphQLSchema.Builder` as part of unifying built-in directive handling. As noted in issue #4259, this method is useful in practice: a common pattern when using `GraphQLSchema.transform` is to collect the non-built-in directives from the source schema, call `clearDirectives()` on the builder, and then add back transformed versions of those directives.

This PR re-adds `clearDirectives()` with semantics that are consistent with the approach introduced in #4229:

- Clears `additionalDirectives` entirely (including any explicitly-added built-ins)
- Built-in directives are always re-added automatically at build time by `ensureBuiltInDirectives()`, so they cannot be permanently removed — all built-ins are treated uniformly

## Changes

- Re-adds `Builder.clearDirectives()` to `GraphQLSchema` with Javadoc illustrating the transform use case
- Adds tests covering:
  - `clearDirectives()` followed by `build()` still yields all 7 built-in directives
  - `clearDirectives()` followed by adding a custom directive yields the expected set
  - The primary use case: using `clearDirectives()` in a `schema.transform()` to replace non-built-in directives
  - Ordering behavior: unoverridden built-ins sort first, then user-supplied directives in insertion order; customized built-ins retain their custom properties

## Related

Fixes #4259. See also #4229.

### DIFF
--- a/src/main/java/graphql/schema/GraphQLSchema.java
+++ b/src/main/java/graphql/schema/GraphQLSchema.java
@@ -931,6 +931,30 @@ public class GraphQLSchema {
             return this;
         }
 
+        /**
+         * Clears all directives from this builder, including any that were previously added
+         * via {@link #additionalDirective(GraphQLDirective)} or {@link #additionalDirectives(Set)}.
+         * Built-in directives ({@code @include}, {@code @skip}, {@code @deprecated}, etc.) will
+         * always be added back automatically at build time by {@code ensureBuiltInDirectives()}.
+         * <p>
+         * This is useful when transforming a schema to replace all non-built-in directives:
+         * <pre>{@code
+         * schema.transform(builder -> {
+         *     List<GraphQLDirective> nonBuiltIns = schema.getDirectives().stream()
+         *         .filter(d -> !Directives.isBuiltInDirective(d))
+         *         .collect(toList());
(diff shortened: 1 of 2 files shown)


andimarek:
Thanks

PR #4287, opened 2026-03-03
Fix findPubliclyAccessibleMethod to search interfaces for accessible methods

### LINKED ISSUE #4278: PropertyFetchingImpl findPubliclyAccessibleMethod algorithm cannot find methods on non-public classes implementing public interfaces
For example, I get

```
java.lang.reflect.InaccessibleObjectException: Unable to make public java.lang.Object java.util.TreeMap$Entry.getValue() accessible: module java.base does not "opens java.util" to unnamed module @7c137fd5
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
        at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:200)
        at java.base/java.lang.reflect.Method.setAccessible(Method.java:194)
        at graphql.schema.PropertyFetchingImpl.findViaSetAccessible(PropertyFetchingImpl.java:306)
        at graphql.schema.PropertyFetchingImpl.lambda$getPropertyValue$3(PropertyFetchingImpl.java:158)
        at graphql.schema.PropertyFetchingImpl.getPropertyViaGetterUsingPrefix(PropertyFetchingImpl.java:216)
        at graphql.schema.PropertyFetchingImpl.getPropertyViaGetterMethod(PropertyFetchingImpl.java:210)
        at graphql.schema.PropertyFetchingImpl.getPropertyValue(PropertyFetchingImpl.java:159)
        at graphql.schema.PropertyDataFetcherHelper.getPropertyValue(PropertyDataFetcherHelper.java:23)
        at graphql.schema.PropertyDataFetcher.getImpl(PropertyDataFetcher.java:131)
        at graphql.schema.PropertyDataFetcher.get(PropertyDataFetcher.java:112)
```

and it happens because here only super classes but no interface implementations are considered:

```java
 private Method findPubliclyAccessibleMethod(CacheKey cacheKey, Class<?> rootClass, String methodName, boolean dfeInUse, boolean allowStaticMethods) throws NoSuchMethodException {
        Class<?> currentClass = rootClass;
        while (currentClass != null) {
            if (Modifier.isPublic(currentClass.getModifiers())) {
                if (dfeInUse) {
                    //
                    // try a getter that takes singleArgumentType first (if we have one)
                    try {
                        Method method = currentClass.getMethod(methodName, singleArgumentType);
                        if (isSuitablePublicMethod(method, allowStaticMethods)) {
                            METHOD_CACHE.putIfAbsent(cacheKey, new CachedMethod(method));
                            return method;
                        }
                    } catch (NoSuchMethodException e) {
                        // ok try the next approach
                    }
                }
                Method method = currentClass.getMethod(methodName);
                if (isSuitablePublicMethod(method, allowStaticMethods)) {
                    METHOD_CACHE.putIfAbsent(cacheKey, new CachedMethod(method));
                    return method;
                }
            }
            currentClass = currentClass.getSuperclass();
        }
        assert rootClass != null;
        return rootClass.getMethod(methodName);
    }
```

and the utility tries to use `findViaSetAccessible` which actually is not necessary e.g. for TreeMap.Entry

```java
    static final class Entry<K,V> implements Map.Entry<K,V> {
        K key;
        V value;
        Entry<K,V> left;
        Entry<K,V> right;
        Entry<K,V> parent;
        boolean color = BLACK;

        /**
         * Make a new cell with given key, value, and parent, and with
         * {@code null} child links, and BLACK color.
         */
        Entry(K key, V value, Entry<K,V> parent) {
            this.key = key;
            this.value = value;
            this.parent = parent;
        }

        /**
         * Returns the key.
         *
         * @return the key
         */
        public K getKey() {
            return key;
        }

        /**
         * Returns the value associated with the key.
         *
         * @return the value associated with the key
         */
        public V getValue() {
            return value;
        }

        /**
         * Replaces the value currently associated with the key with the given
         * value.
         *
         * @return the value associated with the key before this method was
         *         called
         */
  ```

### PR DESCRIPTION
**Pull request by andimarek**
## Summary
- Fixes #4278: `PropertyDataFetcher` fails to fetch properties on non-public classes (e.g., `TreeMap.Entry`) that implement public interfaces (e.g., `Map.Entry`) on Java 16+ where `setAccessible` is restricted for JDK internal classes
- Adds a new `findMethodOnPublicInterfaces` method that recursively searches public interfaces implemented by a class when the class itself is not publicly accessible
- Adds tests verifying that `PropertyDataFetcher` can fetch `key` and `value` from `TreeMap.Entry` via the public `Map.Entry` interface

## Test plan
- [x] Added unit tests for fetching `value` and `key` properties from `TreeMap.Entry` instances
- [x] Tests cover both `PropertyDataFetcher` and `SingletonPropertyDataFetcher`
- [x] Tests explicitly disable lambda factory to exercise the reflection code path

https://claude.ai/code/session_01KSM9EgschAaxJVFeqM1enw

### DIFF
--- a/src/test/groovy/graphql/schema/PropertyDataFetcherTest.groovy
+++ b/src/test/groovy/graphql/schema/PropertyDataFetcherTest.groovy
@@ -872,6 +872,41 @@ class PropertyDataFetcherTest extends Specification {
         "baseValue" | "diamondBaseValue"
     }
 
+    def "fetch via public interface method with DataFetchingEnvironment parameter on non-public class"() {
+        given:
+        // PackagePrivateDfeImpl implements PublicDfeInterface which declares getDfeValue(DataFetchingEnvironment)
+        // This exercises the dfeInUse path in findMethodOnPublicInterfaces (lines 262-267)
+        PropertyDataFetcherHelper.setUseLambdaFactory(false)
+        PropertyDataFetcher.clearReflectionCache()
+
+        def obj = InterfaceInheritanceHolder.createDfeImpl()
+        def environment = env("dfeValue", obj)
+
+        when:
+        def result = new PropertyDataFetcher("dfeValue").get(environment)
+
+        then:
+        result == "dfeValue"
+    }
+
+    def "fetch via interface search hits NoSuchMethodException and continues to next interface"() {
+        given:
+        // PackagePrivateMultiInterfaceImpl implements PublicInterfaceWithoutTarget (no getBaseValue)
+        // and PublicBaseInterface (has getBaseValue). The search must hit NoSuchMethodException
+        // on the first interface and continue to find it on the second.
+        PropertyDataFetcherHelper.setUseLambdaFactory(false)
+        PropertyDataFetcher.clearReflectionCache()
+
+        def obj = InterfaceInheritanceHolder.createMultiInterfaceImpl()
+        def environment = env("baseValue", obj)
+
+        when:
+        def result = new PropertyDataFetcher("baseValue").get(environment)
+
+        then:
+        result == "foundViaSecondInterface"
+    }
+
     def "Can access private property from base class that starts with i in Turkish"() {
         // see https://github.com/graphql-java/graphql-java/issues/3385
         given:
--- a/src/test/groovy/graphql/schema/somepackage/InterfaceInheritanceHolder.java
+++ b/src/test/groovy/graphql/schema/somepackage/InterfaceInheritanceHolder.java
@@ -1,5 +1,7 @@
 package graphql.schema.somepackage;
 
+import graphql.schema.DataFetchingEnvironment;
+
 /**
  * Test fixtures for interface-extends-interface method resolution.
  * <p>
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/main/java/graphql/schema/PropertyFetchingImpl.java [hunk 1]
```diff
@@ -244,12 +244,51 @@ private Method findPubliclyAccessibleMethod(CacheKey cacheKey, Class<?> rootClas
...
+                    if (isSuitablePublicMethod(method, allowStaticMethods)) {
+                        METHOD_CACHE.putIfAbsent(cacheKey, new CachedMethod(method));
+                        return method;
+                    }
+                } catch (NoSuchMethodException e) {
+                    // continue searching
+                }
+            }
+            // Also search super-interfaces of non-public interfaces
+            Method method = findMethodOnPublicInterfaces(cacheKey, iface.getInterfaces(), methodName, dfeInUse, allowStaticMethods);
```
bbakerman, line 284: You allowing interfaces implementing interfaces - fair enough - but there is not tests for that

I think you need to create the complicated structure that tests this

PR #4288, opened 2026-03-03
Fix ShallowTypeRefCollector: resolve type refs in applied directive arguments

rstata:
## Summary

`ShallowTypeRefCollector` (introduced with `FastBuilder` in #4196) performs
shallow scans of types and directives to collect `GraphQLTypeReference` objects
for later replacement. In three places it called `scanArgumentType(arg)` on a
`GraphQLArgument` but never called `scanAppliedDirectives(arg.getAppliedDirectives())`.
It also had no handling to descend into `GraphQLEnumValueDefinition` objects to
scan their applied directives.

The result: when `FastBuilder` is used with programmatically-constructed types
carrying `GraphQLTypeReference` objects in applied directive argument types on
field arguments, enum values, or directive definition arguments, those references
are never resolved. The `AppliedDirectiveArgumentsAreValid` validator then sees a
`GraphQLTypeReference` where it expects a real input type and throws:

    InvalidSchemaException: Invalid argument 'x' for applied directive of name 'y'

(This does not manifest when types are built via `FastSchemaGenerator` + SDL,
because `SchemaGeneratorHelper.buildInputType` resolves types at construction
time. It manifests when `FastBuilder` is used directly with programmatically-built
types.)

## Changes

Three locations in `ShallowTypeRefCollector` fixed:

- **`handleObjectType` / `handleInterfaceType`** — added
  `scanAppliedDirectives(arg.getAppliedDirectives())` inside the field-argument
  loop of both methods, alongside the existing `scanArgumentType(arg)` call.

- **`handleTypeDef`** — added `handleEnumType()` which iterates over
  `GraphQLEnumValueDefinition` values and scans their applied directives. (Applied
  directives on the enum type itself were already covered by the existing
  `GraphQLDirectiveContainer` branch.)

- **`handleDirective`** — added `scanAppliedDirectives(argument.getAppliedDirectives())`
  inside the directive-argument loop alongside the existing `scanArgumentType` call.

Adds four regression tests in `FastBuilderTest` covering all three fix locations:
applied directive on object type field argument, interface field argument, enum
value, and directive definition argument.

## Related

Bug in the `FastBuilder` / `ShallowTypeRefCollector` introduced in #4196.


rstata:
(more internal testing has failed -- stand by...)

andimarek:
@rstata is that ready or are we waiting for more?

rstata:
oops -- i forgot about this!  turns out this version did fix all the problems, viaduct builds and passes all its tests with this fix in place.

PR #4305, opened 2026-03-07
Update agents.md to improve AI assisted PRs

dondonz:
Tries to correct common issues in PRs, this will be read by Claude before starting each session, so it's important this is kept very short to reduce impact on context window

For example: I noticed that AI tools forget we use Spock/Groovy tests

https://claude.ai/code/session_01LQddmhPjeEKJMGyx8ZZVcw


.gitignore
```diff
@@ -17,4 +16,5 @@ docs/_build/
 /.nb-gradle/
 gen
 .DS_Store
-.vscode
\ No newline at end of file
+.vscode
+.claude/scheduled_tasks.lock
```
dondonz, line 20: New `/loop` feature runs cronjobs

PR #4338, opened 2026-03-11
Fix NullAway 0.13.1 compatibility

andimarek:
## Summary
- Fix 3 NullAway errors that occur with NullAway 0.13.1 (blocking PR #4317)
- `ArrayValue.deepCopy()` / `ObjectValue.deepCopy()`: inline `assertNotNull(deepCopy(...))` to avoid `@Nullable` local variable being passed where `@NonNull` is required
- `MergedField.Builder.build()`: replace `assertNotNull` on `@Nullable` Builder field with explicit null check

## Test plan
- [x] Verified compilation passes with NullAway 0.12.10
- [x] Verified compilation passes with NullAway 0.13.1

### DIFF
--- a/src/main/java/graphql/execution/MergedField.java
+++ b/src/main/java/graphql/execution/MergedField.java
@@ -16,7 +16,7 @@ import java.util.Objects;
 import java.util.function.Consumer;
 
 import static graphql.Assert.assertNotEmpty;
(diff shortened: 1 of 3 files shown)


andimarek:
No longer needed — staying on NullAway 0.12.10 for now. See #4339.

PR #4341, opened 2026-03-17
Fix coverage-modified classes being included in published artifacts

dondonz:
The markGeneratedEqualsHashCode task was modifying .class files in-place
in build/classes/java/main and the jar task depended on it, causing
bytecode with injected @Generated annotations to end up in the published
JAR. Fix by copying classes to a separate build/classes-jacoco directory
for modification, and only wiring jacocoTestReport (not jar) to depend
on the task.

https://claude.ai/code/session_01XDZqTUKLBoSJpGPGx9Amxg


dondonz:
I have to look more closely at this - so far only a quick Claude attempt

dondonz:
Fixed by #4343

PR #4343, opened 2026-03-18
Never package bytecode-modified class files in published JARs

andimarek:
## Summary

- The `markGeneratedEqualsHashCode` task was modifying `.class` files **in-place** in `build/classes/java/main` and the `jar` task depended on it, causing every published class with `equals`/`hashCode` to contain a spurious `@Generated` (`Lgraphql/coverage/Generated;`) annotation in the bytecode
- Fix: copy classes to a separate `build/classes-jacoco/` directory before injecting annotations, prepend that directory to the test classpath (so Jacoco CRC64 checksums match the execution data), and point `jacocoTestReport` at the modified copy
- The `jar`/`shadowJar`/publish tasks now only see pristine compiler output — no bytecode modifications leak into artifacts

## Test plan

- [x] Verified original classes in `build/classes/java/main` remain unmodified after running `markGeneratedEqualsHashCode`
- [x] Verified classes in `build/classes-jacoco/java/main` contain the `@Generated` annotation
- [x] Verified the built JAR does **not** contain `graphql.coverage.Generated` annotations
- [x] Verified the shadow JAR does **not** contain `graphql.coverage.Generated` annotations
- [x] All tests pass (`./gradlew test`)
- [x] Jacoco report generates successfully with real coverage data (non-zero counters)
- [x] `equals`/`hashCode` methods are properly excluded from the Jacoco coverage report

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -574,15 +574,17 @@ jacocoTestReport {
         csv.required = false
     }
 
-    // Exclude generated ANTLR code from coverage
+    // Use the modified classes from classes-jacoco/ (with @Generated annotations)
+    // so CRC64 checksums match the execution data recorded during tests.
+    // Also exclude generated ANTLR code and shaded dependencies from coverage.
     afterEvaluate {
-        classDirectories.setFrom(files(classDirectories.files.collect {
-            fileTree(dir: it, exclude: [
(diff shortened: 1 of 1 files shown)


dondonz:
Thanks!!!

PR #4375, opened 2026-05-08
Avoid no-op field fetch adapter allocation

nwjsmith:
## Problem

`beginFieldFetching` keeps older `beginFieldFetch` implementations working by
calling the deprecated method and adapting the returned `InstrumentationContext`.
For instrumentations that inherit the default/no-op `beginFieldFetch`, that means
adapting `SimpleInstrumentationContext.noOp()` into a new anonymous
`FieldFetchingInstrumentationContext`.

That adapter allocation happens once per field fetch. In a `ChainedInstrumentation`
with multiple legacy/no-op field-fetch instrumentations, each instrumentation can
pay that allocation for every field even though all callbacks are no-ops.

## Fix

When the legacy hook returns the canonical `SimpleInstrumentationContext.noOp()`
instance, return the canonical `FieldFetchingInstrumentationContext.NOOP` directly
instead of allocating an adapter.

Real legacy `beginFieldFetch` overrides are still adapted normally, and
instrumentations that override `beginFieldFetching` themselves are unaffected.

## Tests

Added coverage for:

- default `Instrumentation` no-op field fetching
- `SimplePerformantInstrumentation` inherited no-op field fetching
- deprecated overrides that explicitly return `noOp()`
- deprecated overrides that return a real context and still need adaptation

Validated with:

```bash
./gradlew test --tests graphql.execution.instrumentation.InstrumentationDefaultMethodsTest --no-build-cache
```

### DIFF
--- a/src/main/java/graphql/execution/instrumentation/Instrumentation.java
+++ b/src/main/java/graphql/execution/instrumentation/Instrumentation.java
@@ -241,6 +241,9 @@ public interface Instrumentation {
     @Nullable
     default FieldFetchingInstrumentationContext beginFieldFetching(InstrumentationFieldFetchParameters parameters, InstrumentationState state) {
         InstrumentationContext<Object> ctx = beginFieldFetch(parameters, state);
+        if (ctx == noOp()) {
+            return FieldFetchingInstrumentationContext.NOOP;
+        }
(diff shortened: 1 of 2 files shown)


andimarek:
This is now fixed as part as a bigger PR #4391  .. thanks for reporting @nwjsmith

PR #4378, opened 2026-05-11
Add `onError` request parameter

martinbonnin:
See https://github.com/graphql/graphql-spec/pull/1236


bbakerman:
One possible approach might be to "maintain" the old directive WHILE adding the new ExecutionInput based approach.

That way anyone who has used the old will be able to migrate to the new without breaking.  This will cost time and complexity of course but it will be "less impactful" to current users

That said - at what point is the kill date?  One could argue it was "experimental" and hence the kill date is immediate as this PR has done

martinbonnin:
@bbakerman agreed, I'll split this PR in 2 so we can have both at the same time.

Edit: added back `@experimental_disableErrorPropagation` in https://github.com/graphql-java/graphql-java/pull/4378/commits/e91ba024ccef6c71dd6b9f8902c13d523e2fabd8, will do the PR to remove `@experimental_disableErrorPropagation` when needed

martinbonnin:
This is still pretty much up to date. I think we should merge it.

PR #4386, opened 2026-05-19
Fix input cycle validation for non-null lists

### LINKED ISSUE #4214: Schema generator reports false positive unsatisfiable input object cycle
[Per the spec](https://spec.graphql.org/draft/#example-2f832), cyclic input object fields of non-nullable list type with non-nullable items are valid since the field value can be an empty list. However, the `NoUnbrokenInputCycles` rule reports these input objects as invalid.

To reproduce, run the following code:
```java
import graphql.schema.idl.NoopWiringFactory;
import graphql.schema.idl.RuntimeWiring;
import graphql.schema.idl.SchemaGenerator;
import graphql.schema.idl.SchemaParser;

public class Main {
        String query = """
                  input Example {
                    self: [Example!]!
                    value: String
                  }
                
                  type Query {
                    example(example: Example): String
                  }
                """.stripIndent();
        TypeDefinitionRegistry typeDefinitionRegistry = new SchemaParser().parse(query);
        RuntimeWiring runtimeWiring = RuntimeWiring.newRuntimeWiring().wiringFactory(new NoopWiringFactory()).build();
        new SchemaGenerator().makeExecutableSchema(typeDefinitionRegistry, runtimeWiring);
    }
}
```
Observe an `InvalidSchemaException` is thrown with the message `[self!] forms an unsatisfiable cycle`

andimarek:
## Summary
- treat non-null list input fields as breaking unbroken input cycles
- add SDL regression coverage for #4214

## Verification
- Confirmed the new regression test fails on current master with: `[self!] forms an unsatisfiable cycle`
- ./gradlew test --tests graphql.schema.validation.NoUnbrokenInputCyclesTest
- ./gradlew test --tests 'graphql.schema.validation.*'

Fixes #4214

### DIFF
--- a/src/main/java/graphql/schema/validation/NoUnbrokenInputCycles.java
+++ b/src/main/java/graphql/schema/validation/NoUnbrokenInputCycles.java
@@ -6,7 +6,6 @@ import graphql.schema.GraphQLFieldDefinition;
 import graphql.schema.GraphQLInputObjectField;
 import graphql.schema.GraphQLInputObjectType;
 import graphql.schema.GraphQLInputType;
-import graphql.schema.GraphQLList;
 import graphql.schema.GraphQLNonNull;
 import graphql.schema.GraphQLSchemaElement;
 import graphql.schema.GraphQLType;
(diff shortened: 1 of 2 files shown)


andimarek:
Double-checked against the current GraphQL draft spec. Example 83 is exactly the case covered by this PR:

`input Example { self: [Example!]! value: String }`

The spec says this circular reference is valid because `self` may be an empty list, and the Input Object type validation rule says a recursive chain is valid when at least one field in the chain is nullable or a List type.

Spec reference: https://spec.graphql.org/draft/#sec-Input-Objects.Circular-References

PR #4391, opened 2026-05-20
Avoid no-op field fetch instrumentation allocations

andimarek:
## Summary
- include the original #4375 fix to return the canonical field-fetching no-op when legacy field-fetch instrumentation returns `SimpleInstrumentationContext.noOp()`
- avoid chained field-fetch context/list allocation when all chained instrumentations are canonical no-ops
- return the sole real field-fetch context directly when no-op entries surround it
- preserve chained callback behavior when multiple real field-fetch contexts are present

## Verification
- ./gradlew test --tests graphql.execution.instrumentation.InstrumentationDefaultMethodsTest --no-build-cache
- ./gradlew test --tests 'graphql.execution.instrumentation.*'


bbakerman (approved):
Performance matters but gets harder ands harder to find!

nwjsmith:
Thank you!

PR #4398, opened 2026-05-24
Capture partial results on cancellation

AlexandreCarlton:
Web clients can often impose request timeouts. If a GraphQL server takes too long to respond, then nothing is returned. While we can call `ExecutionInput.cancel()` in order to slide under a web client's timeout, there are no usable results for the client to handle.

This change improves upon this - if we set `#capturePartialResultsOnCancel`, then calling `.cancel` will save all the evaluated `DataFetcher` results so that something potentially usable can be returned to the client.

Full disclosure: almost all of this was generated with [Rovo](https://www.atlassian.com/software/rovo), with a preliminary review from @bbakerman before submitting here.

### DIFF
--- a/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
+++ b/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
@@ -2,7 +2,10 @@ package graphql.execution;
 
 import graphql.ExecutionResult;
 import graphql.ExecutionResultImpl;
+import graphql.GraphQLContext;
 import graphql.PublicSpi;
+
+import java.util.concurrent.CompletionException;
 
 import java.util.List;
 import java.util.Map;
@@ -29,8 +32,47 @@ public abstract class AbstractAsyncExecutionStrategy extends ExecutionStrategy {
                 return;
             }
 
-            Map<String, Object> resolvedValuesByField = executionContext.getResponseMapFactory().createInsertionOrdered(fieldNames, results);
-            overallResult.complete(new ExecutionResultImpl(resolvedValuesByField, executionContext.getErrors()));
+            completeResultFuture(overallResult, executionContext, fieldNames, results);
         };
+    }
+
+    protected BiConsumer<List<Object>, Throwable> handleResultsWithPartialData(ExecutionContext executionContext, List<String> fieldNames, CompletableFuture<ExecutionResult> overallResult) {
+        return (List<Object> results, Throwable exception) -> {
+            // when partial results on cancel is enabled the results list will already have partial data
+            // (already-completed fields) so we can build a partial response even if exception is set
+            if (exception != null) {
+                Throwable cause = exception instanceof CompletionException ? exception.getCause() : exception;
+                if (cause instanceof AbortExecutionException && results != null
+                        && capturePartialResults(executionContext)) {
+                    executionContext.addError((AbortExecutionException) cause);
+                    completeResultFuture(overallResult, executionContext, fieldNames, results);
+                    return;
+                }
+                handleNonNullException(executionContext, overallResult, exception);
+                return;
+            }
+
+            // check if cancel fired while results were being gathered (no exception, but cancelled)
+            Throwable cancelException = executionContext.possibleCancellation(null);
+            if (cancelException != null) {
+                Throwable cancelCause = cancelException instanceof CompletionException ? cancelException.getCause() : cancelException;
+                if (cancelCause instanceof AbortExecutionException && results != null
+                        && capturePartialResults(executionContext)) {
+                    // we have partial data from already-completed CFs — use it
+                    executionContext.addError((AbortExecutionException) cancelCause);
+                    completeResultFuture(overallResult, executionContext, fieldNames, results);
+                } else {
+                    handleNonNullException(executionContext, overallResult, cancelException);
+                }
+                return;
+            }
+
+            completeResultFuture(overallResult, executionContext, fieldNames, results);
+        };
+    }
+
+    protected void completeResultFuture(CompletableFuture<ExecutionResult> overallResult, ExecutionContext executionContext, List<String> fieldNames, List<Object> results) {
+        Map<String, Object> resolvedValuesByField = executionContext.getResponseMapFactory().createInsertionOrdered(fieldNames, results);
+        overallResult.complete(new ExecutionResultImpl(resolvedValuesByField, executionContext.getErrors()));
     }
 }
--- a/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
+++ b/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
@@ -4,8 +4,6 @@ import graphql.ExecutionResult;
 import graphql.ExecutionResultImpl;
 import graphql.PublicSpi;
 
-import java.util.concurrent.CompletionException;
-
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
@@ -37,28 +35,18 @@ public abstract class AbstractAsyncExecutionStrategy extends ExecutionStrategy {
 
     protected BiConsumer<List<Object>, Throwable> handleResultsWithPartialData(ExecutionContext executionContext, List<String> fieldNames, CompletableFuture<ExecutionResult> overallResult) {
         return (List<Object> results, Throwable exception) -> {
-            // when partial results on cancel is enabled the results list will already have partial data
-            // (already-completed fields) so we can build a partial response even if exception is set
             if (exception != null) {
-                Throwable cause = exception instanceof CompletionException ? exception.getCause() : exception;
-                if (cause instanceof AbortExecutionException && results != null
-                        && capturePartialResults(executionContext)) {
-                    executionContext.addError((AbortExecutionException) cause);
-                    completeResultFuture(overallResult, executionContext, fieldNames, results);
-                    return;
-                }
                 handleNonNullException(executionContext, overallResult, exception);
                 return;
             }
 
-            // check if cancel fired while results were being gathered (no exception, but cancelled)
+            // No exception, but cancellation may have fired while the already-completed field values
+            // were being gathered. When it has, the results list already holds the partial data from
+            // the fields that completed before cancellation, so we can return it alongside the error.
             Throwable cancelException = executionContext.possibleCancellation(null);
             if (cancelException != null) {
-                Throwable cancelCause = cancelException instanceof CompletionException ? cancelException.getCause() : cancelException;
-                if (cancelCause instanceof AbortExecutionException && results != null
-                        && capturePartialResults(executionContext)) {
-                    // we have partial data from already-completed CFs — use it
-                    executionContext.addError((AbortExecutionException) cancelCause);
+                if (capturePartialResults(executionContext)) {
+                    executionContext.addError((AbortExecutionException) cancelException);
                     completeResultFuture(overallResult, executionContext, fieldNames, results);
                 } else {
                     handleNonNullException(executionContext, overallResult, cancelException);
--- a/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
+++ b/src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
@@ -2,7 +2,6 @@ package graphql.execution;
 
 import graphql.ExecutionResult;
 import graphql.ExecutionResultImpl;
-import graphql.GraphQLContext;
 import graphql.PublicSpi;
 
 import java.util.concurrent.CompletionException;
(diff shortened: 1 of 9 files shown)


src/main/java/graphql/execution/Async.java
```diff
@@ -232,6 +257,101 @@ public CompletableFuture<List<T>> await() {
...
+        public CompletableFuture<List<T>> await(GraphQLContext graphQLContext) {
+            commonSizeAssert();
+            if (cfCount == 0) {
+                return CompletableFuture.completedFuture(materialisedList(array));
+            }
+            if (!graphQLContext.getBoolean(CAPTURE_PARTIAL_RESULTS_ON_CANCEL)) {
+                return await();
+            }
+
+            CompletableFuture<Void> cancellationFuture = graphQLContext.get(CANCELLATION_FUTURE_KEY);
```
bbakerman: I think this should be made more generic - take the KNOWLEDGE of CAPTURE_PARTIAL_RESULTS_ON_CANCEL out of the code

Pass in say `Supplier<CompletableFuture<Void>>` into the method and have the context checks happen for that inside the caller code.

of even just pass in a `CompletableFuture<Void> cancelCF` and not a supplier.  Again have the client code do the preconditions on whether it passes in that `cancelCF` or not

Async is very. much a "dumb" util class that in theory could be re-used in many ways - so take the specific knowledge  out of here

src/main/java/graphql/execution/AsyncExecutionStrategy.java
```diff
@@ -64,14 +68,36 @@ public CompletableFuture<ExecutionResult> execute(ExecutionContext executionCont
         CompletableFuture<ExecutionResult> overallResult = new CompletableFuture<>();
         executionStrategyCtx.onDispatched();
 
-        futures.await().whenComplete((completeValueInfos, throwable) -> {
+        GraphQLContext graphQLContext = executionContext.getGraphQLContext();
+        futures.await(graphQLContext).whenComplete((completeValueInfos, throwable) -> {
```
bbakerman: See me comment above about knowledge in Async class.

I think the "special behavior checks" should be done here

as in

```
CompleteableFuture<Void> cancelCF = useCancelCF(executionContext)
futures.await(cancelCF).....

```

and then do the specialised code in the `useCancelCF` method

src/main/java/graphql/execution/Async.java
```diff
@@ -232,6 +257,101 @@ public CompletableFuture<List<T>> await() {
...
+            if (cancellationFuture == null) {
```
bbakerman, line 326: All this new code should get unit tests without all the GraphqContext aspects.

eg pass in a "cancelCF" adn then write lots of unit tests that JUST tests this new waiting on 2 sets of CFs and partial results.

I know the other tests run it BUT this is meant to be a generic utility - so write it as one and unit test it one

src/test/groovy/graphql/execution/AsyncTest.groovy
```diff
@@ -421,4 +421,145 @@ class AsyncTest extends Specification {
...
+        def resultCF = asyncBuilder.await(cancelCF)
+
+        // complete all CFs before cancellation
+        cf1.complete("X")
+        cf2.complete("Y")
+
+        def list = resultCF.join()
+
+        then:
+        list == ["X", "Y"]
```
bbakerman, line 584: not sure this test does what is says.  It never completes `cancelCF`

So what is it testing?

src/test/groovy/graphql/execution/AsyncTest.groovy
```diff
@@ -421,4 +421,145 @@ class AsyncTest extends Specification {
...
+        when:
+        def cancelCF = new CompletableFuture<Void>()
+        def asyncBuilder = Async.ofExpectedSize(3)
+        asyncBuilder.addObject("A")
+        asyncBuilder.addObject("B")
+        asyncBuilder.addObject("C")
+        def list = asyncBuilder.await(cancelCF).join()
+
+        then:
+        list == ["A", "B", "C"]
```
bbakerman, line 622: Maybe add a test with a mix of materialised and CF objects

src/test/groovy/graphql/execution/AsyncTest.groovy
```diff
@@ -421,4 +421,145 @@ class AsyncTest extends Specification {
...
+
+    def "await with cancelCF on single builder delegates to plain await"() {
+        when:
+        def cancelCF = new CompletableFuture<Void>()
+        def asyncBuilder = Async.ofExpectedSize(1)
+        asyncBuilder.add(completedFuture("A"))
+        def list = asyncBuilder.await(cancelCF).join()
+
+        then:
+        list == ["A"]
```
bbakerman, line 723: Is there some condition then where if there is only 1 Cf then the cancel wont be listened to?

Should we test for that?

src/main/java/graphql/execution/AsyncExecutionStrategy.java
```diff
@@ -64,14 +67,36 @@ public CompletableFuture<ExecutionResult> execute(ExecutionContext executionCont
...
+        CompletableFuture<Void> cancelCF = getCancellationFuture(executionContext);
+        futures.await(cancelCF).whenComplete((completeValueInfos, throwable) -> {
             List<String> fieldsExecutedOnInitialResult = deferredExecutionSupport.getNonDeferredFieldNames(fieldNames);
 
-            BiConsumer<List<Object>, Throwable> handleResultsConsumer = handleResults(executionContext, fieldsExecutedOnInitialResult, overallResult);
             throwable = executionContext.possibleCancellation(throwable);
 
             if (throwable != null) {
-                handleResultsConsumer.accept(null, throwable.getCause());
+                if (capturePartialResults(executionContext) && completeValueInfos != null) {
```
bbakerman: maybe reverse that check `completeValueInfos != null` is cheaper than checking `capturePartialResults(executionContext)`

bbakerman (approved):
I have left some improvement comments but overall I am happy with this

src/test/groovy/graphql/execution/AsyncTest.groovy
AlexandreCarlton: I think we had this with `await with cancelCF returns partial results with mixed objects and CFs` - are you thinking of a different kind of test case?

src/test/groovy/graphql/execution/AsyncTest.groovy
AlexandreCarlton: The idea was that we still get our results out even if the `cancelCF` hasn't completed. I've fixed up the test to clarify this.

src/test/groovy/graphql/execution/AsyncTest.groovy
AlexandreCarlton: Good point - have added a test for this.

src/main/java/graphql/execution/AsyncExecutionStrategy.java
```diff
@@ -64,14 +67,36 @@ public CompletableFuture<ExecutionResult> execute(ExecutionContext executionCont
...
+        CompletableFuture<Void> cancelCF = getCancellationFuture(executionContext);
+        futures.await(cancelCF).whenComplete((completeValueInfos, throwable) -> {
             List<String> fieldsExecutedOnInitialResult = deferredExecutionSupport.getNonDeferredFieldNames(fieldNames);
 
-            BiConsumer<List<Object>, Throwable> handleResultsConsumer = handleResults(executionContext, fieldsExecutedOnInitialResult, overallResult);
             throwable = executionContext.possibleCancellation(throwable);
 
             if (throwable != null) {
-                handleResultsConsumer.accept(null, throwable.getCause());
+                if (completeValueInfos != null && capturePartialResults(executionContext)) {
```
andimarek: So no calling handle results consumer anymore?

src/main/java/graphql/execution/Async.java
```diff
@@ -232,6 +256,96 @@ public CompletableFuture<List<T>> await() {
...
+            }
+
+            CompletableFuture<List<T>> overallResult = new CompletableFuture<>();
+            CompletableFuture<T>[] cfsArr = copyOnlyCFsToArray();
+            CompletableFuture<Void> allOf = CompletableFuture.allOf(cfsArr);
+
+            // race: either all CFs complete normally, or cancellation fires first
+            CompletableFuture.anyOf(allOf, cancellationFuture)
+                    .whenComplete((ignored, exception) -> {
+                        if (overallResult.isDone()) {
```
andimarek: How can overall result be done here? 

Our code coverage check will punish you for non reachable code, everything must be testeable

src/main/java/graphql/execution/Async.java
```diff
@@ -232,6 +256,96 @@ public CompletableFuture<List<T>> await() {
...
+                            List<T> partialResults = harvestPartialResults(array);
+                            overallResult.complete(partialResults);
+                            return;
+                        }
+                        // allOf finished normally: collect all results
+                        overallResult.complete(collectAllResults(array, cfsArr));
+                    });
+
+            // also handle when allOf completes (either normally or exceptionally) after the race
+            allOf.whenComplete((ignored, exception) -> {
```
andimarek: I don't fully understand this here: when can this happen exactly and why do we want to handle it?

andimarek:
@AlexandreCarlton In general we need the 100% code coverage to proof that all new code is actually reachable (because AI is great in inventing unnecessary defensive logic and checks)

src/main/java/graphql/execution/Async.java
bbakerman: This struck me as strange as well.  The first `CF.anyOf()` can ONLY return Atomically if "either" the `cancelCF `is finished or the  `allOf` CF is finished.  In which case the first `.whenComplete` above will have run.

But I think the logic has been split into two things here and its kinda confusing

I think we should change this to one `.whenComplete` and do


* cancelCF completes with exception
     then it harvests what results it can
     `overallResult` is completed successfully with data

* allOf CF completes with successfully
     then all data is harvested as per normal
     `overallResult` is completed successfully with data

* allOf CF completes with exception
     `overallResult` is completed with that exception


* cancelCF completes without exception
     we can never expect this (the cancelCF MUST always be completed via exception)
     `overallResult` is completed an Assert exception about bad code
     this should never happen if the cancelCF is made to only complete via exception


We should be able to unit test this

src/main/java/graphql/execution/Async.java
```diff
@@ -104,6 +118,11 @@ public CompletableFuture<List<T>> await() {
             return typedEmpty();
         }
 
+        @Override
+        public CompletableFuture<List<T>> await(@Nullable CompletableFuture<Void> cancellationFuture) {
+            return await();
+        }
```
bbakerman, line 157: I feel like this is wrong.   This is the "single" value version of the builder.  So this will ALWAYS wait for the data and never wait on the cancellation future

BUT what if?

* the cancellation future finished first
* BUT at the same time so did the "single CF" and hence this is a partial result.

OR

*  the cancellation future finished first
* the single CF is not finished 
* Surely this should return "on cancel" fast!


For consistency this should return IF cancel is done right??

This is a bug and should be unit testable!

bbakerman: Instead I made https://github.com/graphql-java/graphql-java/pull/4399 which  is just the Async changes

bbakerman:
@copilot resolve the merge conflicts in this pull request

src/main/java/graphql/execution/Async.java
```diff
@@ -195,6 +195,11 @@ public CompletableFuture<List<T>> await(@Nullable CompletableFuture<Void> cancel
             return CompletableFuture.completedFuture(Collections.singletonList((T) value));
         }
 
+        @Override
+        public CompletableFuture<List<T>> await(@Nullable CompletableFuture<Void> cancellationFuture) {
+            return await();
+        }
+
```
bbakerman: Sorry this is a bad merge via the web editor and now its not compiling.

I fixed locally but I cant push back to Alexandre branch it seems

src/main/java/graphql/execution/AbstractAsyncExecutionStrategy.java
```diff
@@ -25,12 +25,25 @@ protected BiConsumer<List<Object>, Throwable> handleResults(ExecutionContext exe
             exception = executionContext.possibleCancellation(exception);
 
             if (exception != null) {
+                // A cancellation that fired after some fields already completed arrives here as a
+                // synthesised AbortExecutionException with a non-null results list (a real field
+                // failure always has null results). When partial capture is enabled we keep those
+                // results and attach the cancellation error; otherwise we report the error as usual.
+                if (results != null && capturePartialResults(executionContext)) {
+                    executionContext.addError((AbortExecutionException) exception);
```
AlexandreCarlton, line 33: It dawns on me that it would be very useful to know _which_ fields were still executing when we interrupted (as they or their parents were likely the reason we timed out) - but this would add a considerable amount of complexity to an already complex PR. Maybe for a future one.

AlexandreCarlton:
Hi @andimarek / @bbakerman  (and also @dondonz ) - is there any interest in merging this PR?

As a recap, we would like to terminate a GraphQL request after a set amount of time has passed (e.g. 30s) but recuperate the results that have been processed.

There are two things that I think are missing from this PR, but hope to defer to subsequent PRs as they are edge cases:

 - if a non-null field is interrupted, this is not propagated upward.
 - we do not capture the fields that were interrupted in our errors (which would be useful for diagnostics).

Would it be OK to merge this provided there is nothing outstanding?

PR #4401, opened 2026-06-02
Add support for `@semanticNonNull`

martinbonnin:
`@semanticNonNull` is the counterpart of [`onError`](https://github.com/graphql-java/graphql-java/pull/4378).

* `onError` is aimed to be merged in the spec and allow new services to use true nullability (`!`) without conflating nulls and errors.
* `@semanticNonNull` is  for the transition. It allows existing services that cannot change fields to non-nullable (`!`) because that would increase the blast radius of errors for older clients to still signal semantic nullability to newer clients.

This PR adds support of `@semanticNonNull` so that there can never be null without an error in those locations. If a resolver ever returns `null`, an error is added, just like for non-null fields.

See the [`@semanticNonNull` GAP](https://gaps.graphql.org/GAP-49/draft/) for more details.

### DIFF
--- a/src/main/java/graphql/Directives.java
+++ b/src/main/java/graphql/Directives.java
@@ -1,21 +1,27 @@
 package graphql;
 
 
+import graphql.language.ArrayValue;
 import graphql.language.BooleanValue;
 import graphql.language.Description;
 import graphql.language.DirectiveDefinition;
+import graphql.language.IntValue;
+import graphql.language.ListType;
 import graphql.language.StringValue;
 import graphql.schema.GraphQLDirective;
 import org.jspecify.annotations.NullMarked;
 
+import java.math.BigInteger;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static graphql.Scalars.GraphQLBoolean;
+import static graphql.Scalars.GraphQLInt;
 import static graphql.Scalars.GraphQLString;
 import static graphql.introspection.Introspection.DirectiveLocation.ARGUMENT_DEFINITION;
 import static graphql.introspection.Introspection.DirectiveLocation.DIRECTIVE_DEFINITION;
@@ -35,6 +41,7 @@ import static graphql.language.InputValueDefinition.newInputValueDefinition;
 import static graphql.language.NonNullType.newNonNullType;
 import static graphql.language.TypeName.newTypeName;
 import static graphql.schema.GraphQLArgument.newArgument;
+import static graphql.schema.GraphQLList.list;
 import static graphql.schema.GraphQLNonNull.nonNull;
 
 /**
@@ -51,6 +58,8 @@ public class Directives {
     private static final String ONE_OF = "oneOf";
     private static final String DEFER = "defer";
     private static final String EXPERIMENTAL_DISABLE_ERROR_PROPAGATION = "experimental_disableErrorPropagation";
+    private static final String SEMANTIC_NON_NULL = "semanticNonNull";
+    private static final String INT = "Int";
 
     public static final DirectiveDefinition DEPRECATED_DIRECTIVE_DEFINITION;
     public static final DirectiveDefinition INCLUDE_DIRECTIVE_DEFINITION;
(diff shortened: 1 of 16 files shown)


src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java
```diff
@@ -80,13 +80,7 @@
...
 
 import static graphql.Assert.assertNotNull;
-import static graphql.Directives.DEPRECATED_DIRECTIVE_DEFINITION;
-import static graphql.Directives.IncludeDirective;
-import static graphql.Directives.NO_LONGER_SUPPORTED;
-import static graphql.Directives.ONE_OF_DIRECTIVE_DEFINITION;
-import static graphql.Directives.SPECIFIED_BY_DIRECTIVE_DEFINITION;
-import static graphql.Directives.SkipDirective;
-import static graphql.Directives.SpecifiedByDirective;
+import static graphql.Directives.*;
```
bbakerman: We dont want star imports please

src/main/java/graphql/Directives.java
```diff
@@ -256,6 +282,27 @@ public class Directives {
...
+    /**
+     * The "@semanticNonNull" directive indicates that a field is semantically non-null: it is only null if there is a
+     * matching error in the `errors` array.
+     * <p>
+     * See <a href="https://specs.apollo.dev/nullability/v0.4/">the Apollo nullability specification</a>
+     */
+    @ExperimentalApi
+    public static final GraphQLDirective SemanticNonNullDirective = GraphQLDirective.newDirective()
+            .name(SEMANTIC_NON_NULL)
+            .description("Indicates that a position is semantically non null: it is only null if there is a matching error in the `errors` array.")
```
bbakerman, line 294: I like the javadoc and descriptions - well done

src/test/groovy/graphql/execution/SemanticNonNullTest.groovy
```diff
@@ -0,0 +1,180 @@
...
+        def graphql = TestUtil.graphQL(sdl).build()
+
+        when:
+        def ei = ExecutionInput.newExecutionInput("{ bar { foo } }").root([bar: [foo: null]]).build()
+        def er = graphql.execute(ei)
+
+        then:
+        er.data.bar != null
+        er.data.bar.foo == null
+        er.errors.size() == 1
```
bbakerman, line 92: probably should check its the error type we expect

bbakerman: And should repeat that in all these test classes

src/test/groovy/graphql/schema/diffing/SchemaDiffingTest.groovy
```diff
@@ -935,7 +935,7 @@ class SchemaDiffingTest extends Specification {
         given:
         def schema1 = schema("""
            type Query {
-            foo(arg: Int): String
+            foo(arg: Float): String
```
bbakerman, line 938: Why Int to Float??

src/main/java/graphql/execution/NonNullableFieldValidator.java
```diff
@@ -50,17 +62,73 @@ public <T> T validate(ExecutionStrategyParameters parameters, T result) throws N
 
                 NonNullableFieldWasNullException nonNullException = new NonNullableFieldWasNullException(executionStepInfo, path);
                 final GraphQLError error = new NonNullableFieldWasNullError(nonNullException);
-                if(parameters.getAlternativeCallContext() != null) {
-                    parameters.getAlternativeCallContext().addError(error);
-                } else {
-                    executionContext.addError(error, path);
-                }
```
bbakerman, line 57: Nice refactor into common method

src/test/groovy/graphql/schema/diffing/SchemaDiffingTest.groovy
martinbonnin: Ah yes, sorry should have left a comment there. This is because the `@semanticNonNull` directive definition now puts the `Int` type by defaults in all schemas. So it's either we change to float, or we check `operations.size() == 3` just below. 

Since this class is about testing schema diffs, I thought keeping the diff as large as possible was the best way forward.

src/main/java/graphql/schema/idl/SchemaGeneratorHelper.java
martinbonnin: Woops, forgot about that. I should really tweek my IDE settings 😅 

https://github.com/graphql-java/graphql-java/pull/4401/changes/04c0bff68e85a8d99fa09394c1cd7d156626dfe7

src/test/groovy/graphql/execution/SemanticNonNullTest.groovy
martinbonnin: https://github.com/graphql-java/graphql-java/pull/4401/changes/0e29963516288313cafcc5e8badafdd3073401ec

martinbonnin:
I think this is good to go. 

2 comments:
* @michaelstaib, is that matching what you're doing in HotChocolate? 
* The jacoco test coverage fails but I think only on unreachable paths? The GraphQL definition should make it impossible to have `levels == null` for an example. Sounds like we should update the baseline? But I'm not sure how to do this. Let me know if you have any thoughts around this.

PR #4441, opened 2026-08-18
Validation cleanup

andimarek:
cleanup


bbakerman (approved):
I hope the AI got it right.  I cant read OperationValidator code.

but the tests looks right so...

PR #4442, opened 2026-08-18
parser cleanup

andimarek:
cleanup


src/main/java/graphql/parser/StringValueParsing.java
```diff
@@ -44,45 +41,25 @@ public static String removeIndentation(String rawValue) {
                 }
             }
         }
-        List<String> lineList = new ArrayList<>(Arrays.asList(lines));
-        if (commonIndent != null) {
```
bbakerman, line 48: The old code didnt do any work if `commonIndent == null` - can we do the same ??

bbakerman (approved):
Nice

PR #4445, opened 2026-08-19
Remove dependency annotations from shaded classes

andimarek:
## Summary

- define an explicit list of unbundled CLASS-retention annotations to remove from shaded class files
- add an ASM-powered cleanup stage after shadowing and before final JAR assembly
- preserve invisible annotations whose annotation classes are bundled, including GraphQL Java and relocated dependency annotations
- reject new RuntimeInvisibleAnnotations and RuntimeInvisibleParameterAnnotations whose annotation classes are not bundled and are not in the removal list
- share the class-file rewrite machinery with the existing JaCoCo annotation task
- compile a Java 11 consumer with class-file warnings treated as errors
- make the annotation-cleaned JAR the sole unclassified binary artifact and the sole main artifact used by consumers, signing, and publication

## Artifact pipeline

The regular `jar` and `shadowJar` outputs are internal inputs to the cleanup pipeline. They now have explicit `plain` and `shadow` classifiers and live under `build/intermediates`, outside `build/libs`.

`buildFinalJar` is the only task that writes the unclassified binary JAR to `build/libs`. It consumes the annotation-cleaned class tree directly, so Gradle tracks the actual final archive instead of a temporary archive that is renamed in `doLast`.

The Java API and runtime variants both expose `buildFinalJar`. The Shadow Java variant and assemble hook are disabled, and `shadowRuntimeElements` is non-consumable, so dependency resolution cannot select the uncleaned intermediate. Direct signing, Maven Local publication, and remote publication all depend on verification of the final JAR.

This separation matters because multiple tasks previously owned the same output path. A direct `shadowJar` invocation could overwrite the cleaned JAR while Gradle still considered the cleanup/reassembly tasks up to date. Unique intermediate paths plus a single final producer remove that failure mode.

## Verification

- `./gradlew outgoingVariants`: API and runtime variants expose only `buildFinalJar`; no consumable Shadow variant remains
- `./gradlew clean test assemble verifyShadedClassAnnotations compileShadedJarConsumer`: passed, including 5,737 tests
- `./gradlew check -x test -x testng`: passed
- isolated `publishGraphqlJavaPublicationToMavenLocal`: published only the main, sources, Javadoc, and POM artifacts
- the isolated Maven-local main JAR had the same SHA-256 as the `buildFinalJar` output
- direct `./gradlew jar shadowJar` invocation left the final JAR SHA-256 unchanged

The cleanup currently removes 1,345 annotations from 330 shaded class files.

## Artifact size

Both artifacts were built through their complete publication pipelines with the same fixed `999.0.0-size-test` release version.

| Artifact | Size |
| --- | ---: |
| `origin/master` before removal | 3,878,053 bytes |
| This PR after removal | 3,873,126 bytes |
| Reduction | 4,927 bytes (0.127%) |

The uncompressed JAR contents decrease by 16,152 bytes. Both archives contain 2,394 entries, and their manifests are identical.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -379,6 +379,11 @@ def annotationsToRemoveFromShadedJar = [
         'Lcom/google/j2objc/annotations/Weak;',
         'Ljavax/annotation/meta/TypeQualifierNickname;',
 ] as Set<String>
+def graphqlJavaAnnotationDescriptorPrefix = 'Lgraphql/'
+def shadedDependencyAnnotationDescriptorPrefixes = [
+        'Lgraphql/com/google/',
+        'Lgraphql/org/antlr/',
+]
 
 /**
  * Expands the shaded JAR into a clean working directory and removes any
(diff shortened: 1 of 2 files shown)


bbakerman:
I am glad AI can read it - I cant

bbakerman (approved):
Why not - hope it works

PR #4467, opened 2026-08-28
Fix non-null union field implements nullable union interface field

tonyghita:
According to the spec [IsValidImplementationFieldType()](https://spec.graphql.org/draft/#IsValidImplementationFieldType()), concrete type fields are able to strengthen the nullability guarantees made by an interface type field. 

```graphql
interface I { f: F }
type T implements I { f: F! }
```

This mostly happens, except when the field is a `Union` type.

```graphql
type A { a: ID }
union U = A
interface I { u: U }

type B implements I {
  u: U! # valid, but treated as invalid
}

schema { query: B }
```

This change moves the branch for `isNonNull` up so that non-null unions fall-through hit the unwrap case first. This is closer to how the specification is written.

### DIFF
--- a/src/main/java/graphql/schema/validation/TypesImplementInterfaces.java
+++ b/src/main/java/graphql/schema/validation/TypesImplementInterfaces.java
@@ -189,16 +189,6 @@ public class TypesImplementInterfaces extends GraphQLTypeVisitorStub {
     boolean isCompatible(GraphQLOutputType constraintType, GraphQLOutputType objectType) {
         if (isSameType(constraintType, objectType)) {
             return true;
-        } else if (constraintType instanceof GraphQLUnionType && objectType instanceof GraphQLObjectType) {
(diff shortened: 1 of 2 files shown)


tonyghita:
Alternatively we could always unwrap non-null first?

andimarek:
Have not looked into it further, but what about nullability inside list and list of lists?

tonyghita:
@andimarek I added a more comprehensive test across different type combinations in 2e21d2a. I only found the case when a `union` field strengthens nullability to fail. 

I moved the branch for unwrapping non-null types up so that the control flow is closer to the spec, rather than the extra `instanceOf` check on `objectType`.

tonyghita:
It seems like moving the branches around caused some changes in test branch coverage metrics. 

```
graphql.schema.validation.TypesImplementInterfaces Branch: 90.4% (was 94.2%, delta -3.8%, missed: 5 was 3)
```

I added some tests for the missing branches as well to get things unblocked in 38e843e.

PR #4474, opened 2026-09-18
reduce regex usage in description parsing

xuorig:
Hello folks 👋 

Please let me know if I'm missing something about the regex usage. But I think we can get away with this. I noticed a large part of schema parsing is spent preparing descriptions right now.

These three changes make a difference:

- `replaceAll` → `replace` so that we don't need to compile a regex
- split("\\n") → split("\n"):  String.split has a fast path for single char avoiding regex
- StringWriter → StringBuilder: no need for writer in this context

I know description parsing can be thorny so let me know if I missed something obvious.


bbakerman (commented):
Thanks for that - it seems much more sensible code

PR #4476, opened 2026-09-19
Preserve DataFetcherResult extensions on subscription events

### LINKED ISSUE #4475: Subscription event DataFetcherResult.extensions are missing from emitted ExecutionResult
When a subscription publisher emits a `DataFetcherResult` containing extensions, those extensions are not included in the emitted `ExecutionResult`.

  ### Reproduction

  A subscription data fetcher returns:

  ```java
  return Mono.just(
      DataFetcherResult.newResult()
          .data(event)
          .extensions(Map.of("myKey", "myValue"))
          .build()
  );
```

The emitted result contains the expected data, but:

`executionResult.getExtensions() == null`

Adding the value directly to the ExtensionsBuilder during the initial data-fetcher invocation causes it to appear on the initial subscription result:
 
 ```java
  dfe.getGraphQlContext()
      .get(ExtensionsBuilder.class)
      .addValue("myKey", "myValue");
```

However, extensions supplied by an emitted DataFetcherResult are missing from the corresponding subscription event result.

kalayciburak:
## Summary

Fixes #4475.

When a subscription publisher emits a `DataFetcherResult` that carries `extensions`, those values never appear on the corresponding event `ExecutionResult`. `unboxPossibleDataFetcherResult` records them on the shared `ExtensionsBuilder`, but `executeSubscriptionEvent` builds each event result with only data and errors, and `wrapWithRootFieldName` then reconstructs the result without copying extensions.

Query execution already merges `ExtensionsBuilder` into the overall result. Subscription events are independent results, so the extensions from that event's `DataFetcherResult` need to travel with that event.

### Change

- Capture `DataFetcherResult.getExtensions()` from the event payload.
- Pass them into the event `ExecutionResultImpl`.
- Preserve extensions when wrapping the payload under the root field name.

## Test plan

Executed locally:

- [x] RED: `./gradlew test --tests graphql.execution.SubscriptionExecutionStrategyTest` — 24 tests, 1 failed (`extensions=null` on the new test)
- [x] GREEN: same class — 24/0
- [x] `./gradlew test --tests graphql.execution.DataFetcherResultTest --tests graphql.extensions.ExtensionsBuilderTest` — 36/0


bbakerman (approved):
Nice one

